PR #1030, opened 2014-11-19
Let fleetd schedule unit onto machine via SystemStat

tg123:
engine: enables scheduler decide which machines to run the job

fleetd is now choosing machine to load unit by how many units
are running on it. more reasonable way may be like chooing
a machine that has the most hardware resources, like load5 mem,
on it.

```
[X-Fleet]
Scheduler=LeastSystemStatField
SchedulerMetadata="sysstatfield=load1"
```

`sysstatfield` is now support `cpu_used`, `mem_free`, `load1`, `load5`, `load15`

`sort by loaded unit count` is the default scheduler if no config provided.


bcwaldon:
Tracking resource usage directly in fleet has not been identified as a necessary requirement, and neither has scheduling based on that data. Tracking resources may fall out of #1055, but we need to plan this out before accepting any related changes.

PR #1034, opened 2014-11-20
Switch default fleetctl driver to API

### LINKED ISSUE #921: Switch default fleetctl transport to HTTP API
fleetctl currently hits etcd directly by default. A user can force fleetctl to hit the HTTP API with the `--experimental-api` and `--endpoint` flags. The behavior should be inverted when the API is locked down.

bcwaldon:
Fix #921


cnelson:
FYI, OpenSSH 6.7 contains support for forwarding unix domain sockets.  There's an open request to have it updated from upstream:

https://github.com/coreos/portage-stable/pull/173

crawford:
LGTM

PR #1076, opened 2014-12-24
ssh username support

### LINKED ISSUE #536: fleetctl: make ssh user configurable
Currently, the `core` user is hardcoded throughout the `fleetctl` code, both for tunnelling and the `ssh` action. We would like to be able to customize it, so that users created in our coreos instances via cloudconfig can be used to query fleet remotely, allowing for more granularity in term of security, without having to use vendor-specific mechanisms.

fly:
fixes #536
continuation of #904


bcwaldon:
Please reopen if you're able to address the issues raised by @jonboulle

PR #1077, opened 2014-12-25
Dynamic Metadata

epipho:
Updated 2/28/2015

- Added new PATCH method to the machines collection endpoint to allow changing a machines metadata via the http endpoint.
- API uses the jsonpatch format. Multiple machines can be modified at once, including machines that are not currently part of the cluster.  "add", "replace", and "remove" operations are available.
- Metadata modified via the api is retained even if the machine leaves and rejoins the cluster.
- Dynamic metadata (metadata set by the user via the api) is merged with Machine metadata (metadata defined by the fleet config, env variables, or flags) using the following rules
   -  Any keys that exist in only one of the two collections are added to the final collection as-is
   - Any keys that exist in both, the value set in the dynamic metadata is added to the final collection.
   - Any keys that are the string zero-value in the dynamic metadata are considered deleted and are not included in the final collection. This lets a user persistently delete a value set at configuration time, even if the machine leaves and rejoins the cluster.
- The reconcile will reschedule any units on machines that no longer meet the metadata requirements (pre-existing functionality)

Known Issues:
- Docs not complete

### DIFF
--- a/api/machines.go
+++ b/api/machines.go
@@ -125,7 +125,7 @@ func (mr *machinesResource) patch(rw http.ResponseWriter, req *http.Request) {
 			}
 		}
 	}
-	sendResponse(rw, http.StatusNoContent, "")
+	sendResponse(rw, http.StatusNoContent, nil)
 }
 
 func getMachinePage(cAPI client.API, tok PageToken) (*schema.MachinePage, error) {
--- a/api/machines.go
+++ b/api/machines.go
@@ -15,14 +15,20 @@
 package api
 
 import (
-	"fmt"
+	"encoding/json"
+	"errors"
 	"net/http"
 	"path"
+	"regexp"
 
 	"github.com/coreos/fleet/client"
 	"github.com/coreos/fleet/log"
 	"github.com/coreos/fleet/machine"
 	"github.com/coreos/fleet/schema"
+)
+
+var (
+	metadataPathRegex = regexp.MustCompile("^/([^/]+)/metadata/([A-Za-z0-9_.-]+$)")
 )
 
 func wireUpMachinesResource(mux *http.ServeMux, prefix string, cAPI client.API) {
@@ -35,12 +41,24 @@ type machinesResource struct {
 	cAPI client.API
 }
 
-func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	if req.Method != "GET" {
-		sendError(rw, http.StatusBadRequest, fmt.Errorf("only HTTP GET supported against this resource"))
-		return
-	}
+type machineMetadataOp struct {
+	Operation string `json:"op"`
+	Path      string
+	Value     string
+}
 
+func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	switch req.Method {
+	case "GET":
+		mr.list(rw, req)
+	case "PATCH":
+		mr.patch(rw, req)
+	default:
+		sendError(rw, http.StatusMethodNotAllowed, errors.New("only GET and PATCH supported against this resource"))
+	}
+}
+
+func (mr *machinesResource) list(rw http.ResponseWriter, req *http.Request) {
 	token, err := findNextPageToken(req.URL)
 	if err != nil {
 		sendError(rw, http.StatusBadRequest, err)
@@ -60,6 +78,54 @@ func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 	}
 
 	sendResponse(rw, http.StatusOK, page)
+}
+
+func (mr *machinesResource) patch(rw http.ResponseWriter, req *http.Request) {
+	var ops []machineMetadataOp
+	dec := json.NewDecoder(req.Body)
+	if err := dec.Decode(&ops); err != nil {
+		sendError(rw, http.StatusBadRequest, err)
+		return
+	}
+
+	for _, op := range ops {
+		if op.Operation != "add" && op.Operation != "remove" && op.Operation != "replace" {
+			sendError(rw, http.StatusBadRequest, errors.New("invalid op: expect add, remove, or replace"))
+			return
+		}
+
+		if metadataPathRegex.FindStringSubmatch(op.Path) == nil {
+			sendError(rw, http.StatusBadRequest, errors.New("machine metadata path invalid"))
+			return
+		}
+
+		if op.Operation != "remove" && len(op.Value) == 0 {
+			sendError(rw, http.StatusBadRequest, errors.New("invalid value: add and replace require a value"))
+			return
+		}
+	}
+
+	for _, op := range ops {
+		// regex already validated above
+		s := metadataPathRegex.FindStringSubmatch(op.Path)
+		machID := s[1]
+		key := s[2]
+
+		if op.Operation == "remove" {
+			err := mr.cAPI.DeleteMachineMetadata(machID, key)
+			if err != nil {
+				sendError(rw, http.StatusInternalServerError, err)
+				return
+			}
+		} else {
+			err := mr.cAPI.SetMachineMetadata(machID, key, op.Value)
+			if err != nil {
+				sendError(rw, http.StatusInternalServerError, err)
+				return
+			}
+		}
+	}
+	sendResponse(rw, http.StatusNoContent, "")
 }
 
 func getMachinePage(cAPI client.API, tok PageToken) (*schema.MachinePage, error) {
--- a/client/api.go
+++ b/client/api.go
@@ -21,6 +21,8 @@ import (
 
 type API interface {
 	Machines() ([]machine.MachineState, error)
+	SetMachineMetadata(machID, key, value string) error
+	DeleteMachineMetadata(machID, key string) error
 
 	Unit(string) (*schema.Unit, error)
 	Units() ([]*schema.Unit, error)
--- a/registry/interface.go
+++ b/registry/interface.go
@@ -37,6 +37,8 @@ type Registry interface {
 	SetUnitTargetState(name string, state job.JobState) error
 	SetMachineState(ms machine.MachineState, ttl time.Duration) (uint64, error)
 	UnscheduleUnit(name, machID string) error
+	SetMachineMetadata(machID, key, value string) error
+	DeleteMachineMetadata(machID, key string) error
 
 	UnitRegistry
 }
--- a/registry/interface.go
+++ b/registry/interface.go
@@ -36,6 +36,7 @@ type Registry interface {
 	ScheduleUnit(name, machID string) error
 	SetUnitTargetState(name string, state job.JobState) error
 	SetMachineState(ms machine.MachineState, ttl time.Duration) (uint64, error)
+	MachineState(machID string) (machine.MachineState, error)
 	UnscheduleUnit(name, machID string) error
 	SetMachineMetadata(machID, key, value string) error
 	DeleteMachineMetadata(machID, key string) error
--- a/registry/machine.go
+++ b/registry/machine.go
@@ -44,25 +44,12 @@ func (r *EtcdRegistry) Machines() (machines []machine.MachineState, err error) {
 
 	for _, node := range resp.Node.Nodes {
 		var mach machine.MachineState
-		var metadata map[string]string
-
-		for _, obj := range node.Nodes {
-			if strings.HasSuffix(obj.Key, "/object") {
-				err = unmarshal(obj.Value, &mach)
-				if err != nil {
-					return
-				}
-			} else if strings.HasSuffix(obj.Key, "/metadata") {
-				// Load metadata into a separate map to avoid stepping on it when deserializing the object key
-				metadata = make(map[string]string, len(obj.Nodes))
-				for _, mdnode := range obj.Nodes {
-					metadata[path.Base(mdnode.Key)] = mdnode.Value
-				}
-			}
+		mach, err = readMachineState(&node)
+		if err != nil {
+			return
 		}
 
 		if mach.ID != "" {
-			mach.Metadata = mergeMetadata(mach.Metadata, metadata)
 			machines = append(machines, mach)
 		}
 	}
@@ -101,6 +88,21 @@ func (r *EtcdRegistry) SetMachineState(ms machine.MachineState, ttl time.Duratio
 	}
 
 	return resp.Node.ModifiedIndex, nil
+}
+
+func (r *EtcdRegistry) MachineState(machID string) (machine.MachineState, error) {
+	req := etcd.Get{
+		Key:       path.Join(r.keyPrefix, machinePrefix, machID),
+		Sorted:    true,
+		Recursive: true,
+	}
+
+	resp, err := r.etcd.Do(&req)
+	if err != nil {
+		return machine.MachineState{}, err
+	}
+
+	return readMachineState(resp.Node)
 }
 
 func (r *EtcdRegistry) SetMachineMetadata(machID string, key string, value string) error {
@@ -154,4 +156,27 @@ func mergeMetadata(machineMetadata, dynamicMetadata map[string]string) map[strin
 		}
 	}
 	return finalMetadata
+}
+
+// readMachineState reads machine state from an etcd node
+func readMachineState(node *etcd.Node) (mach machine.MachineState, err error) {
+	var metadata map[string]string
+
+	for _, obj := range node.Nodes {
+		if strings.HasSuffix(obj.Key, "/object") {
+			err = unmarshal(obj.Value, &mach)
+			if err != nil {
+				return
+			}
+		} else if strings.HasSuffix(obj.Key, "/metadata") {
+			// Load metadata into a separate map to avoid stepping on it when deserializing the object key
+			metadata = make(map[string]string, len(obj.Nodes))
+			for _, mdnode := range obj.Nodes {
+				metadata[path.Base(mdnode.Key)] = mdnode.Value
+			}
+		}
+	}
+
+	mach.Metadata = mergeMetadata(mach.Metadata, metadata)
+	return
 }
--- a/registry/machine.go
+++ b/registry/machine.go
@@ -43,17 +43,26 @@ func (r *EtcdRegistry) Machines() (machines []machine.MachineState, err error) {
 	}
 
 	for _, node := range resp.Node.Nodes {
+		var mach machine.MachineState
+		var metadata map[string]string
+
 		for _, obj := range node.Nodes {
-			if !strings.HasSuffix(obj.Key, "/object") {
-				continue
+			if strings.HasSuffix(obj.Key, "/object") {
+				err = unmarshal(obj.Value, &mach)
+				if err != nil {
+					return
+				}
+			} else if strings.HasSuffix(obj.Key, "/metadata") {
+				// Load metadata into a separate map to avoid stepping on it when deserializing the object key
+				metadata = make(map[string]string, len(obj.Nodes))
+				for _, mdnode := range obj.Nodes {
+					metadata[path.Base(mdnode.Key)] = mdnode.Value
+				}
 			}
+		}
 
-			var mach machine.MachineState
-			err = unmarshal(obj.Value, &mach)
-			if err != nil {
-				return
-			}
-
+		if mach.ID != "" {
+			mach.Metadata = mergeMetadata(mach.Metadata, metadata)
 			machines = append(machines, mach)
 		}
 	}
@@ -94,6 +103,23 @@ func (r *EtcdRegistry) SetMachineState(ms machine.MachineState, ttl time.Duratio
 	return resp.Node.ModifiedIndex, nil
 }
 
+func (r *EtcdRegistry) SetMachineMetadata(machID string, key string, value string) error {
+	update := etcd.Set{
+		Key:   path.Join(r.keyPrefix, machinePrefix, machID, "metadata", key),
+		Value: value,
+	}
+
+	_, err := r.etcd.Do(&update)
+	return err
+}
+
+func (r *EtcdRegistry) DeleteMachineMetadata(machID string, key string) error {
+	// Deleting a key sets its value to "" to allow for intelligent merging
+	// between the machine-defined metadata and the dynamic metadata.
+	// See mergeMetadata for more detail.
+	return r.SetMachineMetadata(machID, key, "")
+}
+
 func (r *EtcdRegistry) RemoveMachineState(machID string) error {
 	req := etcd.Delete{
 		Key: path.Join(r.keyPrefix, machinePrefix, machID, "object"),
@@ -103,4 +129,29 @@ func (r *EtcdRegistry) RemoveMachineState(machID string) error {
 		err = nil
 	}
 	return err
+}
+
+// mergeMetadata merges the machine-set metadata with the dynamic metadata to better facilitate
+// machines leaving and rejoining a cluster.
+// Merging metadata uses the following rules:
+// - Any keys that are only in one collection are added as-is
+// - Any keys that exist in both, the dynamic value takes precence
+// - Any keys that have a zero-value string in the dynamic metadata are considered deleted
+//   and are not included in the final collection
+func mergeMetadata(machineMetadata, dynamicMetadata map[string]string) map[string]string {
+	if dynamicMetadata == nil {
+		return machineMetadata
+	}
+	finalMetadata := make(map[string]string, len(dynamicMetadata))
+	for k, v := range machineMetadata {
+		finalMetadata[k] = v
+	}
+	for k, v := range dynamicMetadata {
+		if v == "" {
+			delete(finalMetadata, k)
+		} else {
+			finalMetadata[k] = v
+		}
+	}
+	return finalMetadata
 }
--- a/Documentation/api-v1.md
+++ b/Documentation/api-v1.md
@@ -204,6 +204,32 @@ The request must not have a body.
 
 A successful response will contain a page of zero or more Machine entities.
 
+### Edit Machine Metadata
+
+Add, change, or remove metadata from one or more machines.
+
+#### Request
+
+```
+PATCH /machines HTTP/1.1
+
+[
+  { "op": "add", "path": "/<machine_id>/metadata/<name>", "value": <new value> },
+  { "op": "modify", "path": "/<machine_id>/metadata/<name>", "value": <new value> },
+  { "op": "delete", "path": "/<machine_id>/metadata/<name>" }
+]
+
+```
+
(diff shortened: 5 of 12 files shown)


epipho:
I think this is ready for the first pass of reviews.

bcwaldon:
I haven't looked too far into the actual code, but I wanted to share some high-level feedback since this is likely to change some things around:

I'd like to see the PATCH endpoint implemented using JSON patch [0] rather than a custom request format. Additionally, we should consider implementing the PATCH at the root of the machines collection rather than per-machine.

What utility does the TTL on a specific key provide?

[0] http://jsonpatch.com/

epipho:
Switching to JSONPatch is easy enough as is moving the PATCH at the root level to allow multiple machines to be updated at once. I'll have the changes made in a couple of days.

The original idea was to use it as kind of a health check, a process could check a node to see if it was suitable for a specific type of job and set the TTL'ed metadata. Furthur thoughts make it seem like the TTL idea is half baked / gold plating and maybe doesn't deserve to live, particularly with the new request format.

bcwaldon:
@epipho I do like the idea, but I'm not sure if we should try to squeeze it in right now.

epipho:
@bcwaldon 

- Updated to use jsonpatch on the machines collection level.  Metadata is now updated with the path format /machineid/metadata/key using the add/remove/replace ops.

```
[{"op": "add", "path": "/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/metadata/foo", "value": "bar"},
	 {"op": "replace", "path": "/bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb/metadata/ping", "value": "splat"}]
```

- TTL feature has been stashed away for another day

- Error handling: Currently the handler does as much as it can to validate the patch data before doing any work. Invalid ops, bad path format, or missing values will cause the request to fail with no effect on the data.  If an error occurs during the apply phase, the request terminates immediately and returns the error. Because the changes cannot be made atomic as the metadata keys are split apart this could (rarely) result in a partial patch being applied. Any suggestions on how to handle this or what information should be returned to the caller when this happens?

epipho:
Removing WIP. Ready for initial review and comments.

umiller:
How do I test the dynamic metadata?
I am looking for a safe way to change the CoreOs machine metadata in run-time
Is this feature is the only way? (I blocked CoreOs updates on my cluster..)

registry/interface.go
```diff
@@ -39,6 +39,8 @@ type Registry interface {
 	SetUnitTargetState(name string, state job.JobState) error
 	SetMachineState(ms machine.MachineState, ttl time.Duration) (uint64, error)
 	UnscheduleUnit(name, machID string) error
+	UpdateMachineMetadata(machID string, key string, value string) error
```
bcwaldon: Let's call this `SetMachineMetadata`, since `Update` can mean the key must already exist in some contexts.

And there's no need to respecify `string` three times, just use `SetMachineMetadata(machID, key, value string)`

machine/state.go
```diff
@@ -25,7 +25,7 @@ const (
 type MachineState struct {
 	ID       string
 	PublicIP string
-	Metadata map[string]string
+	Metadata map[string]string `json:"-"`
```
bcwaldon: Why is this change necessary?

api/machines.go
```diff
@@ -64,6 +83,59 @@ func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request)
...
+	}
+
+	// Validate all operations before applying
+	for _, op := range ops {
+		if op.Op != "add" && op.Op != "remove" && op.Op != "replace" {
+			sendError(rw, http.StatusBadRequest, fmt.Errorf("Bad op: %s. Expected add/remove/replace", op.Op))
+			return
+		}
+
+		s := metadataPathRegex.FindStringSubmatch(op.Path)
```
bcwaldon: Is it necessary to store the returned value in a separate variable, or can you just compare `metadataPathRegex.FindStringSubmatch(op.Path)` directly to `nil`?

api/machines.go
```diff
@@ -64,6 +83,59 @@ func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request)
...
+		if s == nil {
+			sendError(rw, http.StatusBadRequest, fmt.Errorf("Bad path: %s. Expected {machine id}/metadata/{key}. Key must only contain [a-zA-Z0-9_.-]", op.Path))
```
bcwaldon: This error message is communicating way too much:
1. error messages should always start with a lowercase character
2. don't write multiple sentences
3. don't mirror back user input

bcwaldon: This could be `fmt.Errorf("machine metadata path invalid")

api/machines.go
```diff
@@ -64,6 +83,59 @@ func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request)
...
+
+	if err := dec.Decode(&ops); err != nil {
+		sendError(rw, http.StatusInternalServerError, err)
+		return
+	}
+
+	// Validate all operations before applying
+	for _, op := range ops {
+		if op.Op != "add" && op.Op != "remove" && op.Op != "replace" {
+			sendError(rw, http.StatusBadRequest, fmt.Errorf("Bad op: %s. Expected add/remove/replace", op.Op))
```
bcwaldon: Simplify this error message, too:

`errors.New("invalid op: expect add, remove or replace")`

api/machines.go
```diff
@@ -64,6 +83,59 @@ func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 	sendResponse(rw, http.StatusOK, page)
 }
 
+func (mr *machinesResource) patch(rw http.ResponseWriter, req *http.Request) {
+	ops := make([]machineMetadataOp, 0)
+
```
bcwaldon: The newlines on lines 88 and 90 aren't necessary

api/machines.go
```diff
@@ -64,6 +83,59 @@ func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request)
...
+	dec := json.NewDecoder(req.Body)
+
+	if err := dec.Decode(&ops); err != nil {
+		sendError(rw, http.StatusInternalServerError, err)
```
bcwaldon: It's not a 500-level error if the user provided data that isn't valid JSON - it should be a StatusBadRequest (400).

api/machines.go
```diff
@@ -64,6 +83,59 @@ func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request)
...
+		return
+	}
+
+	// Validate all operations before applying
```
bcwaldon: This comment isn't necessary - the code is self-explanatory

api/machines.go
```diff
@@ -64,6 +83,59 @@ func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request)
...
+			}
+		} else {
+			err := mr.cAPI.UpdateMachineMetadata(machID, key, op.Value)
+			if err != nil {
+				sendError(rw, http.StatusInternalServerError, err)
+				return
+			}
+		}
+	}
+	sendResponse(rw, http.StatusOK, "")
```
bcwaldon: StatusNoContent, not StatusOK

api/machines.go
```diff
@@ -37,12 +44,24 @@ type machinesResource struct {
 	cAPI client.API
 }
 
+type machineMetadataOp struct {
+	Op    string
```
bcwaldon: Let's spell out Operation rather than shorten it to Op

api/machines.go
```diff
@@ -27,6 +30,10 @@ import (
 	"github.com/coreos/fleet/schema"
 )
 
+var (
+	metadataPathRegex = regexp.MustCompile("^/([0-9a-f]{32})/metadata/([A-Za-z0-9_.-]+$)")
```
bcwaldon: Where did the 32-character limit come from?

registry/machine.go
```diff
@@ -45,17 +45,30 @@ func (r *EtcdRegistry) Machines() (machines []machine.MachineState, err error) {
...
-			if !strings.HasSuffix(obj.Key, "/object") {
-				continue
-			}
+		var mach machine.MachineState
 
-			var mach machine.MachineState
-			err = unmarshal(obj.Value, &mach)
-			if err != nil {
-				return
+		_, machineID := path.Split(node.Key)
```
bcwaldon: mach.ID = path.Base(node.Key)

machine/state.go
bcwaldon: Is it due to how you're unmarshaling both the machine and the metadata into the same object independently in the registry?

registry/machine.go
```diff
@@ -93,9 +106,78 @@ func (r *EtcdRegistry) SetMachineState(ms machine.MachineState, ttl time.Duratio
...
+	if err = r.setMachineMetadata(ms); err != nil {
+		return uint64(0), err
+	}
+
 	return resp.Node.ModifiedIndex, nil
 }
 
+func (r *EtcdRegistry) UpdateMachineMetadata(machID string, key string, value string) error {
+	//Attempt to update key
+	update := etcd.Update{
```
bcwaldon: Why Update then Create? Can't you just use Set?

registry/machine.go
```diff
@@ -93,9 +106,78 @@ func (r *EtcdRegistry) SetMachineState(ms machine.MachineState, ttl time.Duratio
...
+	if isKeyNotFound(err) {
+		return nil
+	} else if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Remove and reset all metadata
+func (r *EtcdRegistry) setMachineMetadata(ms machine.MachineState) error {
```
bcwaldon: Just pass in the fields of MachineState that are relevant to this method: ID and Metadata. It simplifies the testing matrix.

registry/machine.go
```diff
@@ -93,9 +106,78 @@ func (r *EtcdRegistry) SetMachineState(ms machine.MachineState, ttl time.Duratio
...
+
+	return nil
+}
+
+func (r *EtcdRegistry) DeleteMachineMetadata(machID string, key string) error {
+	del := etcd.Delete{
+		Key: path.Join(r.keyPrefix, machinePrefix, machID, "metadata", key),
+	}
+
+	_, err := r.etcd.Do(&del)
```
bcwaldon: This is much simpler:

```
_, err := r.etcd.Do(&del)
if err != nil && isKeyNotFound(err) {
    err = nil
}
return err
```

registry/machine.go
```diff
@@ -45,17 +45,30 @@ func (r *EtcdRegistry) Machines() (machines []machine.MachineState, err error) {
...
+		for _, obj := range node.Nodes {
+			// Load machine object
+			if strings.HasSuffix(obj.Key, "/object") {
+				err = unmarshal(obj.Value, &mach)
+				if err != nil {
+					return
+				}
+			} else if strings.HasSuffix(obj.Key, "/metadata") {
+				// Load metadata
+				mach.Metadata = make(map[string]string)
```
bcwaldon: You can `make` the map with a known length `len(obj.Nodes)`

registry/machine.go
```diff
@@ -45,17 +45,30 @@ func (r *EtcdRegistry) Machines() (machines []machine.MachineState, err error) {
...
+				for _, mdnode := range obj.Nodes {
+					_, key := path.Split(mdnode.Key)
```
bcwaldon: `mach.Metadata[path.Base(mdnode.Key)] = mdnode.Value`

bcwaldon:
@epipho In addition to the inline comments, I have a concern about how the metadata is initially making it into etcd. Machines leave and rejoin the cluster frequently, so metadata that was manually added by someone through etcd will get blown away. I wonder if the metadata published by a given machine should be managed independently from the metadata added through etcd? I think it makes sense for metadata manually added through etcd directly should persist until it is manually removed.

machine/state.go
epipho: Yes. Since the metadata is no longer inside the main machine object, but rather in independent kv pairs, decoding the machine object could wipe out the metadata.

api/machines.go
epipho: Systemd defines a machine id this way.

http://www.freedesktop.org/software/systemd/man/machine-id.html

bcwaldon: fleet's machines don't necessarily have to use systemd's machine-id concept to identify themselves. It can be overridden in config. This needs to be an arbitrary string.

epipho:
@bcwaldon Thank you for taking the time to do a very thorough review. I will address all the small stuff this weekend.

As for your question above:

Wouldn't that be true today since a machine leaving the cluster removes the /machineid/object value that currently holds the metadata?

I'm not sure how to best handle your request, I outlined a few possibilities below.

1) Metadata that is configured via fleet config (file, env variable, cloud config) is automatic, removed by machine leave, and cannot be manipulated at all. All other metdata values are untouched.

2) Metadata configured by fleet config can be manipulated, but will be reset by the machine rejoining. All metadata values are untouched when a machine leaves.

3) Metadata configured by fleet config is only added if the machine does not exist in the registry at all. All metadata values are untouched when a machine leaves.

4) Same as 3, but all metadata is removed when a machine leaves (as currently written).

I worry that separating the metadata will be confusing as some will be able to be manipulated via the api and some will not. Of all the suggestions outlined I like 3, treating the configured values as defaults if the machine has never been seen before but leaving it untouched otherwise.

api/machines.go
epipho: Gotcha, will change.

bcwaldon:
@epipho It's not clear to me how it should work, so I'd like to get an opinion from @jonboulle on this.

epipho:
All feedback has been addressed, just waiting for clarification on the metadata persistence issue.

registry/machine.go
epipho: This was patterned after SetMachineState above. I agree Set makes more sense and have modified the code to use it, should SetMachineState be modified as well?

bcwaldon: No - SetMachineState does an Update then Set so the event sent to downstream watchers can be used to distinguish a new from an existing machine. It is fine to use Set here, as the event generated by the Set isn't used.

registry/machine.go
```diff
@@ -102,5 +165,16 @@ func (r *EtcdRegistry) RemoveMachineState(machID string) error {
 	if etcd.IsKeyNotFound(err) {
 		err = nil
 	}
+
+	// Delete metadata
```
bcwaldon: Could this not be inferred by the `Delete` action against the machine's `metadata` key? This comment is unnecessary.

registry/machine.go
```diff
@@ -91,9 +102,61 @@ func (r *EtcdRegistry) SetMachineState(ms machine.MachineState, ttl time.Duratio
...
+	}
+
+	_, err := r.etcd.Do(&del)
+	if etcd.IsKeyNotFound(err) {
+		err = nil
+	}
+	return err
+}
+
+// Remove and reset all metadata
```
bcwaldon: `// setMachineMetadata...`

registry/machine.go
```diff
@@ -91,9 +102,61 @@ func (r *EtcdRegistry) SetMachineState(ms machine.MachineState, ttl time.Duratio
...
+	// Set the initial metadata on creation
+	if err = r.setMachineMetadata(ms.ID, ms.Metadata); err != nil {
+		return uint64(0), err
+	}
+
 	return resp.Node.ModifiedIndex, nil
 }
 
+func (r *EtcdRegistry) SetMachineMetadata(machID string, key string, value string) error {
+	//Attempt to update key
```
bcwaldon: One last nit-picky thing: comments are `// <Comment text>`. Note the trailing space after the forward slashes and the first character of the comment body is uppercase. It's arbitrary, but let's at least be consistent.

This comment specifically isn't actually necessary, since you are no longer attempting to update the key, you are setting the key.

registry/machine.go
```diff
@@ -43,17 +43,28 @@ func (r *EtcdRegistry) Machines() (machines []machine.MachineState, err error) {
...
+				// Load metadata
+				mach.Metadata = make(map[string]string, len(obj.Nodes))
+				for _, mdnode := range obj.Nodes {
+					mach.Metadata[path.Base(mdnode.Key)] = mdnode.Value
+				}
 			}
+		}
 
+		// only use the machine if there was an object key
+		if len(mach.Version) > 0 {
```
bcwaldon: It would be safer to check for a nonzero machine ID, right? That's the unique piece of identifying information for a Machine object.

machine/state.go
bcwaldon: We really should not let registry-level serialization logic leak out into the machine package. Please copy the MachineState struct over to where this JSON tag matters. We've already taken this approach with `unitModel` and `unit.UnitFile` in registry/unit.go.

epipho: Changed the serialization code to load metadata separately and combine them at the end.

registry/machine.go
epipho: Fixed, originally I was pulling the machine id from the key which would not be overwritten if no object sub-key was found.

epipho:
@bcwaldon  @jonboulle  any thoughts yet on what to do about persistent metadata?

jonboulle:
I could have _sworn_ I wrote a reply to this ages ago but apparently not? :-(

I am still going back and forth but keep coming back to the thought that the sane behaviour would be to have user-set metadata stored completely separately from fleetd-set metadata, and then just make a decision on a simple merge strategy (likely, user-set-always-wins).

epipho:
Seems reasonable. How does this feel:

- Fleet metadata returns to the object.
- User metadata is stored separately as above.
- User metadata is never removed from etcd
- When loaded, any values in user metadata overwrite those from fleet
metadata.
- If a user issues a delete against a key in the fleet metadata, an empty
value is added to the user metadata to ensure the key remains removed if
the machine leaves and rejoins.


This allows for sane user-facing behavior as well as preserving the
behavior of changing the metadata in the fleet config and restarting the
service.

jonboulle:
Sounds reasonable to me, @bcwaldon ?

bcwaldon:
@epipho What does it mean for a user to issue a delete against a key in the fleet metadata?

epipho:
@bcwaldon I should have called them out as "user metadata" and "machine metadata".

If a user deletes a key defined by machine metadata (fleet config file, env variables in fleet.service) an entry will be added with an empty string to the dynamic user metadata to ensure the value stays deleted even if the machine leaves and rejoins the cluster.

epipho:
Merging machine and user metadata has been implemented.

api/machines.go
```diff
@@ -62,6 +80,54 @@ func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 	sendResponse(rw, http.StatusOK, page)
 }
 
+func (mr *machinesResource) patch(rw http.ResponseWriter, req *http.Request) {
+	ops := make([]machineMetadataOp, 0)
```
bcwaldon: I don't know if you actually need to `make` this if you're going to decode JSON into it.

client/api.go
```diff
@@ -21,6 +21,8 @@ import (
 
 type API interface {
 	Machines() ([]machine.MachineState, error)
+	SetMachineMetadata(machID string, key string, value string) error
```
bcwaldon: `SetMachineMetadata(machID, key, value string) error`

client/api.go
```diff
@@ -21,6 +21,8 @@ import (
...
+	DeleteMachineMetadata(machID string, key string) error
```
bcwaldon: `DeleteMachineMetadata(machID, key string) error`

registry/machine.go
```diff
@@ -104,3 +130,27 @@ func (r *EtcdRegistry) RemoveMachineState(machID string) error {
...
+		if _, ok := dynamicMetadata[k]; !ok {
+			dynamicMetadata[k] = v
+		}
+	}
+	for k, v := range dynamicMetadata {
+		if v == "" {
+			delete(dynamicMetadata, k)
+		}
+	}
+	return dynamicMetadata
```
bcwaldon: Wouldn't it be better to allocate a new map? Could this cause confusion for callers of this function?

bcwaldon:
@epipho can you update the PR description with the current state of things? We need some documentation written up for this new feature, too.

pwhiteside:
If a unit is scheduled to a machine based on metadata, and that metadata changes so the the unit no longer matches, should it be rescheduled elsewhere?

api/machines.go
epipho: Turns out you don't. A nil slice was fine.

registry/machine.go
epipho: I wasn't worried about caller confusion since it was internal to the package, but I agree still better to allocate a new map here.

epipho:
@pwhiteside yes, if the machine no longer satisfies the unit's requested metadata the reconciler will reschedule it.

epipho:
@bcwaldon I have updated the first post with all the current information. I am currently investigating the best way to make global units work properly with dynamic metadata, after which I will work on docs.

epipho:
Global units now obey dynamic metadata, see 5941e1f.  This felt like the least intrusive way to ensure that agents had the full machine state during reconciliation.

I did find a strange quirk. 
When the metadata is updated such that an instance of a global unit needs to be stopped fleetctl list-units displays instances of the unit incorrectly for ~10 seconds

The only relevant log appears to be
```
│Mar 01 01:41:07 core-02 fleetd[7113]: INFO manager.go:143: Triggered systemd unit test.service stop: job=24045
01 01:41:07 core-02 fleetd[7113]: INFO manager.go:273: Removing systemd unit test.service
01:41:07 core-02 fleetd[7113]: INFO reconcile.go:315: AgentReconciler completed task: type=UnloadUnit job=test.
│service reason="unit loaded but not scheduled here"
```

The units running on the other nodes appear to disappear and/or flap.  This is an output from fleetctl list-units every 2 seconds.  test.service is a global unit.

```
UNIT            MACHINE                         ACTIVE  SUB
test.service    1d2dca9f.../172.17.8.101        active  running
test.service    5f85299b.../172.17.8.103        active  running
test.service    c39e8d31.../172.17.8.102        active  running
UNIT            MACHINE                         ACTIVE  SUB
test.service    1d2dca9f.../172.17.8.101        active  running
test.service    5f85299b.../172.17.8.103        active  running
test.service    c39e8d31.../172.17.8.102        active  running
```
Change the metadata so that 1d2dca9f is no longer eligible to run the global unit.
```
UNIT            MACHINE                         ACTIVE  SUB
test.service    1d2dca9f.../172.17.8.101        active  running
test.service    5f85299b.../172.17.8.103        active  running
test.service    c39e8d31.../172.17.8.102        active  running
UNIT    MACHINE ACTIVE  SUB
UNIT    MACHINE ACTIVE  SUB
UNIT    MACHINE ACTIVE  SUB
UNIT            MACHINE                         ACTIVE  SUB
test.service    5f85299b.../172.17.8.103        active  running
UNIT            MACHINE                         ACTIVE  SUB
test.service    5f85299b.../172.17.8.103        active  running
UNIT            MACHINE                         ACTIVE  SUB
test.service    5f85299b.../172.17.8.103        active  running
UNIT            MACHINE                         ACTIVE  SUB
test.service    5f85299b.../172.17.8.103        active  running
test.service    c39e8d31.../172.17.8.102        active  running
UNIT    MACHINE ACTIVE  SUB
UNIT    MACHINE ACTIVE  SUB
UNIT    MACHINE ACTIVE  SUB
UNIT    MACHINE ACTIVE  SUB
UNIT            MACHINE                         ACTIVE  SUB
test.service    5f85299b.../172.17.8.103        active  running
UNIT            MACHINE                         ACTIVE  SUB
test.service    5f85299b.../172.17.8.103        active  running
UNIT            MACHINE                         ACTIVE  SUB
test.service    5f85299b.../172.17.8.103        active  running
test.service    c39e8d31.../172.17.8.102        active  running
UNIT            MACHINE                         ACTIVE  SUB
test.service    5f85299b.../172.17.8.103        active  running
test.service    c39e8d31.../172.17.8.102        active  running
UNIT            MACHINE                         ACTIVE  SUB
test.service    5f85299b.../172.17.8.103        active  running
test.service    c39e8d31.../172.17.8.102        active  running
```

GET /v1/state returns the same information during this period. 
The actual units on the other machines do not actually exit or restart.  The same thing happens in master if you change the unit file and restart the service so I don't think it is related to there changes.

epipho:
@bcwaldon  Got stomped by work, I will be back to this this weekend.

I believe the only outstanding item at this point is docs.

epipho:
Added a few bits of documentation about the PATCH machines api. Not sure where to put the explanation on how metadata merging works. Architecture doc perhaps?

epipho:
@bcwaldon  What do we still need here?

krancour:
+1 for this

smothiki:
+1 for this

sbward:
+1

mboersma:
+1; this would solve so many problems for us.

Telvary:
+1 !

rushmorem:
@epipho @bcwaldon @jonboulle What's the status on this?

chr0n1x:
bump

SamuelMarks:
:+1:

dalbani:
It has been a while since this issue has seen some activity.
This feature would be of great use!

dalbani:
Hi @epipho, I've been running your code for a month now and it works just fine. Thanks for your contribution!
I was wondering where I could find the code where you implemented the support for TTL?

dalbani:
There's one point I was wondering about lately: the inability to set a property with an empty value.
For example, sending a PATCH request as:
`[ { "op": "add", "path": "/e7afc665c2dc4a4b82082cd654fc6555/metadata/test", "value": "" } ]`
Returns:
`{"error":{"code":400,"message":"invalid value: add and replace require a value"}}`

Aren't empty properties valid if configured via `/etc/fleet/fleet.conf`?

dongsupark:
Just a short note:
My thought about this PR is simply, "sure, why not?".
It sounds good in general. With code reviews and writing additional tests, we might merge it. I don't think anyone is against this one.
But when I tried to rebase it on top of current master, I somehow gave up. It's a little messy. Especially the newline characters at the end of every line in the document... OMG. 
If someone could rebase it for me, I would really appreciate, be happy to review the code, and test it too.

dalbani:
I can definitely rebase the code for a start.

dalbani:
Well, easier said than done, not being a GitHub expert...
I have imported this PR's code into a branch in my own repo, and made it up to date with _master_: https://github.com/dalbani/fleet/tree/dynamic-metadata
Am I supposed to submit a _new_ pull request then?

dongsupark:
@dalbani Yes, please create a new PR. That would be appropriate, as the original author doesn't seem to be active any more. I'll try to review & test that next week.
Actually what I wanted was rebasing this branch on top of master, not merging master into the branch. Anyway let's discuss about that, looking into the new PR.

dalbani:
I have rebased on top of _master_, instead of merging, and created the new PR #1642.

dongsupark:
Closed in favor of https://github.com/coreos/fleet/pull/1642. Thanks.

Issue #1072, opened 2014-12-19
fleetctl list-unit-files not fully populated for global units


bcwaldon:
I started a global unit `ping.service` in a three-node cluster:

```
$ fleetctl list-unit-files
UNIT            HASH    DSTATE      STATE   TARGET
ping.service    a5e03d5 launched    -       -

$ fleetctl list-units
UNIT            MACHINE                     ACTIVE  SUB
ping.service    c5bb3cdf.../172.17.8.102    active  running
ping.service    d368d34e.../172.17.8.103    active  running
ping.service    f4b6d7de.../172.17.8.101    active  running
```

Note that all three machines are running `ping.service`, but the `DSTATE` and `TARGET` columns are empty. We clearly cannot use the same approach as non-global units, but we can make an attempt:

```
$ fleetctl list-unit-files
UNIT            HASH    DSTATE      STATE       TARGET
ping.service    a5e03d5 launched    launched(3) global
other.service   d368d34 loaded      loaded      f4b6d7de.../172.17.8.101
```

ZeissS:
How would this look like, when one machine has a failed state? So generally, how does this behave with mixed states on different machines?

PR #1084, opened 2015-01-07
Print "global" in fleetctl list-unit-files

bcwaldon:


bcwaldon:
Related to #1072

jonboulle:
Change LGTM, but I can't remember what we decided re: backwards compatibility? Just have a note in the release notes?

bcwaldon:
The field is currently "-" for global/unscheduled units, while it's a machine ID & IP address for scheduled units. The argument here is that the field is utterly useless and misleading in the case of a global unit, so I believe it is safe to look at this as fixing a bug rather than simply a backwards-incompatible change. We'll definitely release-note this.

PR #1085, opened 2015-01-08
log: replace V-levels with Debug method

bcwaldon:
We don't utilize the multi-leveled logs, so let's just drop that overly-complex code. Use `log.Debug[f]` instead.

### DIFF
--- a/machine/coreos.go
+++ b/machine/coreos.go
@@ -170,7 +170,7 @@ func usableAddress(ip net.IP) bool {
 }
 
 func getDefaultGatewayIface() *net.Interface {
-	log.Debugf("Attempting to retrieve IP route info from netlink")
+	log.Debug("Attempting to retrieve IP route info from netlink")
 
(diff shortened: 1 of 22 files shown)


machine/coreos.go
```diff
@@ -170,29 +170,29 @@ func usableAddress(ip net.IP) bool {
 }
 
 func getDefaultGatewayIface() *net.Interface {
-	log.V(1).Infof("Attempting to retrieve IP route info from netlink")
+	log.Debugf("Attempting to retrieve IP route info from netlink")
```
jonboulle: Kinda arbitrary mix of Debugf and Debug, pal

jonboulle:
Sure, red is good. :+1:

machine/coreos.go
bcwaldon: It was a find/replace, pal. I'll fix this bug.

PR #1092, opened 2015-01-16
Finish SSH Username Support

### LINKED ISSUE #536: fleetctl: make ssh user configurable

philk:
I needed this functionality and noticed everything seemed pretty close. My last commit here makes all commands work correctly for me.

There is one caveat that I documented in using-the-client. The `journal` command executes `journalctl` which needs permission to view the journald files. I added a `--sudo` flag to the journal command as a solution but I can't say I'm extremely happy with it but it should work by default since the sudo group is passwordless.

Linking:
Updates #1076 to fix [comment][1] from #904 and should finally close #536

[1]: https://github.com/coreos/fleet/pull/904#issuecomment-55839865

### DIFF
--- a/fleetctl/journal.go
+++ b/fleetctl/journal.go
@@ -48,7 +48,6 @@ func init() {
 	cmdJournal.Flags.BoolVar(&flagFollow, "follow", false, "Continuously print new entries as they are appended to the journal.")
 	cmdJournal.Flags.BoolVar(&flagFollow, "f", false, "Shorthand for --follow")
 	cmdJournal.Flags.BoolVar(&flagSudo, "sudo", false, "Execute journal command with sudo")
-	cmdJournal.Flags.BoolVar(&flagSudo, "s", false, "Shorthand for --sudo")
 }
 
 func runJournal(args []string) (exit int) {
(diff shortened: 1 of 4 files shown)


fleetctl/journal.go
```diff
@@ -46,6 +47,8 @@ func init() {
 	cmdJournal.Flags.IntVar(&flagLines, "lines", 10, "Number of recent log lines to return")
 	cmdJournal.Flags.BoolVar(&flagFollow, "follow", false, "Continuously print new entries as they are appended to the journal.")
 	cmdJournal.Flags.BoolVar(&flagFollow, "f", false, "Shorthand for --follow")
+	cmdJournal.Flags.BoolVar(&flagSudo, "sudo", false, "Execute journal command with sudo")
+	cmdJournal.Flags.BoolVar(&flagSudo, "s", false, "Shorthand for --sudo")
```
bcwaldon: Let's not add the shortened version of the flag until we have a library to help make it less cumbersome.

bcwaldon:
LGTM after you remove the `-s` piece

philk:
Done

jonboulle:
awesome, lgtm, thanks for finishing this up!

PR #1093, opened 2015-01-17
status: fix return code for stopped units

### LINKED ISSUE #1021: `fleetctl status` exit code not consistent with systemd
The `fleetctl status $unit_name` exit code is not consistent with the exit code of `systemctl status $unit_name`. The main issue where this comes into play is stopped & failed units. `systemctl status` on a stopped/failed unit will exit with code 3. `fleetctl status` exits with code 0.

```
# cat sleep1.service 
[Unit]
[Service]
ExecStart=/usr/bin/sleep 1

# cp sleep1.service /run/systemd/system/systemd-sleep1.service

# systemctl daemon-reload

# systemctl start systemd-sleep1.service

# sleep 2

# systemctl status systemd-sleep1.service; echo $?
● systemd-sleep1.service
   Loaded: loaded (/run/systemd/system/systemd-sleep1.service; static)
   Active: inactive (dead)

Nov 03 16:26:17 i-d24d428c systemd[1]: Starting systemd-sleep1.service...
Nov 03 16:26:17 i-d24d428c systemd[1]: Started systemd-sleep1.service.
3

# fleetctl submit sleep1.service

# fleetctl start sleep1.service
Unit sleep1.service launched on 6cdbfbf8.../10.20.24.190

# sleep 2

# fleetctl status sleep1.service; echo $?
● sleep1.service
   Loaded: loaded (/run/fleet/units/sleep1.service; linked-runtime)
   Active: inactive (dead)

Nov 03 16:26:45 i-d24d428c systemd[1]: Starting sleep1.service...
Nov 03 16:26:45 i-d24d428c systemd[1]: Started sleep1.service.
0
```

iankronquist:
Fixes #1021 on github.
If the exit code of systemd is not 0, then return with that exit code.
Previous behavior was to break and return 0.

The github.com/coreos/fleet/ssh test fails on master (at least on arch linux), and my PR does nothing to change that.
I don't see any tests for status. Will any be necessary?

### DIFF
--- a/fleetctl/status.go
+++ b/fleetctl/status.go
@@ -81,7 +81,7 @@ func runStatusUnits(args []string) (exit int) {
 		}
 
 		cmd := fmt.Sprintf("systemctl status -l %s", name)
-		if exit := runCommand(cmd, uMap[name].MachineID); exit != 0 {
+		if exit = runCommand(cmd, uMap[name].MachineID); exit != 0 {
 			break
 		}
 	}


fleetctl/status.go
```diff
@@ -82,7 +82,7 @@ func runStatusUnits(args []string) (exit int) {
 
 		cmd := fmt.Sprintf("systemctl status -l %s", name)
 		if exit := runCommand(cmd, uMap[name].MachineID); exit != 0 {
```
jonboulle: Thanks for the fix; I think the better solution here would be to stop masking the return variable. So, change `if exit :=` to `if exit =`, then leave the break statement as it was. Would you mind making that change?

iankronquist:
@jonboulle Thanks for getting back to me. I made the change you suggested.

jonboulle:
@iankronquist awesome, thanks! LGTM

PR #1095, opened 2015-01-21
Unit states documentation is out of date

jonboulle:
Hi,

I noticed that https://github.com/coreos/fleet/blob/master/Documentation/states.md is out of date. It mentions the `fleetctl show-schedule` command, which doesn't seem to exist anymore:
```
$ fleetctl version
fleetctl version 0.8.3
$ fleetctl show-schedule
fleetctl: unknown subcommand: "show-schedule"
``` 
I'm not familiar enough with fleet yet to say whether any other information about valid states and transitions between them is out-of-date, but maybe it should also be reviewed.

Cheers,
Henrik


bcwaldon:
LGTM

PR #1096, opened 2015-01-22
Change fleet to fleetd in functional test docs

ecnahc515:


jonboulle:
good catch, thank you!

PR #1099, opened 2015-01-23
nspawn: report error message from poweroff

marineam:
Intermittently systemctl is reporting the following error during
poweroff:

    Warning! D-Bus connection terminated.
    Failed to wait for response: Connection reset by peer

This seems harmless but don't have a solution yet. For now at least
report the error to make the issue more obvious.


jonboulle:
lgtm, thanks

PR #1100, opened 2015-01-23
functional: configure machines with cloud-config

marineam:
In preparation for running the fleet functional tests on other platforms
this moves most instance configuration into cloud config. This is not
quite ready to merge, the network configuration is nspawn specific and
overall runs about twice as slow as before and I'm not sure why yet.
Posting for any early feedback if anyone is interested.


functional/util/config.go
```diff
@@ -0,0 +1,169 @@
+/*
```
jonboulle: Please switch to line comments, we just did this elsewhere (block comments interfere with build tags :/)

ecnahc515:
+1 I was just considering doing this myself the other day. What are the possibilities of adding a Vagrantfile for functional tests?

marineam:
Updated, depends on https://github.com/coreos/fleet/pull/1125 and https://github.com/coreos/fleet/pull/1124

The reason this change makes the tests run slower is encoding a big Go binary in YAML significantly slows down coreos-cloudinit apparently.

marineam:
I think the most practical way to leverage cloud-config but not suffer the slowdown would be to support multi-part mime as an input format: http://cloudinit.readthedocs.org/en/latest/topics/format.html#mime-multi-part-archive

But that is not a quick thing to add. Any opinion on whether the slowdown is acceptable or if the alternative of nspawn keeping the existing setup scheme and only new platforms use cloud config is preferable? The latter just introduces a larger difference in behavior between platforms.

jonboulle:
Can we compromise for now and copy the binaries in but use cloudconfig for everything else?

marineam:
@jonboulle updated, making the inline fleetd optional since I still want that for other things.

functional/platform/nspawn.go
```diff
@@ -225,46 +228,41 @@ func (nc *nspawnCluster) prepCluster() (err error) {
 	return nil
 }
 
-func (nc *nspawnCluster) prepFleet(dir, ip, sshKeySrc, fleetdBinSrc string) error {
+func (nc *nspawnCluster) prepFleet(dir string) error {
```
jonboulle: maybe `insertFleetd`

functional/platform/nspawn.go
```diff
@@ -225,46 +228,41 @@ func (nc *nspawnCluster) prepCluster() (err error) {
...
 
-	relSSHKeyDst := path.Join("opt", "fleet", "id_rsa.pub")
-	sshKeyDst := path.Join(dir, relSSHKeyDst)
-	if err := copyFile(sshKeySrc, sshKeyDst, 0644); err != nil {
-		return err
-	}
-
 	fleetdBinDst := path.Join(dir, "opt", "fleet", "fleetd")
-	if err := copyFile(fleetdBinSrc, fleetdBinDst, 0755); err != nil {
+	if err := copyFile(fleetdBinPath, fleetdBinDst, 0755); err != nil {
```
jonboulle: return copyFile

functional/util/config.go
```diff
@@ -0,0 +1,161 @@
...
+	values := configValues{
+		NetworkUnit:   networkUnit,
+		PublicKey:     string(key),
+		Fleetd:        fleetd,
+		EtcdEndpoint:  etcdEndpoint,
+		EtcdKeyPrefix: etcdKeyPrefix,
+		FleetAPIPort:  fleetAPIPort,
+	}
+
+	if err = configTemplate.Execute(dst, &values); err != nil {
```
jonboulle: return configTemplate.Execute....

jonboulle:
few tiny things, but lgtm!

bcwaldon:
Superseded by https://github.com/coreos/fleet/pull/1100

bcwaldon:
er uh https://github.com/coreos/fleet/pull/1185

PR #1104, opened 2015-01-25
fix log.Printf verb for string from %d to %s

shawnps:


bcwaldon:
@shawnps Nice catch. I tweaked your commit message and added a commit of my own: https://github.com/coreos/fleet/pull/1110

PR #1124, opened 2015-02-10
Machine ID

marineam:
These changes turn Member.ID() into the machine ID instead of an integer that just happens to relate to the IP address. This makes way for possible future improvements:

 - Simplify `ReplaceMember()` by removing the need to preserve a static IP.
 - Allow waiting for a specific machine instead of a number as `WaitForNMachines()` does now.


bcwaldon:
LGTM thanksmike

PR #1125, opened 2015-02-10
functional: ignore errors from poweroff

marineam:
A follow up to 529209534, tests are constantly failing with a dbus error
which is probably a bug in systemd or systemctl. Not worth spending lots
of time on right now so just ignore it. If the container really fails to
shutdown the wait will eventually time out instead.


bcwaldon:
@marineam LGTM

PR #1126, opened 2015-02-11
functional: do not assume a consistent machine order from fleetctl

marineam:
There is no guarantee that m0 is the first machine listed, assuming that causes this test to randomly fail when the ordering is different. Just ignore the values from fleetctl and query the id from Member.ID()

Depends on https://github.com/coreos/fleet/pull/1124


bcwaldon:
@marineam Thanks!

PR #1133, opened 2015-02-27
functional: use --sudo with fleetctl journal commands

bcwaldon:
TestUnitSSHAction is failing due to no journal output. Using recent CoreOS builds, one has to `sudo journalctl`. Adding the `--sudo` flag to `fleectl journal` addresses the problem.


bcwaldon:
No longer necessary

PR #1173, opened 2015-04-04
fleetctl: add support for etcd2's IANA ports

crawford:
This addresses half of https://github.com/coreos/fleet/issues/1171. The daemon will still need some work.


jonboulle:
LGTM

PR #1176, opened 2015-04-06
server,fleetctl: clean up duration calculation

crawford:


jonboulle:
LGTM

bcwaldon:
This needs to be reverted. The float64s are being truncated to integers: http://play.golang.org/p/7UiWF7oHju

crawford:
Good catch.

crawford:
Good catch. Eugene just flipped a table.

PR #1177, opened 2015-04-07
fleetd: try ports 2379 and 4001 by default for etcd

crawford:
Fixes https://github.com/coreos/fleet/issues/1171.

### DIFF
--- a/Documentation/using-the-client.md
+++ b/Documentation/using-the-client.md
@@ -9,7 +9,7 @@ The `fleetctl` binary is included in all CoreOS distributions, so it is as simpl
 
 ### Custom etcd Endpoint
 
-fleetctl currently requires direct communication with the etcd cluster that your fleet machines are configured to use. Use the `--endpoint` flag to override the default of `http://127.0.0.1:4001`:
+fleetctl currently requires direct communication with the etcd cluster that your fleet machines are configured to use. Use the `--endpoint` flag to override the default of `http://127.0.0.1:2379,http://127.0.0.1:4001`:
 
     fleetctl --endpoint http://<IP:PORT> list-units
 
--- a/etcd/client.go
+++ b/etcd/client.go
@@ -26,8 +26,11 @@ import (
 )
 
 const (
-	defaultEndpoint = "http://localhost:4001"
-	redirectMax     = 10
+	redirectMax = 10
+)
+
+var (
+	defaultEndpoints = []string{"http://localhost:4001", "http://localhost:2379"}
 )
 
 type Client interface {
@@ -45,7 +48,7 @@ type transport interface {
 
 func NewClient(endpoints []string, transport *http.Transport, actionTimeout time.Duration) (*client, error) {
 	if len(endpoints) == 0 {
-		endpoints = []string{defaultEndpoint}
+		endpoints = defaultEndpoints
 	}
 
 	parsed := make([]url.URL, len(endpoints))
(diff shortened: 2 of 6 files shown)


etcd/client.go
```diff
@@ -45,7 +46,7 @@ type transport interface {
 
 func NewClient(endpoints []string, transport *http.Transport, actionTimeout time.Duration) (*client, error) {
 	if len(endpoints) == 0 {
-		endpoints = []string{defaultEndpoint}
+		endpoints = strings.Split(defaultEndpoints, ",")
```
jonboulle: would rather just define defaultEndpoints as a []string

jonboulle:
LGTM, with the (discussed) caveat that it will result in suboptimal performance in scenarios where it isn't listening on both ports. Probably get an ack from @bcwaldon

etcd/client.go
crawford: fixed

bcwaldon:
Could we switch the ordering to prefer 4001 for the time-being? I believe the client is so dumb it will always start at the beginning of the list of endpoints for every request, ignoring the last known healthy node.

crawford:
Done.

bcwaldon:
@crawford Please update `fleet.conf.sample` and `Documentation/deployment-and-configuration.md` with this new behavior. grep for 4001

crawford:
@bcwaldon updated.

Documentation/using-the-client.md
```diff
@@ -9,7 +9,7 @@ The `fleetctl` binary is included in all CoreOS distributions, so it is as simpl
 
 ### Custom etcd Endpoint
 
-fleetctl currently requires direct communication with the etcd cluster that your fleet machines are configured to use. Use the `--endpoint` flag to override the default of `http://127.0.0.1:4001`:
+fleetctl currently requires direct communication with the etcd cluster that your fleet machines are configured to use. Use the `--endpoint` flag to override the default of `http://127.0.0.1:2379,http://127.0.0.1:4001`:
```
bcwaldon: The order of the default here is backwards.

functional/platform/nspawn.go
```diff
@@ -243,7 +243,7 @@ func (nc *nspawnCluster) prepFleet(dir, ip, sshKeySrc, fleetdBinSrc string) erro
 	}
 
 	cfgTmpl := `verbosity=2
-etcd_servers=["http://172.17.0.1:4001"]	
+etcd_servers=["http://172.17.0.1:2379"]
```
bcwaldon: Does this work?

bcwaldon:
@crawford so close, crawfish. LGTM after my two comments are addressed

Documentation/using-the-client.md
crawford: This is the order that fleetctl uses. (yes, it's opposite of the daemon)

functional/platform/nspawn.go
crawford: If the tests pass, isn't assumed that it did work?

Documentation/using-the-client.md
bcwaldon: Ok.

functional/platform/nspawn.go
bcwaldon: The functional tests don't run in travis.

bcwaldon: Just revert this change for now.

crawford: I see. Reverted.

bcwaldon:
LGTM

PR #1148, opened 2015-03-10
fleetctl: Allow a custom SSH port

hhoover:
This allows end users to specify a custom SSH port with fleetctl when connecting to machines.

This is code that was split out from PR #961 by @ryandub

PR #1178, opened 2015-04-07
fleetctl: Add ssh-port info to ssh runCommand func

ryandub:
#1148 added in the `ssh-port` flag to fleetctl (thanks @hhoover !!). However, there was a slight omission in `ssh.go` to add the port context to `runCommand`. This adds that info so that `--ssh-port` works as expected.


bcwaldon:
LGTM

PR #1185, opened 2015-04-11
Use cloud-config to deploy functional test containers

bcwaldon:


bcwaldon:
I haven't seen a single indeterministic failure using this new approach.

jonboulle:
marineau magic, eh? lgtm

functional/util/config.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	// sanity check that go's use of tabs didn't leak into the yaml
+	if strings.ContainsRune(cloudConfig, '\t') {
+		panic("Aagh, no! Someone got tabs in the YAML!")
+	}
+	configTemplate = template.Must(template.New("cc").Parse(cloudConfig))
+}
+
+func encodeFile(path string) (string, error) {
```
marineam: Nothing appears to be using this and is left over from my code when I was embedding fleetd directly in the yaml to support virtual machines trasparently in the same code. I don't really recommend you use that trick without need though, it *massively* slowed things down. ;-)

bcwaldon: I'll remove it. Thanks for pointing it out

PR #1192, opened 2015-04-16
document FLEETCTL_STRICT_HOST_KEY_CHECKING

visit1985:
Just found out this little detail, and thought it's worth to be documented! :stuck_out_tongue_winking_eye:

### DIFF
--- a/Documentation/using-the-client.md
+++ b/Documentation/using-the-client.md
@@ -239,6 +239,7 @@ If a machine presents a fingerprint that differs from that found in the known_ho
 
 Disable the storage of fingerprints with `--strict-host-key-checking=false`, or change the location of your fingerprints with the `--known-hosts-file=<LOCATION>` flag.
 
+Alternatively you can export the environment variable `FLEETCTL_STRICT_HOST_KEY_CHECKING=false`.
 
 # Remote fleet Access
 


scole-scea:
This is in `fleetctl --help`... (Along with plenty more.) I.e.:
```
$ fleetctl --help
NAME:
	fleetctl - fleetctl is a command-line interface to fleet, the cluster-wide CoreOS init system.

USAGE: 
	fleetctl [global options] <command> [command options] [arguments...]

VERSION:
	0.10.0

COMMANDS:
	cat		Output the contents of a submitted unit
	destroy		Destroy one or more units in the cluster
	fd-forward	Proxy stdin and stdout to a unix domain socket
	help		Show a list of commands or help for one command
	journal		Print the journal of a unit in the cluster to stdout
	list-machines	Enumerate the current hosts in the cluster
	list-unit-files	List the units that exist in the cluster.
	list-units	List the current state of units in the cluster
	load		Schedule one or more units in the cluster, first submitting them if necessary.
	ssh		Open interactive shell on a machine in the cluster
	start		Instruct systemd to start one or more units in the cluster, first submitting and loading if necessary.
	status		Output the status of one or more units in the cluster
	stop		Instruct systemd to stop one or more units in the cluster.
	submit		Upload one or more units to the cluster without starting them
	unload		Unschedule one or more units in the cluster.
	verify		DEPRECATED - No longer works
	version		Print the version and exit

GLOBAL OPTIONS:
	--ca-file=					Location of TLS CA file used to secure communication with the fleet API or etcd
	--cert-file=					Location of TLS cert file used to secure communication with the fleet API or etcd
	--debug=false					Print out more debug information to stderr
	--driver=API					Adapter used to execute fleetctl commands. Options include "API" and "etcd".
	--endpoint=unix:///var/run/fleet.sock		Location of the fleet API if --driver=API. Alternatively, if --driver=etcd, location of the etcd API.
	--etcd-key-prefix=/_coreos.com/fleet/		Keyspace for fleet data in etcd (development use only!)
	-h=false					Print usage information and exit
	--help=false					Print usage information and exit
	--key-file=					Location of TLS key file used to secure communication with the fleet API or etcd
	--known-hosts-file=~/.fleetctl/known_hosts	File used to store remote machine fingerprints. Ignored if strict host key checking is disabled.
	--request-timeout=3				Amount of time in seconds to allow a single request before considering it failed.
	--ssh-timeout=10				Amount of time in seconds to allow for SSH connection initialization before failing.
	--ssh-username=core				Username to use when connecting to CoreOS instance.
	--strict-host-key-checking=true			Verify host keys presented by remote machines before initiating SSH connections.
	--tunnel=					Establish an SSH tunnel through the provided address for communication with fleet and etcd.
	--version=false					Print the version and exit

Global options can also be configured via upper-case environment variables prefixed with "FLEETCTL_"
For example, "some-flag" => "FLEETCTL_SOME_FLAG"

Run "fleetctl help <command>" for more details on a specific command.
```

visit1985:
I know, but it is not explicitely stated there. People like me may search the online documentation for it. Thought it would be a good idea to document the variable name, so google knows it too.

bcwaldon:
I don't know why this would be a special case that needs documenting over something like FLEETCTL_ENDPOINT. If a user needs this functionality, I think it's ok to document the flag and the general approach to env vars (FLEETCTL_*).

Issue #1203, opened 2015-04-28
Better fleet API error messages


robszumski:
When there is an issue (normally etcd related) that prevents operation of the API, this is what you see:

```
$ fleetctl list-machines
Error retrieving list of active machines: googleapi: Error 503: fleet server currently unavailable
```

Can we:
1. Get rid of the `googleapi:` string, its confusing and actively steers you in the wrong direction for troubleshooting
2. Somehow indicate that the API isn't working correctly instead of being unavailable
3. Mention the most likely culprit, which is something to do with etcd

Proposed idea:

```
$ fleetctl list-machines
Error retrieving list of active machines from fleet API (Error 503: fleet server currently unavailable)
Possible issues:
  etcd is unhealthy and/or lost quorum
  connection can't be established to any etcd address [http://10.0.0.1:4001, http://10.0.0.2:4001]
$
```

PR #1204, opened 2015-04-29
api: be more specific when unavailable

bcwaldon:
This does not address all aspects of #1203, but it does take an important step towards directing the user to etcd. Here's what a user will see now:

```
Error retrieving list of active machines: googleapi: Error 503: fleet server unable to communicate with etcd
```


crawford:
LGTM

PR #1205, opened 2015-04-29
fleetctl: set --driver=etcd if --endpoint is etcd-like

bcwaldon:
There is one remaining wart of the switch from --driver=etcd to --driver=API. If someone had been using fleetctl with a custom endpoint like so:

```
fleetctl --endpoint http://example.com:4001 list-machines
```

They now all of a sudden get directed to the fleet API rather than etcd. Given that we can tell (based on the port) that the user wants to use etcd, the default of --driver in this case should remain "etcd".

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -262,6 +262,20 @@ func main() {
 		os.Exit(2)
 	}
 
+	visited := make(map[string]bool, 0)
+	globalFlagset.Visit(func(f *flag.Flag) { visited[f.Name] = true })
+
+	// if --driver is not set, but --endpoint looks like an etcd
+	// server, set the driver to etcd
+	if visited["endpoint"] && !visited["driver"] {
+		if u, err := url.Parse(strings.Split(globalFlags.Endpoint, ",")[0]); err == nil {
+			if _, port, err := net.SplitHostPort(u.Host); err == nil && (port == "4001" || port == "2379") {
+				log.Debugf("Defaulting to --driver=%s as --endpoint appears to be etcd", clientDriverEtcd)
+				globalFlags.ClientDriver = clientDriverEtcd
+			}
+		}
+	}
+
 	if cmd.Name != "help" && cmd.Name != "version" {
 		var err error
 		cAPI, err = getClient()


fleetctl/fleetctl.go
```diff
@@ -262,6 +262,26 @@ func main() {
...
+
+	// if --driver is not set, but --endpoint looks like an etcd
+	// server, set the driver to etcd
+	if _, ok := visited["driver"]; !ok {
+		if _, ok = visited["endpoint"]; ok {
+			if eps := strings.Split(globalFlags.Endpoint, ","); len(eps) >= 1 {
+				if u, err := url.Parse(eps[0]); err == nil {
+					if _, port, err := net.SplitHostPort(u.Host); err == nil {
+						if port == "4001" || port == "2379" {
+							log.Debugf("Defaulting to --driver=%s as --endpoint appears to be etcd", clientDriverEtcd)
```
crawford: 7 levels of indentation?

bcwaldon: deal with it

jonboulle: UMM

crawford: How does this look if it's in an anonymous function, returning after failed conditions?

fleetctl/fleetctl.go
```diff
@@ -262,6 +262,26 @@ func main() {
 		os.Exit(2)
 	}
 
+	visited := make(map[string]struct{}, 0)
+	globalFlagset.Visit(func(f *flag.Flag) { visited[f.Name] = struct{}{} })
+
+	// if --driver is not set, but --endpoint looks like an etcd
+	// server, set the driver to etcd
+	if _, ok := visited["driver"]; !ok {
```
crawford: `if  visited["driver"] {`

crawford: shit

fleetctl/fleetctl.go
```diff
@@ -262,6 +262,26 @@ func main() {
...
+		if _, ok = visited["endpoint"]; ok {
```
crawford: `if visited["endpoint"] {`

fleetctl/fleetctl.go
crawford: Or if you wrap the check in a recover() and just assume that the Split, Parse, and SplitHostPort worked.

crawford:
```go
if visited["driver"] == nil && visited["endpoint"] && u, err := url.Parse(strings.Split(globalFlags.Endpoint, ",")[0]); err == nil  {
 if _, port, err := net.SplitHostPort(u.Host); err == nil && (port == "4001" || port == "2379") {
  log.Debugf("Defaulting to --driver=%s as --endpoint appears to be etcd", clientDriverEtcd)
  globalFlags.ClientDriver = clientDriverEtcd
 }
}
```

crawford:
LGTM

PR #1212, opened 2015-05-05
Use upstream etcd client

### LINKED ISSUE #1209: Replace fleet/etcd with github.com/coreos/etcd/client
The etcd client in fleet/etcd was a bandaid for the trouble we had with go-etcd. Now that the fleet/etcd client has effectively moved into github.com/coreos/etcd/client (and improved a lot!), fleet should just switch to that new client.

bcwaldon:
Closes #1209


ecnahc515:
@bcwaldon Are you just looking for a general review of things here?

bcwaldon:
@ecnahc515 yep, that would be great

ecnahc515:
Things seem to be good. You might wait until https://github.com/coreos/etcd/pull/2913 gets merged because it fixed a bug in canceling watcher contexts that I found. Then you can update the dependency.

Otherwise, it looks good, and everything looks like it was translated properly.

xiang90:
@ecnahc515 It already gets merged into etcd/client.

ecnahc515:
Great, sounds good then.

PR #1214, opened 2015-05-06
fleetd: support the version command

wuqixuan:
Now support "fleetd version" and "fleetd --version"

Fixes #1210

PR #1218, opened 2015-05-09
fleetd: fix support for "fleetd version"

### LINKED ISSUE #1210: fleet version gave me a shock
To find out the version of fleet in our current cluster, today I typed in `fleet version`. This is the wrong command, as it must be `fleetctl version` or `fleet -version`.

Fleet seem to have started and then wanted to remove units (which I guess failed due to running as `coreos`, not `root`):

```
$ fleet version
INFO fleet.go:58: Starting fleetd version 0.9.1
INFO fleet.go:162: No provided or default config file found - proceeding without
INFO server.go:153: Establishing etcd connectivity
INFO server.go:164: Starting server components
INFO manager.go:262: Writing systemd unit companyd-presence@2.service (1154b)
INFO reconcile.go:313: AgentReconciler task failed: type=LoadUnit job=companyd-presence@2.service reason="unit scheduled here but not loaded" err=open /run/fleet/units/companyd-presence@2.service: permission denied
^CINFO fleet.go:107: Gracefully shutting down
INFO client.go:291: Failed getting response from http://localhost:4001/: cancelled
ERROR event.go:107: etcd watcher {Watch /_coreos.com/fleet/job} returned error: cancelled
INFO client.go:291: Failed getting response from http://localhost:4001/: cancelled
ERROR event.go:107: etcd watcher {Watch /_coreos.com/fleet/job} returned error: cancelled
^C^C^C^C^CERROR manager.go:147: Failed to trigger systemd unit cluster-backup.service stop: Interactive authentication required.
INFO manager.go:275: Removing systemd unit cluster-backup.service
```

Also I had to use `kill -9` from another shell, as `Ctrl + C` didn't work.

Thus I have the following questions?
1) Can we make `fleet version` work too (consistency with `fleetctl`) ?
2) Is hitting `Ctrl+C` supposed to work? It says `INFO fleet.go:107: Gracefully shutting down` after the first time, but doesn't stop.
3) Is intended that I can run multiple agents on the same machine?

marineam:
Commit 32a51814 broke fleetd, making "fleetd version" the *only* thing it
could do and causing a index error panic otherwise. Now in addition to
properly checking the number of unparsed args exit with a usage error if
given any other unknown arguments.

Simply exiting if there are any unparsed args would have been sufficient
to fix #1210 but on the other hand allowing "fleetd version" doesn't
hurt so might as well keep it.

Fixes #1214 and #1210

### DIFF
--- a/fleetd/fleet.go
+++ b/fleetd/fleet.go
@@ -48,7 +48,15 @@ func main() {
 		os.Exit(1)
 	}
 
-	if *printVersion || userset.Args()[0] == "version" {
+	args := userset.Args()
+	if len(args) == 1 && args[0] == "version" {
+		*printVersion = true
+	} else if len(args) != 0 {
+		userset.Usage()
+		os.Exit(1)
+	}
+
+	if *printVersion {
 		fmt.Println("fleetd version", version.Version)
 		os.Exit(0)
 	}


fleetd/fleet.go
```diff
@@ -48,7 +48,15 @@ func main() {
 		os.Exit(1)
 	}
 
-	if *printVersion || userset.Args()[0] == "version" {
+	args := userset.Args()
+	if len(args) == 1 && args[0] == "version" {
+		*printVersion = true
+	} else if len(args) != 0 {
```
bcwaldon: What case will trigger this condition?

marineam: `fleetd list-machines` `fleetd foo bar` and so on

bcwaldon: ah, right - forgot this was fleetd.

bcwaldon:
LGTM

marineam:
Thanks :)

PR #1221, opened 2015-05-10
fleetd: prevent to run more than one fleetd deamon

### LINKED ISSUE #1220: Fleetd should prevent to run more than one daemon
Currently, fleetd doesn't prevent to run more than one daemon, it may has some potential problem. Should create fleet.pid in /var/run/

wuqixuan:
Add /var/run/fleet.pid to prevent running more than
one fleetd deamon

Fixes #1220

### DIFF
--- a/fleetd/fleet.go
+++ b/fleetd/fleet.go
@@ -18,8 +18,10 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -35,6 +37,7 @@ import (
 
 const (
 	DefaultConfigFile = "/etc/fleet/fleet.conf"
+	PidFile           = "/var/run/fleet.pid"
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


fleetd/fleet.go
```diff
@@ -213,6 +222,33 @@ func listenForSignals(sigmap map[os.Signal]func()) {
 	}
 }
 
+func createPidFile(pidfile string) error {
+	if pidstring, err := ioutil.ReadFile(pidfile); err == nil {
+		pid, err := strconv.Atoi(string(pidstring))
+		if err == nil {
+			if _, err := os.Stat(fmt.Sprintf("/proc/%d/", pid)); err == nil {
```
scole-scea: This stale pidfile check can be spoofed if some other process now has this pid. Which makes me question the whole premise of this change: Isn't the locking provided by opening network sockets good enough to prevent multiple fleet daemons from running?

wuqixuan: @scole-scea , Thanks for your suggestion. 
Have modify to check "/proc/%d/comm" whether it's "fleetd" also. 
Actually, if one process exits, the pid is reused again, need wait a long time. And the process name check can void the other process to use the same pid. 
Please refer to my another PR #1229

wuqixuan: I have raised another PR for the new code, this one will be closed.

Issue #615, opened 2014-07-03
agent: check for existing machine-ids on startup


jonboulle:
If multiple systems in a cluster happen to have the same machine-id, all hell can break loose. It would be nice if the agent could be able to do a simple sanity check when first registering itself in the cluster; if there's an existing machine with its machine-id, it can log something useful and/or back off and try again (e.g. in case a previously-pushed machineState just hasn't TTLed yet)

steveeJ:
I strongly vote for including a sanity check to ensure unique machine-ids. I came across this due to cloned VMs and spent hours of debugging until @jonboulle pointed me in the right direction.

jasonkeene:
+1 this was a head scratcher

tixxdz:
@antrik could you take this one, and it seems there is a PR https://github.com/coreos/fleet/pull/1288

PR #1229, opened 2015-05-17
fleetd: prevent to run more than one fleetd deamon

### LINKED ISSUE #1220: Fleetd should prevent to run more than one daemon

wuqixuan:
Add /var/run/fleet.pid to prevent running more than
one fleetd deamon

Fixes #1220

### DIFF
--- a/fleetd/fleet.go
+++ b/fleetd/fleet.go
@@ -18,8 +18,10 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


ecnahc515:
I don't think it's strictly important to enforce this. It makes sense in some contexts to run multiple fleet instances, and fleet has all the options to namespace everything in etcd as well.

bcwaldon:
@wuqixuan I agree with @ecnahc515 on this. I don't think this is necessary, and will prevent legitimate use cases where multiple fleet daemons can run on the same machine. Solving https://github.com/coreos/fleet/issues/615 solves the actual issue here.

wuqixuan:
Yes, I agree.

PR #1230, opened 2015-05-17
fleetctl: Add the args check for some subcommand

wuqixuan:
Some subcommand like destroy/load/start/status/stop/submit/unload,
need to give prompt if no args for them.
Add the same prompt "One unit file must be provided" as the "cat"
subcommand is doing:
[root@localhost fleet-0.10.1]# fleetctl cat
One unit file must be provided


wuqixuan:
Before the code, destroy/load/start/status/stop/submit/unload subcommand without any args will give 0 return value. It should return error code, like -1 as "cat" is doing now. That's why to add those codes.

scole-scea:
Maybe: "At least one unit file must be specified."

wuqixuan:
Currently, "cat" subcommand give "One unit file must be provided". So other subcommand  I gave the same.

scole-scea:
Yes, the `cat` subcommand does say "one". However, the `cat` subcommand _only_ takes one unit as an argument. 

`destroy`/`load`/`start`/`status`/`stop`/`submit`/`unload` all take one or more units as arguments.

wuqixuan:
So I think we need modify the message of "cat", instead.

scole-scea:
Well, `cat` has a different command-line interface. There's no reason it's error messages shouldn't reflect that difference.

fleetctl/destroy.go
```diff
@@ -29,6 +29,10 @@ Destroyed units are impossible to start unless re-submitted.`,
 }
 
 func runDestroyUnits(args []string) (exit int) {
+	if len(args) == 0 {
+		stderr("One unit file must be provided.")
```
bcwaldon: One or more unit files must be provided

fleetctl/stop.go
```diff
@@ -52,12 +52,22 @@ func init() {
...
+		return 1
+	}
+
 	units, err := findUnits(args)
 	if err != nil {
 		stderr("%v", err)
 		return 1
 	}
 
+	if len(units) == 0 {
```
bcwaldon: I don't think we need this check. The fleetctl commands are idempotent, so it is ok to do nothing.

fleetctl/load.go
```diff
@@ -46,6 +46,10 @@ func init() {
 }
 
 func runLoadUnits(args []string) (exit int) {
+	if len(args) == 0 {
+		stderr("One unit file must be provided.")
```
bcwaldon: One or more unit files must be provided

fleetctl/start.go
```diff
@@ -54,6 +54,10 @@ func init() {
 }
 
 func runStartUnit(args []string) (exit int) {
+	if len(args) == 0 {
+		stderr("One unit file must be provided.")
```
bcwaldon: One or more unit files must be provided

fleetctl/status.go
```diff
@@ -44,6 +44,10 @@ func init() {
 }
 
 func runStatusUnits(args []string) (exit int) {
+	if len(args) == 0 {
+		stderr("One unit file must be provided.")
```
bcwaldon: One or more unit files must be provided

fleetctl/stop.go
```diff
@@ -52,12 +52,22 @@ func init() {
 }
 
 func runStopUnit(args []string) (exit int) {
+	if len(args) == 0 {
+		stderr("One unit file must be provided.")
```
bcwaldon: One or more unit files must be provided

fleetctl/submit.go
```diff
@@ -36,6 +36,10 @@ func init() {
 }
 
 func runSubmitUnits(args []string) (exit int) {
+	if len(args) == 0 {
+		stderr("One unit file must be provided.")
```
bcwaldon: One or more unit files must be provided

fleetctl/unload.go
```diff
@@ -36,6 +36,10 @@ func init() {
 }
 
 func runUnloadUnit(args []string) (exit int) {
+	if len(args) == 0 {
+		stderr("One unit file must be provided.")
```
bcwaldon: One or more unit files must be provided

wuqixuan:
@bcwaldon, are all commands or just stop need not give error if no args failing ? 
If all commands, then "cat" code need remove the check? 
```
func runCatUnit(args []string) (exit int) {
	if len(args) != 1 {
		stderr("One unit file must be provided")
		return 1
	}

Or some need check, some need not check ? 
```

mischief:
@wuqixuan please update the messages as @bcwaldon asked and then lgtm.

wuqixuan:
@mischief, actually I don't know what @bcwaldon exact suggestions are. 
Which commands are idempotent without checking, while which need to be checked ?

PR #1234, opened 2015-05-22
scripts: add SSH key replacement & removal

zvbuhl:
fleetctl-inject-ssh.sh is a good tool for granting access to the core
user, but it is unable to replace SSH keys or revoke access; this change
adds two scripts to do this.

Usage:

    `cat ~/.ssh/*pub | ./fleetctl-replace-ssh-key.sh USER`
    `./fleetctl-remove-ssh-user USER`

The above are orthogonal to:

    `cat ~/.ssh/*pub | ./fleetctl-inject-ssh.sh USER`

### DIFF
--- a/scripts/fleetctl-remove-ssh-user.sh
+++ b/scripts/fleetctl-remove-ssh-user.sh
@@ -1,0 +1,13 @@
+#!/bin/bash -x
+
+name=$1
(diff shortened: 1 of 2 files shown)


bcwaldon:
@zvbuhl thanks for the contribution! LGTM

PR #1235, opened 2015-05-24
fleetctl: Should return fail if give wrong arg to destroy or unload

wuqixuan:
Should return fail if give wrong arg to destroy or unload.

### DIFF
--- a/fleetctl/unload.go
+++ b/fleetctl/unload.go
@@ -42,6 +42,11 @@ func runUnloadUnit(args []string) (exit int) {
 		return 1
 	}
 
+	if len(units) == 0 {
+		stderr("No correct unit is found.")
+		return 1
+	}
+
 	wait := make([]string, 0)
 	for _, s := range units {
 		if !suToGlobal(s) {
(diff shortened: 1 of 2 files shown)


carmstrong:
:+1:

fleetctl/unload.go
```diff
@@ -42,6 +42,11 @@ func runUnloadUnit(args []string) (exit int) {
 		return 1
 	}
 
+	if len(units) == 0 {
+		stderr("No correct unit is found.")
```
carmstrong: Maybe "%s not found" if we have the unit name here?

wuqixuan: Hi, @carmstrong, thanks for your comments. The scenario is no correct units are found. If some are found, some cannot be found, it's another scenario, we can give another patch.

scole-scea: _correct_ still feels like the wrong word (i.e.: who is the judge?). I'd phrase this as: "The specified unit or units were not found."

bcwaldon:
I don't think I buy this one. Why would you fail if none of the units are found vs some of the units aren't found? Why is one scenario worse?

wuqixuan:
This is my fault to get the two modification to the same patch. Let's talk one by one.
1. For destroy.go  I think just adding one warning is ok. How about you?
2. For unload.go ?

dongsupark:
This is already done.
See commit 9530eedb ("fleetctl: block while destroying units to ensure units are destroyed") as well as b445699e ("fleetctl: print error message when units were not found in registry").
So I'm closing it. Thanks.

PR #961, opened 2014-10-09
fleetctl: Add restart and ssh-port functionality

ryandub:
Starts and stops a unit in one command.
Adds support to specify SSH port for SSH based commands:
* journal
* status
* ssh
* restart

refs #760
refs #932

PR #1238, opened 2015-06-01
fleetctl: Adding restart command

### LINKED ISSUE #975: Add "fleetctl restart"
Occasionally I rebuild a base Docker image and want to restart the unit, or occasionally a container gets stuck and requires a restart.

Under plain systemd I can just "systemctl restart whatever", but I routinely try that with fleetctl and fail. Sure I can stop then start, or just start if the unit is already stopped, but it's an area where fleetctl isn't symmetric with the tool it claims to emulate which throws me regularly. :)

Thanks.

hhoover:
This allows users to easily do rolling restarts of units in the cluster.

This is the squash requested in #1082, as requested. I didn't want to wait.
This closes #1082 and closes #975

### DIFF
--- a/Documentation/using-the-client.md
+++ b/Documentation/using-the-client.md
@@ -75,11 +75,16 @@ Start and stop units with the `start` and `stop` commands:
 $ fleetctl start goodbye.service
 Unit goodbye.service launched on 85c0c595.../172.17.8.102
 
+$ fleetctl restart goodbye.service
+Unit goodbye.service loaded on 85c0c595.../172.17.8.102
+Unit goodbye.service launched on 85c0c595.../172.17.8.102
+
 $ fleetctl stop goodbye.service
 Unit goodbye.service loaded on 85c0c595.../172.17.8.102
 ```
 
 If the unit does not exist when calling `start`, fleetctl will first search for a local unit file, submit it and schedule it.
+If the unit is stopped when calling `restart`, fleetctl will start it.
 
 ### Scheduling units
 
--- a/fleetctl/restart.go
+++ b/fleetctl/restart.go
@@ -1,0 +1,104 @@
+/*
+   Copyright 2014 CoreOS, Inc.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"os"
+
+	"github.com/coreos/fleet/job"
+	"github.com/coreos/fleet/log"
+	"github.com/coreos/fleet/schema"
+)
+
+var cmdRestartUnit = &Command{
+	Name:    "restart",
+	Summary: "Instruct systemd to rolling restart one or more units in the cluster.",
+	Usage:   "[--block-attempts=N] UNIT...",
+	Description: `Restarts one or more units in the cluster. If they are stopped it starts them.
+
+Instructs systemd on the host machine to stop then start the unit, deferring to systemd
+completely for any custom restart directives (i.e. ExecStop options in the unit
+file).
+
+For units which are not global, restart operations are performed synchronously,
+which means fleetctl will block until it detects that the unit(s) have
+transitioned to a stopped state and then back to a started state. This behaviour can be configured with the
+respective --block-attempts options. Restart operations on global
+units are always non-blocking.
+
+Restart a single unit:
+	fleetctl restart foo.service
+
+Restart an entire directory of units with glob matching, without waiting:
+	fleetctl restart myservice/*`,
+	Run: runRestartUnit,
+}
+
+func init() {
+	cmdRestartUnit.Flags.IntVar(&sharedFlags.BlockAttempts, "block-attempts", 0, "Wait until the units are stopped/started, performing up to N attempts before giving up. A value of 0 indicates no limit. Does not apply to global units.")
+}
+
+func runRestartUnit(args []string) (exit int) {
+	units, err := findUnits(args)
+	if err != nil {
+		stderr("%v", err)
+		return 1
+	}
+	if len(units) == 0 {
+		stderr("No units were found")
+		return 1
+	}
+	for _, u := range units {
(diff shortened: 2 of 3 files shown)


hhoover:
@bcwaldon do you mind reviewing this PR

Documentation/using-the-client.md
```diff
@@ -75,11 +75,16 @@ Start and stop units with the `start` and `stop` commands:
 $ fleetctl start goodbye.service
 Unit goodbye.service launched on 85c0c595.../172.17.8.102
 
+$ fleetctl restart goodbye.service
```
bcwaldon: Please move the restart-related documentation to the bottom of the "Start and stop units" block.

fleetctl/restart.go
```diff
@@ -0,0 +1,104 @@
+/*
+   Copyright 2014 CoreOS, Inc.
```
bcwaldon: 2015

fleetctl/restart.go
```diff
@@ -0,0 +1,104 @@
...
+				return exit
+			}
+		}
+		log.Infof("Unit(%s) was restarted.", u.Name)
+	}
+	return
+}
+
+func setUnitStateAndWait(unit schema.Unit, targetState job.JobState, blockAttempts int) (exit int) {
+	cAPI.SetUnitTargetState(unit.Name, string(targetState))
```
bcwaldon: Handle the error from this

bcwaldon:
This doesn't seem to work for global units. For an actual restart to happen, we need to guarantee a state transition from launched through loaded and back to launched. If we don't block for global units, they aren't actually guaranteed to restart, which this command is trying to provide.

Having said that, I don't think it is possible given the current data model to get what we want with global units, so I would just ignore them and warn the user that we can't restart globals.

guilhem:
Hi all,
I have a huge problem with the way you are doing restart.

`systemctl stop service && systemctl start service` != `systemctl restart service`.

Why:
If `service 2` `Requires` `service 1`, restarting `service 1` will `stop` `service 2`, than restart (stop/start) `service 1` BUT (and this is the important point), it will `start` again `service 2`.

With a simple stop/start... at the end of `service 1` "restart" process, `service 2` will be stopped...

antrik:
@guilhem I hope I'm not saying something stupid here -- but it seems to me that if all you want to do is have `systemctl --restart` executed, you could just do `fleetctl ssh <unit> systemctl restart <unit>`?... (+/- syntax)

The point about dependencies is worth consideration though.

antrik:
I believe the fact that this doesn't work for global units, is just a symptom of this approach -- imperatively launching start and stop commands (or using `systemctl restart` as in https://github.com/coreos/fleet/pull/961 ) -- being fundamentally incompatible with fleet's declare & reconcile model.

It seems to me the right way to handle this is to *declare* that we want the unit to have been started after a certain timestamp (defaulting to `now()`), and storing this intention in the registry; upon which *the reconciler* will check this condition, and restart the unit as necessary.

@jonboulle does that strike you as reasonable?

jonboulle:
@antrik I think your suggestion is eminently reasonable, but it's a fairly significant change in API. I think we could punt that to a possible v1+ feature and close this particular request with your advice in https://github.com/coreos/fleet/pull/1238#issuecomment-199939959 (perhaps adding documentation to that effect). Thoughts?

antrik:
@jonboulle well, the advice only applies to the specific use case when people just want `systemctl restart` behaviour. (And to https://github.com/coreos/fleet/pull/961/files , which just wanted to add a wrapper for that...) I still think a "proper" implementation following fleet semantics would be valuable -- especially considering how often this feature has been requested in one form or another...

You are certainly right though that this is not a trivial change; and while it sounds like an interesting task that I would like to work on, it's up to you to decide priorities.

Either way, I do agree that it's probably best to close all the outstanding PRs, as none of them use the "right" approach; and add a note to this effect in https://github.com/coreos/fleet/issues/975 .

jonboulle:
@antrik +1 on your summary, mind doing that?

antrik:
@jonboulle doing what exactly? Closing the PRs? I don't have permission for that...

tixxdz:
@antrik 
> Either way, I do agree that it's probably best to close all the outstanding PRs, as none of them use the "right" approach; and add a note to this effect in #975 .

Could you please open a new issue or a feature request based on your comment https://github.com/coreos/fleet/pull/1238#issuecomment-199947799 which will be +v1 feature

Then please a PR to add documentation based on your comment: https://github.com/coreos/fleet/pull/1238#issuecomment-199939959  and yes systemd dependencies do matter when restarting services and even the order since some services may need a socket unit to be already running...

Ping @hhoover @guilhem any comments ?

antrik:
@tixxdz I don't think we should open a separate wishlist issue, as it would essentially just be a duplicate of https://github.com/coreos/fleet/issues/975 . I will post a summary of the conclusions there; and the tags/milestones can be adjusted there as appropriate. (Not by me though, as I don't have the necessary permissions...)

tixxdz:
> @tixxdz I don't think we should open a separate wishlist issue, as it would essentially just be a duplicate of #975 . I will post a summary of the conclusions there; and the tags/milestones can be adjusted there as appropriate. (Not by me though, as I don't have the necessary permissions...)

Sounds fine, if you think that we should also add documentation based on your comment https://github.com/coreos/fleet/pull/1238#issuecomment-199939959 then please open a PR, thanks!

hhoover:
> Ping @hhoover any comments ?

This PR was opened quite a while ago. In that time I've changed companies/jobs/life. As such my toolset also changed and I no longer use fleet. At the time I just used a forked version that included this code. Use it, don't use it, do something else (like what @antrik is suggesting).

I'm not emotionally invested in this issue either way at this point and what I think doesn't really matter. It's up to the folks who maintain/use this today to decide how best to move forward. If this code helps the community, great. If it doesn't, trash it. Thanks for thinking of me though. :smile:

guilhem:
For my part I don't understand new points. To much information in so many PR for my brain ^^

My only point is that it's painful to manage unit change.
For example: have you ever try to manage an edit on a `gobal` `.mount` unit...
Quite tricky when many other services rely on it ;)

What I have in view is to be able to "load" a new version of a unit without stopping existing one. And then `systemctl restart` and converge with _systemd_ dependency tree (_systemd_ will automatically stop and restart any other unit who `Requires` it).
But from what I can see, _fleet_ isn't made for this kind of purpose and I try to migrate to another solution / design (some part may rely on _fleet_).

So for this one, do your best as you always do and it will be good for me. :)

antrik:
@guilhem so using `fleetctl ssh` with `systemctl restart` doesn't work for you? Could you please explain why?

tixxdz:
Thanks all for the comments, more details on how the feature should be implemented used are here: https://github.com/coreos/fleet/issues/975#issuecomment-211388403

Closing.

PR #1225, opened 2015-05-13
fleetctl: fix issue that status command cannot find unit name

wuqixuan:
In func runStatusUnits, fmt.Sprintf("systemctl status -l %q", name)
will get the cmd as 'systemctl status -l \"name\"', there are
double quotation marks embrace unit name. When the whole command
got passed to systemd, in my pc is version 208, systemd cannot find
the unit file as per the name because of the double quotation marks.
It gives error as below:
[root@localhost fleet-0.10.1]# fleetctl --endpoint="http://127.0.0.1:4001" 
status hello.service
   \x22hello.service\x22.service
   Loaded: not-found (Reason: No such file or directory)
   Active: inactive (dead)
The \x22 is the double quotation marks. Modify "%q" to "%s", then
the command will be passed to systemd without double quotation
marks.

Fixes #1224

PR #1252, opened 2015-06-16
Revert "fleetctl: use correct escape format in status command"

jonboulle:
This reverts commit 9b8ee4b5b67d4086de0a1cd72237f9a1cae5bf8e.

This was breaking all remote fleetctl status commands as they were
passing escaped quotes to systemctl.


jonboulle:
This is redundant with #1225 but prefers to revert the commit in question for posterity

bcwaldon:
LGTM

PR #1254, opened 2015-06-18
fleetctl: destroy using wildcards can search in the repository

### LINKED ISSUE #710: fleetctl using wildcards can be dangerous
currently using wildcards with `fleetctl destroy x` picks up the filenames from the current directory instead of whats shown in `fleetctl list-units`

this can be dangerous as the current directory does not necessarily reflect whats in list-units. 

the correct behaviour might be to use the wildcards for units that are already there or not accept wildcards at all.

wuqixuan:
Curerntly, fleetctl destroy *.service, it will search the units
in the local directory. The args with wildcards have already parsed
by golang package before be passed into runDestroyUnits. So
runDestroyUnits cannot do anything.

This patch support wildcards just search in the repository,
not from local directory. But need support a new interface. The new
interface is:
    fleetctl destroy "xx*.service".
We found that if bracket the service name, the arg will not be
parsed by golang package to local directory. Then we can do match it
with the service name in the repository.

Fixes #710

### DIFF
--- a/fleetctl/destroy.go
+++ b/fleetctl/destroy.go
@@ -14,6 +14,10 @@
 
 package main
 
+import (
+	"path"
+)
+
 var cmdDestroyUnit = &Command{
 	Name:    "destroy",
 	Summary: "Destroy one or more units in the cluster",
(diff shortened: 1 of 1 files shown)


wuqixuan:
Will pull PR by another patch.

PR #1263, opened 2015-06-24
Implement follow_only flag

### LINKED ISSUE #1166: Add ability to limit leadership to a dedicated set of machines
Currently the leadership can be taken by every agent that is running within a cluster. However it would be nice to limit the leadership to just a subset of the machines available. This could make sense for setups like ["Production Cluster with Central Services"](https://coreos.com/docs/cluster-management/setup/cluster-architectures/#production-cluster-with-central-services) where the fleet leader could be only running on machines that e.g. have `role=services` configured as metadata.

miekg:
If set this fleetd will never assume mastership in a fleet cluster.
This will cut down on the interactions with Etcd.

Note: this PR is rather *hackish*, it would be nice to know if something like will be considered for merging (if cleaned up, etc .etc.). IOW: comments welcome!

Fix #1166

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -31,6 +31,7 @@ type Config struct {
 	RawMetadata             string
 	AgentTTL                string
 	TokenLimit              int
+	DisableEngine           bool
 	VerifyUnits             bool
 	AuthorizedKeysFile      string
 }
--- a/Documentation/architecture.md
+++ b/Documentation/architecture.md
@@ -11,6 +11,10 @@ Every system in the fleet cluster runs a single `fleetd` daemon. Each daemon enc
 - The engine uses a _lease model_ to enforce that only one engine is running at a time. Every time a reconciliation is due, an engine will attempt to take a lease on etcd. If the lease succeeds, the reconciliation proceeds; otherwise, that engine will remain idle until the next reconciliation period begins.
 - The engine uses a simplistic "least-loaded" scheduling algorithm: when considering where to schedule a given unit, preference is given to agents running the smallest number of units.
 
+The reconciliation loop of the engine can be disabled with the `--disable-engine` flag. This means that
+this `fleetd` daemon will *never* become a cluster leader. If all running daemons have this setting,
+your cluster is dead; i.e. no jobs will be scheduled. Use with care.
+
 ### Agent
 
 - The agent is responsible for actually executing Units on systems. It communicates with the local systemd instance over D-Bus.
@@ -19,7 +23,7 @@ Every system in the fleet cluster runs a single `fleetd` daemon. Each daemon enc
 
 ## etcd
 
-etcd is the sole datastore in a fleet cluster. All persistent and ephemeral data is stored in etcd: unit files, cluster presence, unit state, etc. 
+etcd is the sole datastore in a fleet cluster. All persistent and ephemeral data is stored in etcd: unit files, cluster presence, unit state, etc.
 
 etcd is also used for all internal communication between fleet engines and agents.
 
--- a/Documentation/fleet-scaling.md
+++ b/Documentation/fleet-scaling.md
@@ -5,6 +5,7 @@ minimizing the load it puts on etcd. This is true for reads, writes, and
 watches.
 
 ## Known issues
+
 Currently when fleet schedules a job *all* `fleetd`s are woken up (via a watch)
 and then do a recursive GET on the Unit file in etcd to figure out if it should
 schedule a job. This is a very expensive operation.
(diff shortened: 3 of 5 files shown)


bcwaldon:
@miekg In what scenario has this helped you so far? I worry that a cluster will degrade to only those nodes that are --follow-only and the cluster will mysteriously stop making any progress.

miekg:
[ Quoting <notifications@github.com> in "Re: [fleet] Implement follow_only f..." ]
>@miekg In what scenario has this helped you so far? 

We basically have two type of nodes in our clusters: worker and admin nodes. The
worker nodes come and go (in large numbers) and these should just be used to
schedule jobs. We never want them to become leaders. The admin nodes control
the entire cluster and we just want one of those to become the fleet leader.

> I worry that a cluster will
> degrade to only those nodes that are --follow-only and the cluster will
> mysteriously stop making any progress.

So, this is why I'm also adding metrics to fleet. Without admin nodes our
cluster is useless anyway (because no etcd - among other things), so fleet will
then be also broken, regardless of where the fleet leader actually is.




/Miek

--
Miek Gieben

bcwaldon:
@miekg ok, that makes sense. Could you start a "scaling" document that helps people understand how to use these features to scale out fleet?

engine/engine.go
```diff
@@ -58,7 +58,7 @@ func New(reg *registry.EtcdRegistry, lManager lease.Manager, rStream pkg.EventSt
 	}
 }
 
-func (e *Engine) Run(ival time.Duration, stop chan bool) {
+func (e *Engine) Run(ival time.Duration, follower bool, stop chan bool) {
```
bcwaldon: Shouldn't the Engine simply not run if follow-only == true?

config/config.go
```diff
@@ -30,6 +30,7 @@ type Config struct {
 	Verbosity               int
 	RawMetadata             string
 	AgentTTL                string
+	FollowOnly              bool
```
bcwaldon: How about we simplify this down to `NoScheduler` or something?

miekg: [ Quoting <notifications@github.com> in "Re: [fleet] Implement follow_only f..." ]
>> @@ -30,6 +30,7 @@ type Config struct {
>>  	Verbosity               int
>>  	RawMetadata             string
>>  	AgentTTL                string
>> +	FollowOnly              bool
>
>How about we simplify this down to `NoScheduler` or something?

NoLeader? But I dislike the "No" part, so "NeverLeader" ?

/Miek

--
Miek Gieben

miekg:
[ Quoting <notifications@github.com> in "Re: [fleet] Implement follow_only f..." ]
>@miekg ok, that makes sense. Could you start a "scaling" document that helps people understand how to use these features to scale out fleet?

Ok, will do.

miekg:
Ok, reworked PR to not start the scheduler at all and renamed the flag to `no_scheduler`. PTAL.

miekg:
Friendly ping @bcwaldon

mwitkow-io:
@mischief, can you take a look at this? This change is crucial in making fleet scale to ~1000s of machines.

miekg:
Dear fleet devs :) @jonboulle @mischief ... can we merge this as well?

jonboulle:
What made you choose "scheduler" rather than "engine"?

miekg:
wrt the naming in the PR? Wasn't I conscious choice, felt more natural to use 'scheduler', but 'engine' might be better describing what it does code wise. Or maybe 'no_master' . Happy to make rename it to whatever you like best.

jonboulle:
Oh, I see, @bcwaldon originally suggested it @ https://github.com/coreos/fleet/pull/1263#discussion_r34280368 . Personally I would find `--disable-engine` much more intuitive and consistent with the documentation/code ([e.g.](https://github.com/coreos/fleet/blob/master/Documentation/architecture.md#engine))

miekg:
I have no problem making that change.

jonboulle:
I think I'd like to see 
a) change to `--disable-engine`
b) an explanation in the architecture document
c) some **big bold warnings** in all of the relevant documentation (to try mitigate [this concern](https://github.com/coreos/fleet/pull/1263#issuecomment-119803962))

Other than that, the implementation seems fine, fairly straightforward. Thanks!

miekg:
Ok. Change to --disable-engine. Added docs, squashed the commits.
PTAL.

jonboulle:
Thanks! LGTM.

miekg:
Cool! Thank you.

PR #1264, opened 2015-06-24
Scale hack

miekg:
Comment out the use of Etcd watches and only do time-based wake-ups. Add jitter to not have the whole cluster wake-up at the same time (make it even more less likely).

Note: this PR is rather *hackish*, it would be nice to know if something like will be considered for merging (if cleaned up, etc .etc.). IOW: comments welcome!

### DIFF
--- a/pkg/reconcile.go
+++ b/pkg/reconcile.go
@@ -15,6 +15,7 @@
 package pkg
 
 import (
+	"math/rand"
 	"time"
 
 	"github.com/coreos/fleet/Godeps/_workspace/src/github.com/jonboulle/clockwork"
@@ -54,7 +55,7 @@ type reconciler struct {
 }
 
 func (r *reconciler) Run(stop chan bool) {
-	trigger := make(chan struct{})
+	//	trigger := make(chan struct{})
 	go func() {
 		abort := make(chan struct{})
 		for {
@@ -62,13 +63,14 @@ func (r *reconciler) Run(stop chan bool) {
 			case <-stop:
 				close(abort)
 				return
-			case <-r.eStream.Next(abort):
-				trigger <- struct{}{}
+				// Kill with fire.
+				//			case <-r.eStream.Next(abort):
+				//				trigger <- struct{}{}
 			}
 		}
 	}()
 
-	ticker := r.clock.After(r.ival)
+	ticker := r.clock.After(r.ival + jitter(r.ival))
 
 	// When starting up, reconcile once immediately
 	log.Debug("Initial reconciliation commencing")
@@ -80,14 +82,16 @@ func (r *reconciler) Run(stop chan bool) {
 			log.Debug("Reconciler exiting due to stop signal")
 			return
 		case <-ticker:
-			ticker = r.clock.After(r.ival)
+			ticker = r.clock.After(r.ival + jitter(r.ival))
 			log.Debug("Reconciler tick")
-			r.rFunc()
-		case <-trigger:
-			ticker = r.clock.After(r.ival)
-			log.Debug("Reconciler triggered")
 			r.rFunc()
 		}
 	}
 
+}
+
+// jitter returns randomized jitter in the internal [0, ival/2).
+func jitter(ival time.Duration) time.Duration {
+	j := rand.Int63n(int64(ival / 2))
+	return time.Duration(j)
 }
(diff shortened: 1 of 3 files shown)


aledbf:
@miekg how many units do you need/have in the cluster to see a difference with this change?

miekg:
[ Quoting <notifications@github.com> in "Re: [fleet] Scale hack (#1264)..." ]
>@miekg how many units do you need/have in the cluster to see a difference with this change?

We still need to test this ourselves. Hopefully this will happen this week.

Currently fleet pounds etcd into the ground with 500+ machines/fleetd's and 500+
unit files.

/Miek

--
Miek Gieben

aledbf:
@miekg I saw this: after ~100 units per node (5 node cluster) it was not possible to schedule more units or handle the failure of any node. And if you use the `X-Conflicts` in more than 100 units is not possible to reach 300 units in the cluster. Even `fleetctl list-units` takes too long (changing the tokenLimit helps but just in that operation)

miekg:
[ Quoting <notifications@github.com> in "Re: [fleet] Scale hack (#1264)..." ]
>@miekg I saw this: after ~100 units per node (5 node cluster) it was not possible to schedule more units or handle the failure of any node. And if you use the `X-Conflicts` in more than 100 units is not possible to reach 300 units in the cluster. Even `fleetctl list-units` takes too long (changing the tokenLimit helps but just in that operation)

Hmmm, that is a good datapoint. With our current setup we were able to go to 535
machines and unit files (need to check if we use X-Conflicts there - I don't
think so, however).

We've also instrumented Etcd with Prometheus counters and we saw 60 watches/s
being set and we expect most of them were from fleet, so cutting back on those
will likely help. But as said, we're going to do tests for this as well.

aledbf:
@miekg I found a difference: if you use etcd v0.4.x is possible to reach 600 units but in v2.0.x no

miekg:
[ Quoting <notifications@github.com> in "Re: [fleet] Scale hack (#1264)..." ]
>@miekg I found a difference: if you use etcd v0.4.x is possible to reach 600 units but in v2.0.x no

WTF?! :-) Older Etcd's scale better :) We saw stability issues with etcd < v2
so there is no going back to that version. But it is a nice data point.

In general it is good to limit the number of recursive queries and watches, esp.
how fleet currently is doing it: which is basically O(n^2)

/Miek

--
Miek Gieben

aledbf:
@miekg I saw this "issue" with fleet after I upgrade etcd to v2 (always using alpha).

miekg:
How do we go forward with my changes? Should I massage them into code that can be merged? Is killing the watches something you want?

jonboulle:
I'm receptive to this, the watches were an optimisation to improve responsiveness but I'm more inclined to trade off for scale here. @bcwaldon @mischief ?

bcwaldon:
I'm ok with the general approach as a stop-gap for now, but I would actually like to consider rearchitecting fleet to provide per-machine schedules and get away from using TTLs for presence altogether. Those two pieces should help fleet scale much further than it can today.

miekg:
[ Quoting <notifications@github.com> in "Re: [fleet] Scale hack (#1264)..." ]
>I'm ok with the general approach as a stop-gap for now, 

Ack. I will cleanup this branch to make it ready for it to be considered for
merging.

> but I would actually
>like to consider rearchitecting fleet to provide per-machine schedules and get
>away from using TTLs for presence altogether. Those two pieces should help
>fleet scale much further than it can today.

yes! Note that with the current hacks we are able to scale to 1000-2000 machines
(with a 5 node etcd cluster). That's a pretty large cluster by any means. 

/Miek

--
Miek Gieben

miekg:
Is this still relevant? I.e. should I clean it up and ask for a "PTAL" ?

mwitkow-io:
@bcwaldon or @mischief, can you guys take a look a this? we're actually running like this in prod and everything is fine :)

miekg:
Let me clean it, hide it behind a flag and add some text to the relevant docs.

jonboulle:
Right, I'd be receptive to a cleaner version of this that's configurable.

miekg:
I am going to cleanup this PR and prolly resubmit as a different. I.e. hide it behind a flag +  docs etc.
Work should be fairly minimal so I should have something later today.

jonboulle:
SGTM

jonboulle:
Fixed by #1406

PR #1267, opened 2015-06-24
fleetctl: sanity check unit names in submit/load/start and disallow templates in load/start

mischief:


fleetctl/fleetctl.go
```diff
@@ -845,3 +845,28 @@ func suToGlobal(su schema.Unit) bool {
 	}
 	return u.IsGlobal()
 }
+
+// checkUnitNames sanity checks unit names
+func checkUnitNames(names []string, allowTemplate bool) error {
+	units := make([]string, 0, len(names))
```
mwhooker: forgive me if I'm misunderstanding the intention, but I think this should be rewritten as 

```go
    func checkUnitNames(names []string, allowTemplate bool) error {
	for _, un := range units {
		uni := unit.NewUnitNameInfo(unitNameMangle(un))
		// snip...
```

mischief: sounds good. my original intention was to drop unit names that failed validation from the slice, but i feel like it's better to abort if any unit name fails checks.

unfortunately the combination of `unit.NewUnitNameInfo` and `unitNameMangle` is not terribly helpful, becuase `unit.NewUnitNameInfo` only returns nil if the unit is missing a `.` and `unitNameMangle` is appending `.service` for us. more checks are probably needed in `unit.NewUnitNameInfo`.

mwhooker: yeah, might be best to write some tests that demonstrate which names fail the check and which succeed

mischief:
closing in favor of checking for this in the fleet api.

PR #1271, opened 2015-06-26
fleetctl: Support truthy values in boolean unit options

### LINKED ISSUE #1108: Support truthy values in boolean unit options
The `Global` option is lower-cased then checked directly against `true` [0], while systemd units accept "truthy" values (yes, on, 1, t) [1]. fleet should follow suit and accept the same truthy values when parsing boolean options.

[0] https://github.com/coreos/fleet/blob/d76e5ade2f285ff9ec3d0fbe57fc17f5af94d61a/job/job.go#L116
[1] http://www.freedesktop.org/software/systemd/man/systemd.unit.html#Description

wuqixuan:
Support truthy values in boolean unit options(yes, on, 1, t)

Fixes #1108

### DIFF
--- a/job/job.go
+++ b/job/job.go
@@ -113,7 +113,8 @@ func (u *Unit) IsGlobal() bool {
 	}
 	// Last value found wins
 	last := values[len(values)-1]
-	return strings.ToLower(last) == "true"
+	lastLower := strings.ToLower(last)
+	return lastLower == "true" || lastLower == "yes" || lastLower == "1" || lastLower == "on" || lastLower == "t"
 }
 
 // NewJob creates a new Job based on the given name and Unit.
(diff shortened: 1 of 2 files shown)


job/job.go
```diff
@@ -113,7 +113,8 @@ func (u *Unit) IsGlobal() bool {
 	}
 	// Last value found wins
 	last := values[len(values)-1]
-	return strings.ToLower(last) == "true"
+	lastLower := strings.ToLower(last)
+	return lastLower == "true" || lastLower == "yes" || lastLower == "1" || lastLower == "on" || lastLower == "t"
```
bcwaldon: Break this logic out into a helper function and test it independently of Unit.IsGlobal

PR #1281, opened 2015-06-30
scripts/bump-release: print commands as they are run, fail immediately on errors

mischief:


scripts/bump-release
```diff
@@ -3,6 +3,9 @@
 # $1 = version string (e.g. 0.8.0)
 # $2 = remote configured to track github.com/coreos/fleet (defaults to "origin" if unset)
 
+# print commands, fail hard and fast
+set -xeo pipefail
```
crawford: Why `-o pipefail`?

mischief: not needed. script's already called with bash -e anyway, so closing this.

PR #1287, opened 2015-07-02
agent: do not call wg.Done before wg.Add

mischief:


mischief:
fixes race seen in https://github.com/coreos/fleet/pull/1287

jonboulle:
weird that this didn't come up earlier. LGTM

PR #1291, opened 2015-07-04
fleetctl: Remove unit content from etcd

### LINKED ISSUE #1290: Fleetctl destroy unit does not remove it from etcd registry
Fleetctl destroy unit does not remove the unit file content from etcd registry(/_coreos.com/fleet/unit/XXXX). It will cause so many junk to leave in the etcd.

wuqixuan:
Fleetctl destroy unit does not remove the unit file content from
etcd registry(/_coreos.com/fleet/unit/XXXX). It will cause so
many junk to leave in the etcd. The modification removes it from
etcd when destroying unit.

Fixes #1290

### DIFF
--- a/registry/job.go
+++ b/registry/job.go
@@ -290,7 +290,12 @@ func (r *EtcdRegistry) DestroyUnit(name string) error {
 	opts := &etcd.DeleteOptions{
 		Recursive: true,
 	}
-	_, err := r.kAPI.Delete(r.ctx(), key, opts)
+	u, err := r.Unit(name)
+	if err != nil {
+		log.Warningf("r.Unit error, name=%s\n", name)
+		u = nil
+	}
+	_, err = r.kAPI.Delete(r.ctx(), key, opts)
 	if err != nil {
 		if isEtcdError(err, etcd.ErrorCodeKeyNotFound) {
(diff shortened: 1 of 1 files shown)


registry/job.go
```diff
@@ -300,6 +305,17 @@ func (r *EtcdRegistry) DestroyUnit(name string) error {
 	}
 
 	// TODO(jonboulle): add unit reference counting and actually destroying Units
+	if u != nil { //delete unit
+		path := u.Unit.Hash().String()
+		key = r.prefixed(unitPrefix, path)
+		_, err = r.kAPI.Delete(r.ctx(), key, opts)
```
bcwaldon: This isn't safe as there may be two units that hash to the same key in etcd.

wuqixuan:
Yes, it can be.  Then we can't store the unit files using hash value, right? Maybe we can store it in the job folder. When we delete the job folder, delete the unit file at the same time. How do you think?

bcwaldon:
I'd prefer to create a garbage collector that cleans up the etcd keyspace periodically. We can safely check for job models that refer to the unit hash before deleting it.

dongsupark:
Closed in favor of https://github.com/coreos/fleet/pull/1510

PR #1292, opened 2015-07-05
fleetctl: Give DEBUG print if need wait for launch when start unit

### LINKED ISSUE #1146: fleetctl: friendlier messages for invalid X-Fleet sections
I understand that globbing patterns are invalid in a MachineOf option. However I would expect fleetctl to print an error or some more information about what is happening. Currently if I run `fleetctl start foo-follower.service` with a file like below, it just hangs with no output and nothing is printing to the fleet logs either. It's difficult to know what is happening.

```
[X-Fleet]
MachineOf=foo*.service
```

I would expect a friendlier output. Something like: "Unit requirements not met, waiting to schedule unit...", or "Waiting for foo*.service...", etc.

wuqixuan:
When start unit, if [X-Fleet] give peer by "MachineOf=", the
fleetctl cannot know the exact reason. Only fleetd check the peer.
But still we can give a DEBUG print to inform launch is happing.
It's not the problem that the peer is not existing. After the
peer is running, the fleetctl will go on to the end. So just
add a debug print if --debug=true is given to fleetctl.

Fixes #1146

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -779,7 +779,12 @@ func assertUnitState(name string, js job.JobState, out io.Writer) (ret bool) {
 		log.Warningf("Error retrieving Unit(%s) from Registry: %v", name, err)
 		return
 	}
-	if u == nil || job.JobState(u.CurrentState) != js {
+	if u == nil {
+		log.Warningf("Unit %s not found", name)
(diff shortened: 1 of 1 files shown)


bcwaldon:
LGTM, thanks @wuqixuan

PR #1294, opened 2015-07-06
fleetd: support operators in metadata

### LINKED ISSUE #1143: Operators in Metadata
Would the authors consider allowing operators in metadata?

ie:

[X-fleet]
MachineMetadata=ram>=2048

This + way to update metadata easier = dynamic metadata, awesome? lame? opinions?

thanks

wuqixuan:
If define metadata in fleet conf, such as "ram=1024", we can
define the operator in [X-Fleet] unit as below:
[X-Fleet]
MachineMetadata=ram>=2048
The operators have been supported: "<=", ">=", "!=", "<", ">", "="
If the operatior are "<=", ">=", "!=", "<", ">", the value should
be integer, otherwise, the unit will never be launched.

Fixes #1143

### DIFF
--- a/Documentation/unit-files-and-scheduling.md
+++ b/Documentation/unit-files-and-scheduling.md
@@ -125,6 +125,14 @@ MachineMetadata=region=us-west-1
 
 This would allow a machine to match just one of the provided values to be considered eligible to run.
 
+`MachineMetadata` also support relational operators, including `<=`, `>=`, `<`, `>` and `!=`:
+
+```
+[X-Fleet]
+MachineMetadata=ram<1024
+```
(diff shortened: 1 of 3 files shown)


bcwaldon:
1. I'd like to see the operator `==` defined (equivalent to the existing `=`)
2. we need significant testing of these new operators
3. both the API and fleetctl need to validate metadata against these new rules

mwitkow-io:
This would be awesome :)

osterman:
Really looking forward to this!

dongsupark:
Closed in favor of https://github.com/coreos/fleet/pull/1632. Thanks.

PR #1295, opened 2015-07-06
fleetctl: support overwrite command

### LINKED ISSUE #760: Redeploy unit in place
It would be quite nice to be able to replace an existing unit in one go, if a change must be made, preserving its location.

* `fleetctl submit some.unit.service`
* `fleetctl run some.unit.service`
* _edit some.unit.service_
* `fleetctl reload some.unit.service`

Fleet should replace the unit conf in-place, stop the running unit, and restart it using the new conf.

wuqixuan:
If unit file modified, can use "overwrite" command to update.
First stop and remove the old unit, and submit the new unit and restore
the status as the previous one.

Fixes #760

### DIFF
--- a/fleetctl/overwrite.go
+++ b/fleetctl/overwrite.go
@@ -1,0 +1,114 @@
+package main
+
+import (
+	"os"
+
+	"github.com/coreos/fleet/job"
+)
+
+var cmdOverwriteUnit = &Command{
+	Name:    "overwrite",
+	Summary: "Overwrite one or more units in the cluster",
+	Usage:   "UNIT...",
+	Description: `Overwrite one or more running or submitted units from the cluster.
+
+Act as a procedure of stop-->destroy-->commit-->start(if needed) on unit(s)`,
+
+	Run: runOverwriteUnits,
+}
+
+func runOverwriteUnits(args []string) (exit int) {
+	for _, v := range args {
+		v = maybeAppendDefaultUnitType(v)
+		name := unitNameMangle(v)
+
+		u, err := cAPI.Unit(name)
+		if err != nil {
+			stderr("error retrieving Unit(%s) from Registry: %v", name, err)
+			return 1
+		}
+		if u == nil {
+			stdout("Unit(%s) in can not be found in Registry, nothing to do with it", name)
+			continue
+		}
+
+		hash_mismatch := warnOnDifferentLocalUnit(v, u)
+		if hash_mismatch == false {
+			stdout("Nothing different between Unit(%s) in registory and local unit file, just skip", name)
+			continue
+		}
+
+		stopping := make([]string, 0)
+		stopping = append(stopping, u.Name)
+		if job.JobState(u.CurrentState) == job.JobStateLaunched || suToGlobal(*u) {
+			stdout("Stop Unit(%s) first, so that we can overwrite unit file", u.Name)
+			cAPI.SetUnitTargetState(u.Name, string(job.JobStateLoaded))
+			errchan := waitForUnitStates(stopping, job.JobStateLoaded, 0, os.Stdout)
+			for err := range errchan {
+				stderr("Error waiting for units: %v", err)
+				exit = 1
+			}
+		}
+
+		err = cAPI.DestroyUnit(name)
+		if err != nil {
+			continue
+		}
+
+		recreating := make([]string, 0)
+		recreating = append(recreating, v)
+		stdout("Recreating unit(%s)", v)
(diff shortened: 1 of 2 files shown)


fleetctl/overwrite.go
```diff
@@ -0,0 +1,114 @@
...
+	"os"
+
+	"github.com/coreos/fleet/job"
+)
+
+var cmdOverwriteUnit = &Command{
+	Name:    "overwrite",
+	Summary: "Overwrite one or more units in the cluster",
+	Usage:   "UNIT...",
+	Description: `Overwrite one or more running or submitted units from the cluster.
```
bcwaldon: Replace the description with this:

Overwrite stops (if needed), unload (if needed), and destroys, then submits, loads (if needed) and starts (if needed) the unit after replacing the unit file with that found on the local filesystem

fleetctl/overwrite.go
```diff
@@ -0,0 +1,114 @@
...
+		v = maybeAppendDefaultUnitType(v)
+		name := unitNameMangle(v)
+
+		u, err := cAPI.Unit(name)
+		if err != nil {
+			stderr("error retrieving Unit(%s) from Registry: %v", name, err)
+			return 1
+		}
+		if u == nil {
+			stdout("Unit(%s) in can not be found in Registry, nothing to do with it", name)
```
bcwaldon: Remove the first "in"

fleetctl/overwrite.go
```diff
@@ -0,0 +1,114 @@
...
+			continue
+		}
+
+		hash_mismatch := warnOnDifferentLocalUnit(v, u)
+		if hash_mismatch == false {
+			stdout("Nothing different between Unit(%s) in registory and local unit file, just skip", name)
```
bcwaldon: Replace "just skip" with "ignoring"

bcwaldon:
@wuqixuan you don't actually have to wait for units to transition through the loaded state on teardown or startup if that isn't the desired final state. You can trust that fleet will drive units towards the desired state, regardless of what intermediate state there may be.

politician:
@bcwaldon Can you merge this? The changes you've requested (last year) just deal with copy, and not functionality.  I'm sure the people who really want/need restart (overwrite) capabilities can deal with some bad text.

esatterwhite:
As a user a fleet, I think the primary use case here would be rolling update that didn't have to bring down each instance of a running container. If I have `service@1`, `service@2`, and `service@3` I don't want to have to stop all of them. I'd really want to update the unit file inplace, And then restart them 1 by 1. And honestly, I'd be happy to do the restarting manually, that isn't difficult to do. 

But having  to `stop` & `destroy` all of the running before an update can be applied is a bit of a show stopper.

sstarcher:
+1

jonboulle:
To recap:
- the fleet agent already supports "replacing" units of a given name by comparing their current local state with that of the unit in the registry: https://github.com/coreos/fleet/blob/81d6c4a0f77f65b987b6223e5a71d966e1cf1d20/agent/reconcile.go#L256
- hence, there's no need for `fleetctl` to do individual stop/destroy/load operations, as this PR currently does (indeed, that would not really be any different to just doing a `fleetctl stop foo && fleetctl destroy foo && fleetctl load foo && fleetctl start foo`(
- what we need to do is allow `fleetctl` to overwrite a unit in-place - the agent will then do the necessary work to actually replace that unit on the system
- in terms of user experience with fleetctl, it could behave similarly to how start/stop do today, polling by default (unless `--no-block` is passed): https://github.com/coreos/fleet/blob/81d6c4a0f77f65b987b6223e5a71d966e1cf1d20/fleetctl/start.go#L77 
but instead of only testing for the unit to reach a given state, it would also need to check that the unit's hash reaches the desired value (of the replacement)

kayrus:
I've implemented bash based [POC](https://gist.github.com/kayrus/f13c84092a7fef18c8c4) which works with the fleet v0.11.5.

Didn't test it with templates.

antrik:
See also https://github.com/coreos/fleet/pull/1442 for a slightly different approach (regarding user interface) for this functionality, which seems worth considering.

jonboulle:
Per OOB discussion, let's proceed with a `--replace` (or `--overwrite`) flag for start/submit/load. It'll probably look quite similar to #1442 but needs a bit more work to ensure we can sanely block (looking for the new hash) etc

tixxdz:
Closing this in favor of https://github.com/coreos/fleet/pull/1509

@wuqixuan thank you for the patch, could you followup or test #1509 , thanks!

PR #1300, opened 2015-07-08
test: fix bug about not able to run gofmt

### LINKED ISSUE #1299: test script has an issue
test script has an issue, which cause not able to run gofmt for all package.

wuqixuan:
Now test can run gofmt and go vet

Fixes #1299

### DIFF
--- a/build
+++ b/build
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash  
 
 ORG_PATH="github.com/coreos"
 REPO_PATH="${ORG_PATH}/fleet"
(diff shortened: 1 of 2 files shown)


test
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash
```
wuqixuan: Remove -e to avoid exiting from "go test" failing.

test
```diff
@@ -35,22 +35,33 @@ fi
...
+testRes=$(go test $COVER $@ $TEST)
+if [ $? != 0 ]; then
+	echo -e "go test checking failed:\n\n${testRes}"
+	RET=255
+else
+	echo -e "go test succeeds:\n\n${testRes}"
+fi
 
-echo "Checking gofmt..."
+echo -e "\nChecking gofmt..."
```
wuqixuan: Here adding "\n" is just for goodlooking.

build
```diff
@@ -16,9 +16,16 @@ eval $(go env)
 if [ ${GOOS} = "linux" ]; then
 	echo "Building fleetd..."
 	CGO_ENABLED=0 go build -o bin/fleetd -a -installsuffix netgo -ldflags '-s' ${REPO_PATH}/fleetd
```
wuqixuan: Because test script remove "-e", here need to check the result for each build. Otherwise, the test script will go ahead even if build failing.

wuqixuan:
I cannot see the reason of the failing. When I click the details button, it shows the blank page. 
I tested it in my board and passed.

wuqixuan:
@mischief @bcwaldon, do you have another comment for my latest modification?

wuqixuan:
Gently ping @jonboulle . : )

jonboulle:
Ehhhhhh, I'm -1 on this because it's a pattern we use pretty generally in our other projects and I'm not sure the existing behaviour is that bad

dongsupark:
I'll close this one due to a lack of activity. Honestly it's hard for me to understand what was the original problem.

PR #1303, opened 2015-07-12
Add '/fleet/v1' to all calls

meyskens:
The documentation can be confusing when not knowing `/fleet/v1` has to be added in front of all requests.


robszumski:
LGTM but I'm interested why @bcwaldon didn't include this in the first place

wuqixuan:
LGTM

PR #1316, opened 2015-07-23
consistent reads

mischief:
this fixes a race encountered when fleetd does a PUT followed by GET but does not use `quorum=true` for the get, resulting in a 404 from the local etcd.

/cc @bcwaldon @jonboulle @marineam


marineam:
LGTM

PR #1327, opened 2015-08-13
fleetctl: call checkUnitState when unit exists on disk

mischief:
in 1816c53ab58ca430914637cefc23985353b72d31 i missed a case where
checkUnitState was skipped by 'continue' in the case that the unit
exists on disk.


vcaputo:
lgtm pending green

PR #1337, opened 2015-08-27
Godeps: bump etcd deps

### LINKED ISSUE #1336: etcd/client: use new version that randomizes clients
A large number of fleetd members pointed at the same etcd list will put all load on the first etcd member in that list. We need to fix the etcd client to randomize this list which depends on: https://github.com/coreos/etcd/issues/3386

mischief:
fixes #1336

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,6 +1,6 @@
 {
 	"ImportPath": "github.com/coreos/fleet",
-	"GoVersion": "go1.3.3",
+	"GoVersion": "go1.4.2",
 	"Packages": [
 		"./..."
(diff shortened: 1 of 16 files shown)


xiang90:
LGTM. But have you verified this change does fix https://github.com/coreos/fleet/issues/1336?

mischief:
@xiang90 no. don't merge yet.

mischief:
@xiang90 @philips verified the change with etcd's goreman Procfile and wireshark.

xiang90:
Thanks! Let us get it merged.

vcaputo:
1337!

PR #1341, opened 2015-09-03
fleetctl: optimize status command

### LINKED ISSUE #1340: fleetctl status slow on large number of services
We have a cluster with ~2500 services. Running `time fleetctl status <service-name>` takes many minutes. I noticed fleetctl is retrieving all Units, only to display the status on a handful of them. I hope to submit a pull request to fix this soon.

royvandewater:
Use the Unit API instead of Units API to speed up lookups on
environments with a large number of distinct services.

Fixes #1340

### DIFF
--- a/fleetctl/status.go
+++ b/fleetctl/status.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 
 	"github.com/coreos/fleet/job"
-	"github.com/coreos/fleet/schema"
 )
 
 var cmdStatusUnits = &Command{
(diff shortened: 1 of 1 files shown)


royvandewater:
When `fleetctl status <service>` is run against 2.5k services without this patch:

```shell
real    4m24.150s
user    0m0.475s
sys     0m0.155s
```

And with this patch:

```shell
real    0m2.344s
user    0m0.146s
sys     0m0.020s
```

royvandewater:
Whoops, failed the formatter, gimme a sec

royvandewater:
I think this build failure may not be me, looks like it's the same one that's failing in master.

mischief:
@royvandewater it is a race condition in the test. it was not your fault.

mischief:
@royvandewater thank you.

PR #1349, opened 2015-09-11
etcd: enable basic authentication

thepwagner:
Supported by client, boring glue ties to fleetd config.

### DIFF
--- a/Documentation/deployment-and-configuration.md
+++ b/Documentation/deployment-and-configuration.md
@@ -103,6 +103,12 @@ Provide TLS configuration when SSL certificate authentication is enabled in etcd
 
 Default: ""
 
+#### etcd_username, etcd_password
+
+Provide credentials when [authentication](https://github.com/coreos/etcd/blob/master/Documentation/authentication.md) is enabled in etcd endpoints.
+
+Default: ""
+
 #### etcd_key_prefix
 
 Keyspace path for fleet data in etcd.
--- a/config/config.go
+++ b/config/config.go
@@ -24,6 +24,8 @@ type Config struct {
 	EtcdKeyFile             string
 	EtcdCertFile            string
 	EtcdCAFile              string
+	EtcdUsername            string
+	EtcdPassword            string
 	EtcdRequestTimeout      float64
 	EngineReconcileInterval float64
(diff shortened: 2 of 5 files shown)


thepwagner:
unsure about error, ./test passing for me on debian stretch
go version go1.4.2 linux/amd64

jonboulle:
I think it was spurious, I'll kick travis again

mischief:
that test tries to compare maps without sorting the contents, so it will pretty much never work.

jonboulle:
@mischief I thought it was a different issue. if that's really all it is
can we get that fixed?

mischief:
@jonboulle i forgot the exact details but i think that test also contains a race condition. but, yes, comparing maps should be removed from the tests.

jonboulle:
@mischief The test in question is using reflect.DeepEqual which handles map
ordering. There must be a difference in the contents

mischief:
indeed you are correct. sorry for the noise.

aheckmann:
+1

xiang90:
There is actually no test for this. Has anyone tested this manually? 

The auth is still an experimental feature for etcd 2.2. (https://github.com/coreos/etcd/blob/master/Documentation/auth_api.md)

Also enabling auth will significantly increase the latency of requests (https://github.com/coreos/etcd/issues/3223). I am a little bit worry about this since fleet does send some load to etcd.

thepwagner:
I tested manually before push: enabled auth@etcd in a coreos-vagrant sandbox and verified fleet could still schedule a unit. I didn't look at performance impact or even leave it running for long.

w.r.t. auth being an experimental feature, lack of support for it in fleet is preventing us from experimenting with it. Catch 22.

xiang90:
@thepwagner Thanks. I am fine with enabling it. Just express some of my concerns.

DeviaVir:
+1

wikiwi:
+1

dongsupark:
This PR looks nearly the same as #1506. Once #1506 got merged, I'm going to close this one.

dongsupark:
Closed in favor of https://github.com/coreos/fleet/pull/1630. Thanks.

Issue #1248, opened 2015-06-12
Transient test failures


bcwaldon:
```
$ ./test
Building fleetd...
Building fleetctl...
Running tests...
ERROR reconcile.go:253: Desired state and current state of Job(foo.service) nil, not sure what to do
ERROR reconcile.go:253: Desired state and current state of Job(foo.service) nil, not sure what to do
ERROR unit_state.go:224: Refusing to push UnitState(foo.service), no MachineID: &unit.UnitState{LoadState:"", ActiveState:"active", SubState:"", MachineID:"", UnitHash:"", UnitName:"foo.service"}
--- FAIL: TestUnitStatePublisherRunWithDelays-2 (0.00 seconds)
	unit_state_test.go:626: bad published UnitStates: got []string{"0.service", "2.service", "4.service", "1.service", "3.service"}, want []string{"0.service", "1.service", "2.service", "3.service", "4.service"}
	unit_state_test.go:640: bad published UnitStates: got []string{"0.service", "2.service", "4.service", "1.service", "3.service"}, want []string{"0.service", "1.service", "2.service", "3.service", "4.service"}
	unit_state_test.go:658: bad UnitStates: got []string{"foo.service", "baz.service", "bar.service"}, want []string{"foo.service", "bar.service", "baz.service"}
FAIL
coverage: 70.6% of statements
FAIL	github.com/coreos/fleet/agent	0.009s
ok  	github.com/coreos/fleet/api	0.010s	coverage: 80.4% of statements
ok  	github.com/coreos/fleet/config	0.004s	coverage: 100.0% of statements
ok  	github.com/coreos/fleet/engine	0.009s	coverage: 54.5% of statements
ok  	github.com/coreos/fleet/fleetctl	0.010s	coverage: 23.3% of statements
ok  	github.com/coreos/fleet/job	0.009s	coverage: 89.2% of statements
ok  	github.com/coreos/fleet/machine	0.007s	coverage: 35.1% of statements
ok  	github.com/coreos/fleet/pkg	0.009s	coverage: 81.9% of statements
ok  	github.com/coreos/fleet/pkg/lease	0.006s	coverage: 14.3% of statements
ok  	github.com/coreos/fleet/registry	0.008s	coverage: 26.7% of statements
ok  	github.com/coreos/fleet/ssh	0.009s	coverage: 42.5% of statements
ok  	github.com/coreos/fleet/systemd	0.006s	coverage: 15.5% of statements
[...]

PR #1354, opened 2015-09-16
Enable testing on Go 1.5

marineam:
Looks like this is all kinds of sad right now :(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,10 @@ language: go
 matrix:
   include:
     - go: 1.4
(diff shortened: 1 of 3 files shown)


marineam:
The test failure appears to be the same race condition reported in https://github.com/coreos/fleet/issues/1248 according to @mischief

marineam:
Woo, passes now by disabling the busted test. ;-)

mischief:
lgtm, but that test should really get fixed soon. 8-D

Issue #1377, opened 2015-10-13
*: version should include git sha


jonboulle:
`0.11.5+git` is not particularly helpful; `0.11.5+git436b082` would be much more so. We'll need to move from using a const to using linker flags.

jonboulle:
fixed by #1378

PR #1378, opened 2015-10-13
version: use git to determine version number

crawford:


jonboulle:
Eyeballing it looks fine but you made travis sad.

crawford:
Yeah, travis doesn't copy the git tree into its environment.

build
```diff
@@ -2,6 +2,8 @@
 
 ORG_PATH="github.com/coreos"
 REPO_PATH="${ORG_PATH}/fleet"
+VERSION=$(git describe --dirty)
+GLDFLAGS="-X github.com/coreos/fleet/version.Version \"${VERSION}\""
```
jonboulle: where did we end up on the space vs equals? https://github.com/coreos/ignition/blob/master/build#L7

crawford: I'll leave the space as long as we are testing with 1.4.

crawford:
My workaround is to set the default version to `0.0.0`. Thoughts?

jonboulle:
0.0.0+wasnotbuiltproperly :-)

crawford:
Done!

crawford:
Fixes https://github.com/coreos/fleet/issues/1377.

jonboulle:
Thanks buddy.

PR #1386, opened 2015-11-06
fleet binary accidentally invoked instead of fleetctl

jonboulle:
reported by @normelton https://gist.github.com/normelton/6b597287e791b5b3be26
At first glance it doesn't harm but aesthetically doesn't look good.

probably we have to add command interpreter

### DIFF
--- a/fleetd/fleetd.go
+++ b/fleetd/fleetd.go
@@ -37,6 +37,11 @@ const (
 	DefaultConfigFile = "/etc/fleet/fleet.conf"
 )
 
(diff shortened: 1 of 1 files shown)


jonboulle:
PTAL

kayrus:
LGTM

tixxdz:
Hi @jonboulle, having to check external arguments of another program may add an extra burden on us, so I just created this PR https://github.com/coreos/fleet/pull/1413

It seems that bug was already fixed by another commit, so it is already backward incompatible.

Thanks

jonboulle:
Closing in favour of #1413

PR #1393, opened 2015-11-23
Support for targeting `systemd --user`

hmalphettes:
Hi! I am looking at using fleet to manage a handful of machines with ssh access but without root access.
It might be a far shot or a misguided idea: please let me know if this is a dead-end.

If this is interesting I added 2 options to fleetd's configuration file to support that:
```
# fleetd.conf
units_dir=/home/vagrant/fleetunits
systemd_user=true
```

On archlinux, I can deploy `example/hello.service` but I get
`AgentReconciler task failed: type=LoadUnit job=hello.service reason="unit scheduled here but not loaded" err=No such file or directory`.
Some debugging shows that it comes from dbus's call to `LinkUnitFiles`; I have not figured out which folder it tries to link into yet.

Thanks for your attention.

### DIFF
--- a/fleetd/fleetd.go
+++ b/fleetd/fleetd.go
@@ -86,6 +86,7 @@ func main() {
 	cfgset.String("public_ip", "", "IP address that fleet machine should publish")
 	cfgset.String("metadata", "", "List of key-value metadata to assign to the fleet machine")
 	cfgset.String("agent_ttl", agent.DefaultTTL, "TTL in seconds of fleet machine state in etcd")
+	cfgset.String("units_directory", "/run/fleet/units/", "Path to the fleet units directory")
 	cfgset.Int("token_limit", 100, "Maximum number of entries per page returned from API requests")
 	cfgset.Bool("disable_engine", false, "Disable the engine entirely, use with care")
 	cfgset.Bool("disable_watches", false, "Disable the use of etcd watches. Increases scheduling latency")
@@ -225,6 +226,7 @@ func getConfig(flagset *flag.FlagSet, userCfgFile string) (*config.Config, error
 		DisableEngine:           (*flagset.Lookup("disable_engine")).Value.(flag.Getter).Get().(bool),
 		DisableWatches:          (*flagset.Lookup("disable_watches")).Value.(flag.Getter).Get().(bool),
 		VerifyUnits:             (*flagset.Lookup("verify_units")).Value.(flag.Getter).Get().(bool),
+		UnitsDirectory:          (*flagset.Lookup("units_directory")).Value.(flag.Getter).Get().(string),
 		TokenLimit:              (*flagset.Lookup("token_limit")).Value.(flag.Getter).Get().(int),
 		AuthorizedKeysFile:      (*flagset.Lookup("authorized_keys_file")).Value.(flag.Getter).Get().(string),
 	}
--- a/fleetd/fleetd.go
+++ b/fleetd/fleetd.go
@@ -87,6 +87,7 @@ func main() {
 	cfgset.String("metadata", "", "List of key-value metadata to assign to the fleet machine")
 	cfgset.String("agent_ttl", agent.DefaultTTL, "TTL in seconds of fleet machine state in etcd")
 	cfgset.String("units_directory", "/run/fleet/units/", "Path to the fleet units directory")
+	cfgset.Bool("systemd_user", false, "When true use systemd --user)")
 	cfgset.Int("token_limit", 100, "Maximum number of entries per page returned from API requests")
 	cfgset.Bool("disable_engine", false, "Disable the engine entirely, use with care")
 	cfgset.Bool("disable_watches", false, "Disable the use of etcd watches. Increases scheduling latency")
@@ -227,6 +228,7 @@ func getConfig(flagset *flag.FlagSet, userCfgFile string) (*config.Config, error
 		DisableWatches:          (*flagset.Lookup("disable_watches")).Value.(flag.Getter).Get().(bool),
 		VerifyUnits:             (*flagset.Lookup("verify_units")).Value.(flag.Getter).Get().(bool),
 		UnitsDirectory:          (*flagset.Lookup("units_directory")).Value.(flag.Getter).Get().(string),
+		SystemdUser:             (*flagset.Lookup("systemd_user")).Value.(flag.Getter).Get().(bool),
 		TokenLimit:              (*flagset.Lookup("token_limit")).Value.(flag.Getter).Get().(int),
 		AuthorizedKeysFile:      (*flagset.Lookup("authorized_keys_file")).Value.(flag.Getter).Get().(string),
 	}
--- a/functional/systemd_test.go
+++ b/functional/systemd_test.go
@@ -35,7 +35,7 @@ func TestSystemdUnitFlow(t *testing.T) {
 	}
 	defer os.RemoveAll(uDir)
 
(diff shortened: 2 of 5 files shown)


hmalphettes:
Hi and thanks for your attention.
I found out `systemd --user` was disabled in centos/redhat-7.
It turns out that this feature might be refactored heavily by the systemd developers before it goes any further:
https://bugzilla.redhat.com/show_bug.cgi?id=1198655#c3

So maybe it is too early to work on this for fleet?

If you guys have further insights regarding the future of `systemd --user` I am interested.

antrik:
@hmalphettes at a guess, your `fleetd.conf` might be the culprit: although I haven't tried following the code in question, I have serious doubts this can possibly work with a relative directory...

Aside from that, I have no idea whether the approach is fundamentally feasible.

fleetd/fleetd.go
```diff
@@ -81,6 +81,7 @@ func main() {
 	cfgset.String("metadata", "", "List of key-value metadata to assign to the fleet machine")
 	cfgset.String("agent_ttl", agent.DefaultTTL, "TTL in seconds of fleet machine state in etcd")
 	cfgset.String("units_directory", "/run/fleet/units/", "Path to the fleet units directory")
+	cfgset.Bool("systemd_system", true, "When false use systemd --user)")
```
antrik: Why not the other way around? I hope it's not just me -- but this looks kinda backwards...

hmalphettes: Agreed! Thanks for the review!

jonboulle:
> It turns out that this feature might be refactored heavily by the systemd developers before it goes any further:

I actually like how simple/clean this patch is, but yeah, might be worth being a little more certain about the future of --user before committing to it. Could you ask systemd-devel upstream?

fleetd/fleetd.go
antrik: Maybe you could just squash this into the previous commit, if it's not too much trouble?... If we all agree it's better this way around, there is no need to have the extra churn in permanent history :-)

hmalphettes: Sure! Done.

hmalphettes:
@jonboulle I did ask the systemd-devel upstream on bugzilla and here is the response from Lukáš Nykrýn:
```
We are not even sure that --user will stay in upstream as is in the future.
```
https://bugzilla.redhat.com/show_bug.cgi?id=1198655#c4

I'll try some other avenues to get more info then.

@antrik I did not take the time to test with an absolute path for the `units_dir`... yet.

jonboulle:
@hmalphettes I was actually referring to [systemd-devel](https://lists.freedesktop.org/mailman/listinfo/systemd-devel) (or the [GitHub project](https://github.com/systemd/systemd)) - the bugzilla is just for RH, not really upstream.

hmalphettes:
@jonboulle here we go: https://lists.freedesktop.org/archives/systemd-devel/2016-March/036091.html
thanks for the nudge!

jonboulle:
Let's punt this until we see what they say.

lemenkov:
@jonboulle @hmalphettes folks, you should really take a look at the Fedora fleet package. It does runs fleet under a "fleet" user. The secret sause is polkit rule:

http://pkgs.fedoraproject.org/cgit/rpms/fleet.git/tree/fleet.polkit.rules

lemenkov:
Sorry for the noise - didn't get the subject of this thread properly :)

hmalphettes:
So far 2 answers on the mailing list of downstream adopters who are interested:
https://lists.freedesktop.org/archives/systemd-devel/2016-April/036169.html
https://lists.freedesktop.org/archives/systemd-devel/2016-April/036170.html

I also went on IRC to get this idea reviewed. None of the systemd-devs pitched in. Apologies, I can't find the logs.
2 persons answered. They basically stated that `systemd --user` was feature complete and they were surprised I asked about it.

I'll keep waving my arms and asking questions when the opportunity shows up.
Cheers!

hmalphettes:
@jonboulle I can't make it to Berlin for CoreOS Fest. But some of you will be there and Systemd's own Lennart will be too.
https://coreos.com/blog/coreos-fest-schedule.html

will it be possible to relay then question then?

kayrus:
[IRC log](https://gist.githubusercontent.com/kayrus/a9f10bbe7cb621c220b9d436a8aab23d/raw/3c775192a21b8cffe40ec7bd70539c670db9c79d/systemd_user.log)

dongsupark:
+1 for this PR.
BTW ``functional/systemd_test.go`` needs to be also updated, as it doesn't build.

tixxdz:
How about logind and maybe startups without user login `loginctl enable-linger username` ?

hectorj2f:
I strongly believe that using `systemd --user` could offer a better response time when `starting` units with fleet. The reason is that fleet always runs a `systemd link` operation for each unit, and this operation is quite expensive, performance and scalability wise. So, if `systemd` knows where the units are located, there is not need for such a link operation.

dongsupark:
@hmalphettes gentle ping.
I think the only thing we need to do is updating a functional test, as I [mentioned](https://github.com/coreos/fleet/pull/1393#issuecomment-215406626). After it's done, I'd be happily merging this PR.

hmalphettes:
@dongsupark thanks for the reminder and cheers to that!
I did not realise we were this close to merge it.

functional/systemd_test.go
```diff
@@ -35,7 +35,7 @@ func TestSystemdUnitFlow(t *testing.T) {
 	}
 	defer os.RemoveAll(uDir)
 
-	mgr, err := systemd.NewSystemdUnitManager(uDir)
+	mgr, err := systemd.NewSystemdUnitManager(uDir, false)
```
hmalphettes: @dongsupark Fixed `functional/systemd_test.go` here.
Thanks again for the reminder!

dongsupark:
@hmalphettes Thanks. LGTM.
Though it's strange, semaphoreci isn't triggered automatically. Anyway I just ran the functional tests manually, and it's working.
One minor nit is that commit titles are not prefixed by subsystem names. But I'd say, that's no big deal, as this PR by itself is so simple as anyone can understand.
As a long-running PR is pain for everyone, I'd just merge this one next week, unless anyone has other opinion.
Thanks.

hmalphettes:
@dongsupark OK, I'll rewrite the commit messages.

hmalphettes:
@dongsupark Sorry I had missed https://github.com/coreos/fleet/blob/master/CONTRIBUTING.md#format-of-the-commit-message

hmalphettes:
@dongsupark One thing that worries me when reading this PR: the initial description suggests a value for the `units_dir` that is wrong: a relative path.

May I edit the initial description and replace `units_dir=./fleetunits` by `units_dir=/home/vagrant/fleetunits` ?

dongsupark:
@hmalphettes LGTM.
And sure, you can edit the initial description to replace a relative path with an absolute path.

hmalphettes:
@dongsupark ok. Edited. Cheers!

systemd/manager.go
```diff
@@ -64,6 +60,13 @@ func NewSystemdUnitManager(uDir string) (*systemdUnitManager, error) {
 	return &mgr, nil
 }
 
+func createDbusConnection(systemdUser bool) (*dbus.Conn, error) {
```
jonboulle: The indirection of this function seems unnecessary - I'd just inline the conditional into NewSystemdUnitManager

PR #1399, opened 2015-12-18
fleetctl: add -output to journal

### LINKED ISSUE #1398: Fleetctl journal needs a way to specify output format
`journalctl` has an `--output` flag that can be used to specify the output format. It would be helpful to include that in `fleetctl journal` as well, so we can choose to receive more/less verbose, or even machine readable output.

royvandewater:
this adds the --output flag to journal, which is passed through
to journalctl, so that you can specify its output format.

Fixes #1398

### DIFF
--- a/fleetctl/journal.go
+++ b/fleetctl/journal.go
@@ -24,10 +24,11 @@ var (
 	flagLines  int
 	flagFollow bool
 	flagSudo   bool
+	flagOutput string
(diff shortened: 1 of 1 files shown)


tixxdz:
Updated PR https://github.com/coreos/fleet/pull/1416

Thanks!

jonboulle:
Closing in favour of #1416

PR #1410, opened 2016-01-26
[PATCHv2] fleet:build: remove build warnings on go1.5

tixxdz:
When building fleet with go1.5 we got the following warning:
Building fleetd...
link: warning: option -X github.com/coreos/fleet/version.Version v0.10.1-117-g7b2a547 may not work in future releases; use -X github.com/coreos/fleet/version.Version=v0.10.1-117-g7b2a547
Building fleetctl...
link: warning: option -X github.com/coreos/fleet/version.Version v0.10.1-117-g7b2a547 may not work in future releases; use -X github.com/coreos/fleet/version.Version=v0.10.1-117-g7b2a547

Newer go versions will only support '-X name=value' format, so make our
build system smarter. It will make sure to use the right format for
future builds.

Reference: https://github.com/golang/go/issues/13498

### DIFF
--- a/build
+++ b/build
@@ -3,7 +3,10 @@
 ORG_PATH="github.com/coreos"
 REPO_PATH="${ORG_PATH}/fleet"
 VERSION=$(git describe --dirty)
-GLDFLAGS="-X github.com/coreos/fleet/version.Version \"${VERSION}\""
+CODE_VER_SYMBOL="github.com/coreos/fleet/version.Version"
+GO_VERSION=$(go version)
+
+GLDFLAGS=""
 
 if [ ! -h gopath/src/${REPO_PATH} ]; then
 	mkdir -p gopath/src/${ORG_PATH}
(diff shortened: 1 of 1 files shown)


jonboulle:
I might be missing something. Can't we just keep this super simple and do what ignition did? https://github.com/coreos/ignition/commit/b07829764f413d2265308abaaf88b01b0b2cee0b

tixxdz:
@jonboulle with go1.4.2 if I convert name=value format the code does not compile and we have the usage message from the linker... that was introduced in go1.5 ref: https://github.com/golang/go/issues/13498

kayrus:
@jonboulle let's use this solution in all our projects: https://github.com/coreos/flannel/blob/master/build

jonboulle:
I don't understand why ignition builds fine for me (and travis) on 1.4...

kayrus:
@jonboulle you are correct. that is because of `-installsuffix netgo`. Do we need to use this option?

tixxdz:
Hi @jonboulle @kayrus  I'm not sure about "-installsuffix netgo" and go1.4 but as said that format "-X name value" will be removed from go1.6. The warning is there to tell us to update, so I suggest to just patch build with https://github.com/coreos/flannel/blob/master/build

jonboulle:
OK fine, let's go with that for now, we can clean up when we drop 1.4

Djalal Harouni <notifications@github.com> schrieb am Mi., 27. Jan. 2016
11:17:

tixxdz:
@jonboulle ok this was merged in https://github.com/coreos/fleet/pull/1412

Please feel free to add a note or an issue to remember that this code can be dropped in case we require go1.5 or greater.

Thanks!

PR #1415, opened 2016-01-28
prometheus metrics for fleet

miekg:
This PR add prometheus metrics to fleet.

### DIFF
--- a/Documentation/metrics.md
+++ b/Documentation/metrics.md
@@ -1,0 +1,29 @@
+# Metrics
+
+**NOTE: The metrics feature is considered experimental. We may add/change/remove metrics without warning in future releases.**
+
+fleet uses [Prometheus](http://prometheus.io/) for metrics reporting in the
+server. The metrics can be used for real-time monitoring and debugging.
+
+See the [etcd
+metrics.md](https://github.com/coreos/etcd/blob/master/Documentation/metrics.md)
+for more information about metrics used in etcd and Prometheus.
+
+The naming of metrics follows the suggested best practice of Prometheus.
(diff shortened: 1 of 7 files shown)


miekg:
Note: didn't add prometheus (and friends) in gopath.

antrik:
@miekg before I try to review this in detail: on a quick glance, it looks like all the later commits just improve upon the original one, rather than making some distinct changes? In that case, they should probably all be squashed into a single commit.

You mentioned that you haven't added the prometheus dependency, but you didn't explain why?

One detail I noticed is that the copyright years are wrong. It should clearly be 2016 for the newly added file; and as far as I can see, the convention around here is to also bump the year of existing files when making substantial changes.

(Strictly speaking, when making changes, the original date should be kept along with the updated one; and also, since I guess you haven't signed over copyright to CoreOS Inc., they should mention your or your employer's name... But as far as I can see, these rules haven't been followed for fleet in the past either -- so not sure how to handle this...)

BTW, would it be feasible to add any unit tests and/or functional tests for this new feature?...

jonboulle:
@miekg I know your situation changed recently - let us know if you have
time to shepherd this through, otherwise one of us can take it over.

miekg:
[ Quoting <notifications@github.com> in "Re: [fleet] prometheus metrics for ..." ]
>@miekg I know your situation changed recently - let us know if you have
>time to shepherd this through, otherwise one of us can take it over.

Thanks. I don't think I can spend cycles on this one...

Thanks for stepping in.

/Miek

--
Miek Gieben

joshix:
Superseded closed by #1524

PR #1416, opened 2016-02-01
fleetctl: add -output to journal

### LINKED ISSUE #1398: Fleetctl journal needs a way to specify output format

tixxdz:
This adds the --output flag to journal, which is passed through
to journalctl, so that you can specify its output format.

Fixes #1398

Improved-by: Djalal Harouni <djalal@endocode.com>

### DIFF
--- a/fleetctl/journal.go
+++ b/fleetctl/journal.go
@@ -24,10 +24,11 @@ var (
 	flagLines  int
 	flagFollow bool
 	flagSudo   bool
+	flagOutput string
 	cmdJournal = &Command{
 		Name:    "journal",
 		Summary: "Print the journal of a unit in the cluster to stdout",
-		Usage:   "[--lines=N] [--ssh-port=N] [-f|--follow] <unit>",
+		Usage:   "[--lines=N] [--ssh-port=N] [-f|--follow] [--output=STRING] <unit>",
 		Run:     runJournal,
 		Description: `Outputs the journal of a unit by connecting to the machine that the unit occupies.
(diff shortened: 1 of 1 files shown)


jonboulle:
@tixxdz my concern is that then we're still then chasing journal arguments. What about just saying that it's passed through unaltered and hence supports whatever journalctl does?

tixxdz:
@jonboulle yes makes sense, ok updated. Thanks!

fleetctl/journal.go
```diff
@@ -47,6 +48,7 @@ func init() {
 	cmdJournal.Flags.BoolVar(&flagFollow, "f", false, "Shorthand for --follow")
 	cmdJournal.Flags.IntVar(&sharedFlags.SSHPort, "ssh-port", 22, "Connect to remote hosts over SSH using this TCP port")
 	cmdJournal.Flags.BoolVar(&flagSudo, "sudo", false, "Execute journal command with sudo")
+	cmdJournal.Flags.StringVar(&flagOutput, "output", "short", "Journalctl output mode")
```
jonboulle: can we make this 
> "journalctl output mode. This will be passed unaltered to journalctl on the remote host, and hence supports the same modes as that command."

jonboulle:
LGTM after one suggested change

Issue #1383, opened 2015-10-21
fleetctl (0.11.5) destroy is not backward compatible


btalbot:
Older versions of fleetctl seems to ignore the request to destroy units which do not actually exist. This was a handy command to use for startup / provisioning / startup scrips to stop and destroy a unit that may or may not actually be loaded.

```sh
core@ip-172-11-22-333 ~ $ fleetctl --version
fleetctl version 0.10.2
core@ip-172-11-22-333 ~ $ fleetctl destroy foo
core@ip-172-11-22-333 ~ $ echo $?
0
```

However, fleetctl 0.11.5 causes destroy show an unhelpful error message and exit with code 1 if the unit is not loaded

```sh
core@core-01 ~ $ fleetctl --version
fleetctl version 0.11.5
core@core-01 ~ $ fleetctl destroy fubar
Error destroying units: googleapi: Error 404: unit does not exist
core@core-01 ~ $ echo $?
1
```

This has now hit the beta channel and broke some of our development and test deployment scripts.

Is the new behavior a regression, bug, or a feature?

mischief:
this was introduced by https://github.com/coreos/fleet/commit/9530eedb82ac8a3af7412c62afc5dd30d17d6892

btalbot:
Note that other fleetctl commands for units that do not exist do not generate the error message or exit code:

```sh
core@core-01 ~ $ fleetctl --version
fleetctl version 0.11.5
core@core-01 ~ $ fleetctl stop fubar; echo $?
0
core@core-01 ~ $ fleetctl unload fubar; echo $?
0
```

megahall:
Would it be possible to backport the fix in #1417 into `Fleet v0.11.x` or ship a CoreOS which includes `Fleet 0.12.x`? Because no current version of CoreOS includes `Fleet 0.12.x` as of yet.

dongsupark:
@megahall :+1:
However this issue is already closed. How about creating a new issue on https://github.com/coreos/bugs/issues, so that other maintainers could easily track it?

megahall:
Downstream report filed as https://github.com/coreos/bugs/issues/1448 .

PR #1417, opened 2016-02-01
fleetctl: on destroy command ignore units that do not exist

tixxdz:
fleetctl used to ignore requests to destory units which do not actually exist.
This behaviour was changed by commit 9530eed, restore the older behaviour by calling findUnits()
to get the right list of units.

This fixes: https://github.com/coreos/fleet/issues/1383

### DIFF
--- a/fleetctl/destroy.go
+++ b/fleetctl/destroy.go
@@ -33,9 +33,14 @@ Destroyed units are impossible to start unless re-submitted.`,
 }
 
 func runDestroyUnits(args []string) (exit int) {
-	for _, v := range args {
-		name := unitNameMangle(v)
-		err := cAPI.DestroyUnit(name)
+	units, err := findUnits(args)
+	if err != nil {
+		stderr("%v", err)
+		return 1
+	}
+
+	for _, v := range units {
+		err := cAPI.DestroyUnit(v.Name)
 		if err != nil {
 			stderr("Error destroying units: %v", err)
 			exit = 1
@@ -56,7 +61,7 @@ func runDestroyUnits(args []string) (exit int) {
 			}
 
 			for retry() {
-				u, err := cAPI.Unit(name)
+				u, err := cAPI.Unit(v.Name)
 				if err != nil {
 					stderr("Error destroying units: %v", err)
 					exit = 1
@@ -70,7 +75,7 @@ func runDestroyUnits(args []string) (exit int) {
 			}
 		}
 
-		stdout("Destroyed %s", name)
+		stdout("Destroyed %s", v.Name)
 	}
 	return
 }


kayrus:
@jonboulle Have checked. It works, but I'd argue with this. If unit doesn't exist - fleetctl should return non-zero code. And we have to document all possible exit statuses.

jonboulle:
capturing @mirkoboehm's suggestion - we could add a `--sensitive` option (i.e. `--error-on-not-exist`) to destroy/stop/unload - then we can restore the previous default behaviour but provide an option for this behaviour

fleetctl/destroy.go
```diff
@@ -33,9 +33,14 @@ Destroyed units are impossible to start unless re-submitted.`,
...
-		name := unitNameMangle(v)
-		err := cAPI.DestroyUnit(name)
+	units, err := findUnits(args)
+	if err != nil {
+		stderr("%v", err)
+		return 1
+	}
+
+	for _, v := range units {
+		err := cAPI.DestroyUnit(v.Name)
```
jonboulle: instead of doing two separate calls, can we check the type of the error returned, and ignore it if it's not-exist?

tixxdz: I thought that findUnits() is needed since it retrieves all units then it filters them, making sure to mangle names etc. It will only fail if it can't retrieve any unit all which is another problem. Or perhaps I'm missing something ?

antrik:
@jonboulle please clarify: do you think we need actually to add the --error-on-not-exist option, or is this regression fix fine to be merged alone? (Once the feedback comment on the patch has been addressed...)

jonboulle:
I think this regression fix should be merged, and then if we want to
support that use case, we can do a follow up feature request.

fleetctl/destroy.go
jonboulle: @tixxdz I'm not really happy this got merged without a follow up to this comment (or an LGTM). It's actually a change in behaviour (and does not restore to the semantic of pre- https://github.com/coreos/fleet/commit/9530eedb82ac8a3af7412c62afc5dd30d17d6892 as the commit message implies).

Previously, if we received a command `fleetctl destroy a b c`, and `a` and `b` existed but `c` did not, we would nevertheless successfully remove `a` and `b`. Now, we will immediately error out instead, because `c` couldn't be found.

My suggestion was to just check the response from each `cAPI.DestroyUnit()` call, and ignore the error if it's one indicating the unit doesn't exist (this might require some error plumbing). As it is now, we have a race condition between getting the list of units from `findUnits` and doing the `Destroy` call.

antrik: Seems there was a misunderstanding regarding the discussion in the main thread. Will follow up with a proper fix.

tixxdz: @jonboulle oh for the race condition you have a point here, yes. Adding checks for Destroy will fix it of course, it could be done without removing findUnits() call since my aim was to have more consistency and avoid hard coding logic for each command. @antrik will follow up with a fix.

megahall:
Would it be possible to backport this fix in #1417 into `Fleet v0.11.x` or ship a CoreOS which includes `Fleet 0.12.x`? Because no current version of CoreOS includes `Fleet 0.12.x` as of yet.

dongsupark:
@megahall :+1:
However this PR is already closed. How about creating a new issue on https://github.com/coreos/bugs/issues, so that other maintainers could easily track it?

megahall:
Downstream report filed as https://github.com/coreos/bugs/issues/1448 .

dongsupark:
@megahall FYI I just released [v0.12.0](https://github.com/coreos/fleet/releases/tag/v0.12.0) that includes this fix #1417.

Issue #478, opened 2014-05-19
filter ListUnits dbus method to subscribed units


jonboulle:
Per philips, we should do something like the following to narrow down what messages we actually listen to
http://cgit.freedesktop.org/systemd/systemd/commit/?id=cdc06ed7b6120c1049305fa7033f228ee9d86043

jonboulle:
This might be low hanging fruit to potentially improve performance with large numbers of units. Needs benchmarking.

kayrus:
@jonboulle this filter applies on units' status only. and there is no method which will request units info only for specified unit names or filter. here is experimental implementation of this feature:
https://github.com/endocode/systemd/commit/ebe011c3bc64cf67096b65b66472894f99f8e399

kayrus:
@jonboulle @antrik @tixxdz quick benchmark of the https://github.com/endocode/systemd/commit/ebe011c3bc64cf67096b65b66472894f99f8e399:

patched systemd and systemctl version:

```sh
time for i in {1..1000}; do ./systemctl list-units fleet.service > /dev/null 2>&1; done

real    0m4.087s
user    0m0.901s
sys     0m1.023s
```

original version:

```sh
time for i in {1..1000}; do systemctl list-units fleet.service > /dev/null 2>&1; done

real    0m7.305s
user    0m1.914s
sys     0m1.279s
```

tixxdz:
@jonboulle this one is not really a low hanging fruit, if fixed correctly from systemd to other layers it will also make fleet much better, since it does not only covers some if not all case of PR #1418 but it will also reflect on other parts of the code. However, the task is not really easy and it needs several fixes.

Thanks!

PR #1418, opened 2016-02-02
Add unit state cache to mitigate the iteration between fleet and systemd - dbus

hectorj2f:
Unit state cache: 

**Motivation:** Inactive units induce an overload to systemd-dbus communication which linearly increases proportionally to the amount of existing inactive units. Therefore, we decided to research about solutions to minimize the workload of fleet caused by this issue. We decided to build a cache to reduce the amount of calls from systemd to dbus. Fleet periodically performs operations to gather the state of the units in the cluster, so for each inactive unit located in a host, a read operation is triggered to obtain the state from dbus. Therefore, you could have an impression of how this operation could use more resources than the desired.

Nevertheless, we thought about another optimal solution during the development of this cache. This solution requires to modify the `go-dbus` library implementation to provide an event-driven mechanism. This event-driven mechanism would notify every time an operation affects the state of a unit. These notifications would enable to know when the state of a unit change without having to constantly query systemd to get its state.  Obviously we declined this solution as it required the modification of third party software used by fleet.

## Implementation
We added a new flag `enable_unitstate_cache` that will activate the usage of the unit state cache and therefore to minimize the usage of the system resources. The unit state cache is used to give the state of inactive units whenever no operation has been recently triggered over them. Otherwise the system will query to dbus in order to get its final state. Consequently, we register operations that are being triggered over the units in order to decide whether or not to use its cached state.

### Test scenario:
- 3 node cluster
- fleet version 0.11.3
- CoreOS stable (835.9.0)
- etcd version 2.2.1

## Metrics analyzed:
- CPU usage
- Memory usage
- FS bytes usage
- NET bytes IN
- NET bytes OUT
- NET bytes TOTAL 
- Capacity used

### Savings deducted from these experiments:

A brief and quick summary:
- There is a clear CPU reduction using the unit state cache. In the following experiments, this benefit can be identified as approx. 10% CPU usage reduction. However this amount will increase/decrease based on the amount of inactive units.
- There is a clear Memory usage reduction using the unit state cache.  In the following experiments, this benefit is up to approx. 100Mb Memory usage reduction. However this amount will increase/decrease based on the amount of inactive units.
- FS bytes used during the experiments slightly INCREASED with the unit state implementation.
- NET bytes IN/OUT used during the experiments slightly INCREASED in the implementation WITHOUT unit cache.
- Capacity used obviously decreased when using the unit state cache.

### Experiment 1:
Without any iteration in the cluster, we evaluated how 3 inactive units could use the system resources of a host.

#### CPU usage of our implementation using an UNIT STATE CACHE:
screen shot 2016-01-21 at 6 06 52 pm

#### Memory usage of our implementation using an UNIT STATE CACHE:
screen shot 2016-01-21 at 6 07 11 pm

#### CPU usage WITHOUT unit state cache:
screen shot 2016-01-21 at 6 04 31 pm

#### Memory usage WITHOUT unit state cache:
screen shot 2016-01-21 at 6 04 59 pm

As shown above, there is a slight difference caused by these periodic queries to dbus in order to get the state of inactive units.

### Experiment 2:
We used a benchmarking tool that starts 900 units in the cluster, and consequently stop+destroy all after a timeout.

#### CPU usage of our implementation using an UNIT STATE CACHE:
screen shot 2016-01-21 at 5 39 26 pm

#### CPU usage WITHOUT unit state cache:
screen shot 2016-01-21 at 6 15 52 pm
--------------------
### Memory usage of our implementation using an UNIT STATE CACHE:
screen shot 2016-01-21 at 3 59 26 pm

### Memory usage WITHOUT unit state cache:
screen shot 2016-01-21 at 6 15 38 pm
--------------------
### FS bytes used of our implementation using an UNIT STATE CACHE:
screen shot 2016-01-21 at 3 59 37 pm
### FS bytes used WITHOUT unit state cache:
screen shot 2016-01-21 at 3 57 13 pm
--------------------
### Capacity usage of our implementation using an UNIT STATE CACHE:
screen shot 2016-01-21 at 3 59 44 pm
### Capacity used WITHOUT unit state cache:
screen shot 2016-01-21 at 3 57 21 pm
--------------------
### Net bytes IN of our implementation using an UNIT STATE CACHE:
screen shot 2016-01-21 at 3 59 54 pm

### Net bytes IN WITHOUT unit state cache:
screen shot 2016-01-21 at 3 56 42 pm
----------------------
### Net bytes OUT of our implementation using an UNIT STATE CACHE:
screen shot 2016-01-21 at 4 00 01 pm

### Net bytes OUT WITHOUT unit state cache:
screen shot 2016-01-21 at 3 56 50 pm

--------------------
### Net bytes total of our implementation using an UNIT STATE CACHE:
screen shot 2016-01-21 at 4 00 10 pm
### Net bytes total WITHOUT unit state cache:
screen shot 2016-01-21 at 6 38 58 pm

NOTE: As shown in above Figures (CPU usage, Mem usage, ...) there are two spikes that represent the start operation and the stop/destroy operation which is performed after a fixed timeout.

### DIFF
--- a/fleetd/fleetd.go
+++ b/fleetd/fleetd.go
@@ -91,6 +91,7 @@ func main() {
 	cfgset.Bool("disable_watches", false, "Disable the use of etcd watches. Increases scheduling latency")
 	cfgset.Bool("verify_units", false, "DEPRECATED - This option is ignored")
 	cfgset.String("authorized_keys_file", "", "DEPRECATED - This option is ignored")
+	cfgset.Bool("enable_unitstate_cache", true, "Enable an unit state cache to minimize the systemd and dbus communication overhead.")
 
 	globalconf.Register("", cfgset)
 	cfg, err := getConfig(cfgset, *cfgPath)
@@ -227,8 +228,12 @@ func getConfig(flagset *flag.FlagSet, userCfgFile string) (*config.Config, error
 		VerifyUnits:             (*flagset.Lookup("verify_units")).Value.(flag.Getter).Get().(bool),
 		TokenLimit:              (*flagset.Lookup("token_limit")).Value.(flag.Getter).Get().(int),
 		AuthorizedKeysFile:      (*flagset.Lookup("authorized_keys_file")).Value.(flag.Getter).Get().(string),
+		EnableUnitStateCache:    (*flagset.Lookup("enable_unitstate_cache")).Value.(flag.Getter).Get().(bool),
 	}
 
+	if cfg.EnableUnitStateCache {
+		log.Info("Config option enable_unitstate_cache is activated")
+	}
 	if cfg.VerifyUnits {
 		log.Error("Config option verify_units is no longer supported - ignoring")
 	}
--- a/fleetd/fleetd.go
+++ b/fleetd/fleetd.go
@@ -91,7 +91,7 @@ func main() {
 	cfgset.Bool("disable_watches", false, "Disable the use of etcd watches. Increases scheduling latency")
 	cfgset.Bool("verify_units", false, "DEPRECATED - This option is ignored")
 	cfgset.String("authorized_keys_file", "", "DEPRECATED - This option is ignored")
-	cfgset.Bool("enable_unitstate_cache", true, "Enable an unit state cache to minimize the systemd and dbus communication overhead.")
+	cfgset.Bool("enable_unitstate_cache", false, "Enable an unit state cache to minimize the systemd and dbus communication overhead.")
 
 	globalconf.Register("", cfgset)
 	cfg, err := getConfig(cfgset, *cfgPath)
--- a/systemd/manager.go
+++ b/systemd/manager.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/coreos/fleet/Godeps/_workspace/src/github.com/coreos/go-systemd/dbus"
 
+	"github.com/coreos/fleet/config"
 	"github.com/coreos/fleet/log"
 	"github.com/coreos/fleet/pkg"
 	"github.com/coreos/fleet/unit"
@@ -38,9 +39,24 @@ type systemdUnitManager struct {
 
 	hashes map[string]unit.Hash
 	mutex  sync.RWMutex
+
+	enableUnitStateCache bool
+	unitStateCache       *unitStateCache
 }
 
-func NewSystemdUnitManager(uDir string) (*systemdUnitManager, error) {
+type unitStateCache struct {
+	unitStates            map[string]*unit.UnitState
+	registerUnitOperation map[string]string
+}
+
+func NewSystemdUnitManager(cfg config.Config, uDir string) (*systemdUnitManager, error) {
+	statesCache := &unitStateCache{}
+	if cfg.EnableUnitStateCache {
+		statesCache = &unitStateCache{
+			unitStates:            make(map[string]*unit.UnitState),
+			registerUnitOperation: make(map[string]string),
+		}
+	}
 	systemd, err := dbus.New()
 	if err != nil {
 		return nil, err
@@ -56,10 +72,12 @@ func NewSystemdUnitManager(uDir string) (*systemdUnitManager, error) {
 	}
 
 	mgr := systemdUnitManager{
-		systemd:  systemd,
-		unitsDir: uDir,
-		hashes:   hashes,
-		mutex:    sync.RWMutex{},
+		systemd:              systemd,
+		unitsDir:             uDir,
+		hashes:               hashes,
+		mutex:                sync.RWMutex{},
+		enableUnitStateCache: cfg.EnableUnitStateCache,
+		unitStateCache:       statesCache,
 	}
 	return &mgr, nil
 }
@@ -107,6 +125,10 @@ func (m *systemdUnitManager) Load(name string, u unit.UnitFile) error {
 		return err
 	}
 	m.hashes[name] = u.Hash()
+	if m.enableUnitStateCache {
+		m.unitStateCache.registerUnitOperation[name] = "load"
+	}
+
 	return nil
 }
 
@@ -117,6 +139,10 @@ func (m *systemdUnitManager) Unload(name string) {
 	defer m.mutex.Unlock()
 	delete(m.hashes, name)
 	m.removeUnit(name)
+
+	if m.enableUnitStateCache {
+		m.unitStateCache.registerUnitOperation[name] = "unload"
+	}
 }
 
 // TriggerStart asynchronously starts the unit identified by the given name.
@@ -127,6 +153,10 @@ func (m *systemdUnitManager) TriggerStart(name string) {
 		log.Infof("Triggered systemd unit %s start: job=%d", name, jobID)
 	} else {
 		log.Errorf("Failed to trigger systemd unit %s start: %v", name, err)
+	}
+
+	if m.enableUnitStateCache {
+		m.unitStateCache.registerUnitOperation[name] = "start"
 	}
 }
 
@@ -139,6 +169,9 @@ func (m *systemdUnitManager) TriggerStop(name string) {
 	} else {
 		log.Errorf("Failed to trigger systemd unit %s stop: %v", name, err)
 	}
+	if m.enableUnitStateCache {
+		m.unitStateCache.registerUnitOperation[name] = "stop"
+	}
 }
 
 // GetUnitState generates a UnitState object representing the
@@ -146,6 +179,13 @@ func (m *systemdUnitManager) TriggerStop(name string) {
 func (m *systemdUnitManager) GetUnitState(name string) (*unit.UnitState, error) {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
+
+	// If there is not any operation previously registered, we return the cached state
+	_, ok := m.unitStateCache.registerUnitOperation[name]
+	if m.enableUnitStateCache && !ok && m.unitStateCache.unitStates[name] != nil {
+		return m.unitStateCache.unitStates[name], nil
+	}
+
 	us, err := m.getUnitState(name)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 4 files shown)


hectorj2f:
I performed a different experiment, this time we had 6 inactive units per host with a total of 18 inactive units in the whole cluster. We used our benchmarking tool called fleemmer (as done above) and deployed 900 units, which were stopped/destroy after a fixed timeout.

### CPU usage WITHOUT unit state cache:
screen shot 2016-01-25 at 11 23 26 am


### CPU usage using our implementation of an UNIT STATE CACHE:

screen shot 2016-01-25 at 12 13 04 pm

**CONCLUSION:** There is a slight increment in the CPU usage due to the inactive units approx. 7-8% CPU usage increment. However, the CPU usage always increases in a higher percentage when scheduling units how shown in the spikes.

tixxdz:
@hectorj2f thank you very much for the detailed information, we will read it again carefully and try to push it forward.

For the moment I'm just a bit curious, you say "We decided to build a cache to reduce the amount of calls from systemd to dbus. Fleet periodically performs operations to gather the state of the units in the cluster, so for each inactive unit located in a host, a read operation is triggered to obtain the state from dbus. Therefore, you could have an impression of how this operation could use more resources than the desired." and later you also say that it can be solved by updating go-dbus, for me that's the best solution. Your patches seem clean and they don't touch lot of code. However I would first ask if we can support notification from systemd ?  subscribe() then get UnitNew() and UnitRemoved() signals  https://www.freedesktop.org/wiki/Software/systemd/dbus/   systemd supports notification and it should not trigger any load at all... is this currently not possible ?

Thank you!

hectorj2f:
@tixxdz  Subscriptions are already used in the function `GetUnitStates()` of `systemd/manager.go`. In the code, fleet defines subscribers for any instance to get its status update, e.g. when running async operations. However those subscribers are just `for loops` that use systemd-dbus to get the current state (look at SubscribeUnitsCustom function in go-dbus). This is done to be aware of any state change. 

The dbus Subscribe() function you mentioned, it reacts against any systemd dbus event. I believe it is not what we would want. We want to define subscribers for specific fleet units, (similar to what subscriptor set offers) but in a smart way. The subscribers should be asynchronous to report whenever a state changes instead of done via for loops querying all the states and filtering the units.

Indeed, my long-term proposal was to refactor go-dbus to report state updates only when the state changes and not by constantly querying its state.

jonboulle:
For a little history: this was previously purely event-driven - similar to how we previously used etcd - but as with etcd, we could end up missing events and have no way to sanely recover, so we ended up moving this to the reconciliation model as well.

hectorj2f:
@jonboulle Thanks for the information :). Then it makes more sense to me cause it made it better regarding data consistency and failure recovery. Unfortunately we also found some situations on which fleet had a different state than the one in systemd, but that is another issue. 

@jonboulle Do you think the current solution could have similar problems ? or do you refer to the implementation of dbus watchers ?

This PR aims to reduce the workload caused by this periodic unit state calls for all the subscribed units. @tixxdz the state of the subscribed units is collected via dbus `ListUnits` calls. Moreover for all those units that don't appear in the `ListUnits` call, an explicit `getProperties` dbus call is triggered for each unit.

hectorj2f:
Do you see any blocker or potential issue coming out when using this caching approach @jonboulle ?

tixxdz:
Hi @hectorj2f 
In the mean time could you please force push your branch: giantswarm:unit_state_cache  so the pull request is updated which will trigger the functional tests.

Thank you!

hectorj2f:
@tixxdz Done. However there is an error that seems to be unrelated :/ .
https://travis-ci.org/coreos/fleet/jobs/117506419

hectorj2f:
ping @tixxdz

tixxdz:
@hectorj2f yes thank you for the update! the failing is due to travis-ci. Hopefully we will give it more time next week as it is in the 0.13 priority.

Thanks!

hectorj2f:
@tixxdz test are passing, what's next ?

antrik:
@hectorj2f to be honest, I think the main hurdle is that we are still pretty new to this, and it's hard for any of us to judge whether such a change is indeed a good idea... (While @jonboulle doesn't really have the time to think about this too much I guess.)

I haven't really looked into this myself; one thing that comes to mind though is that AIUI the cache gets out of sync if a unit state changes because of "external" factors, not at a direct request of fleet?... How does that does that deal with crashing units for example, or unit dependencies fleet is not aware of?

BTW, this is a bit tangential, but there is one thing in your original message I do not agree with: "Obviously we declined this solution as it required the modification of third party software used by fleet." To quote a certain Lennart Poettering: "You own the entire stack." This is free software -- you never have to take limitations in underlying infrastructure as granted, and work around them. Fix problems at the source.

tixxdz:
@hectorj2f hi sorry for delay, we had to handle some other issues, at the same time I'm reviewing this issue https://github.com/coreos/fleet/issues/478 since it relates the one exposed here, just trying to get the whole context from your comments and @jonboulle ones ;-)

hectorj2f:
@tixxdz Thanks

> How does that deal with crashing units for example, or unit dependencies fleet is not aware of?

@antrik Well, that has a very simple answer. This cache won't interfere as the state of the systemd units remains with the same logic than before. Please have a look on line 242 in `systemd/manager.go`. This cache aims to reduce the workload of the inactive units.

fleetd/fleetd.go
```diff
@@ -91,6 +91,7 @@ func main() {
 	cfgset.Bool("disable_watches", false, "Disable the use of etcd watches. Increases scheduling latency")
 	cfgset.Bool("verify_units", false, "DEPRECATED - This option is ignored")
 	cfgset.String("authorized_keys_file", "", "DEPRECATED - This option is ignored")
+	cfgset.Bool("enable_unitstate_cache", true, "Enable an unit state cache to minimize the systemd and dbus communication overhead.")
```
tixxdz: Could this be made false by default ?

tixxdz: Yes at least when it's merged we let it false by default others can run it if they want and after some time it can be switch to true. Thanks

systemd/manager.go
```diff
@@ -218,27 +258,49 @@ func (m *systemdUnitManager) GetUnitStates(filter pkg.Set) (map[string]*unit.Uni
...
-		if h, ok := m.hashes[name]; ok {
-			us.UnitHash = h.String()
+		// If there is any operation previously registered we collect the states from systemd
+		_, ok := m.unitStateCache.registerUnitOperation[name]
+		if m.enableUnitStateCache && !ok && m.unitStateCache.unitStates[name] != nil {
+			unitStates[name] = m.unitStateCache.unitStates[name]
+		} else {
+			us, err := m.getUnitState(name)
+			if err != nil {
+				return unitStates, err
```
tixxdz: Previously GetUnitStates() was returning a nil and an error here, now errors are ignored and states are returned.

fleetd/fleetd.go
hectorj2f: Done @tixxdz

systemd/manager.go
hectorj2f: Done. good catch!

hectorj2f:
@tixxdz PRs seem to fail. Please, ping me when it is fixed to push it again. Thx

tixxdz:
@hectorj2f just an update, we are making the functional tests run when state cache is set and not, will merge the functional test change first, then later this PR. I may add some minor comments.

Thank you!

tixxdz:
So more notes:

1) Did you investigate if you put the transition of states inside say
     `systemdUnitManager.cache_hashes {
            hashes map[string]unit.Hash ;  state string => {active or inactive }
      `
  
So we have a new cache_hashes with two fields where we unload the unit we don't remove it but we put the state into the inactive state, the other code should check if the state is always active and the addition should check if the state is inactive. This will allow us to use the same cache but in the other way I'm not sure if it will really scale ? cause with the current implementation we just forget about the old cache https://github.com/coreos/fleet/pull/1418/files#diff-561bbb797eb2f941c1384271efe28c0bR273  which will be collected and the `registerUnitOperation` cache is also updated where it's appropriate, this of course saves memory but at the end we may have three hashes running at the same time, may be we can achieve the same with two or one hash which will be survive the calls and one used inside the functions as a temporary hash to update the final one ?

2) Would it be possible to have some benchmark comparison with more inactive units a higher number ? Just to make sure before merging.

3) It would be nice if we put the details of the logic inside functions, say instead of:
`if m.enableUnitStateCache { delete(m.unitStateCache.registerUnitOperation, name) }`

`func (m *systemdUnitManager) cacheUpdateActiveUnits(blabla....) { if !m.enableUnitStateCache { return } .... continue ....}`

It will allows us to hide the logic and maybe improve the cache later ?


So 1) needs further thinking, do you think it makes sense ? I need more time to came up with something that can work, otherwise just use what's already here and improve it later.

And for 2) and 3) it would be awesome to update the PR based on these! and just matter of time before we merge this ;-)

Thanks for your patience.

tixxdz:
Also, squashing the commits where it's possible would be nice. btw we merged more functional tests for unit states these past days, I guess that's enough or do you think that there are corner cases that can be added to assert the functionality ?

hectorj2f:
@tixxdz I like the idea but I need to think about the corner cases. 

Regarding 2) and 3), they are feasible and understandable :).

tixxdz:
@hectorj2f for 1) if it's too much, then please don't bother. Without it, the feature will be an option disabled by default, later if it can be done then it could be enabled by default for every one. At the same time if you use or depend on this PR https://github.com/coreos/fleet/pull/1556 would be nice to test it, confirm.

Thank you!

kayrus:
@hectorj2f I'm not quite sure, but I suppose this improvement should be done in low-level side. I've started to implement systemd-dbus optimizations withing this PR: https://github.com/coreos/fleet/pull/1564

hectorj2f:
@kayrus My concern are mainly the `inactive` - `dead` units. I'll test your changes at the end of this week, and let you to know if there are any improvements. I am currently quite busy with other stuff :-/.

kayrus:
@hectorj2f I've reworked my patch a bin and created systemd pull request. If you wish to test systemd v225 which is currectly used by stable CoreOS, please use this patch: https://github.com/endocode/systemd/commit/1fdb12b6d5a6c6286fde5140b7e04a5a6db02abe

hectorj2f:
After talking with @kayrus, I think this PR might get obsolete once https://github.com/coreos/fleet/pull/1564 is merged. I'd like to test it once that is done.

jonboulle:
Let's consider this on hold in favour of #1564

Issue #1404, opened 2016-01-19
docs: clarify relationship between start/stop/load/...


jonboulle:
The existing explanations for the behaviour of different fleetctl verbs (start, load, etc) are very brief and the relationship between these is a frequent source of confusion for users. 

We should add a doc that walks through this in detail - in particular that fleet  models things declaratively, and that the commands are not imperative actions.

PR #1420, opened 2016-02-02
Added unit file lifecycle in fleet

kayrus:
/cc @jonboulle @joshix @robszumski 

resolves https://github.com/coreos/fleet/issues/1404

### DIFF
--- a/Documentation/states.md
+++ b/Documentation/states.md
@@ -1,6 +1,6 @@
 ## Unit states
 
-There are currently three cluster-level states for a unit:
+fleet uses a _declarative model_ to evaluate unit state. This means that operations to change the state of units (e.g. `fleetctl` commands, or calls to the fleet API) change the desired state, rather than directly performing any state change. There are currently three cluster-level states for a unit:
 
 - `inactive`: known by fleet, but not assigned to a machine
 - `loaded`: assigned to a machine and loaded into systemd there, but not started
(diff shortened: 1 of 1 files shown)


robszumski:
Two comments:

1. This doesn't seem like it requires a new file. Is there an existing doc where it makes sense to place this?
2. https://github.com/coreos/fleet/issues/1404 specifically calls out:

  > We should add a doc that walks through this in detail - in particular that fleet models things declaratively, and that the commands are not imperative actions.

  Which is not covered in this. This is a start, but I wouldn't say that it fixes https://github.com/coreos/fleet/issues/1404.

jonboulle:
This actually already exists in this document: https://github.com/coreos/fleet/blob/master/Documentation/states.md#unit-states

I think it just needs to be made clearer, and/or maybe moved elsewhere, to make it more obvious.

kayrus:
moved scheme to `states.md` and added info about declarative model.

Documentation/states.md
```diff
@@ -12,15 +12,14 @@ The desired and last known states are exposed in the `DSTATE` and `STATE` column
 
 The `fleetctl` commands to act on units change the *desired state* of a unit. fleet itself is then responsible for performing the necessary state transitions to move a unit to the desired state. The following table explains the relationship between each `fleetctl` command and unit states.
 
-| Command | Desired State | Valid Previous States |
-|---------|--------------|-----|
-| `fleetctl submit`  | `inactive`  | `(unknown)`
```
robszumski: I know it's probably not _technically_ correct, but I'd lean towards saying "none" vs "unknown", but I don't feel strongly about it.

Documentation/states.md
```diff
@@ -12,15 +12,14 @@ The desired and last known states are exposed in the `DSTATE` and `STATE` column
...
+| Command | Description | Desired State | Valid Previous States | Is an alias for |
+|---------|-------------|--------------|-----|----|
+| `fleetctl submit`  | Submits unit file into etcd storage | `inactive`  | `(unknown)` | |
+| `fleetctl load`    | Submits and schedules unit file into machines' systemd but doesn't start it | `loaded` | `(unknown)` or `inactive` | `submit+load` |
+| `fleetctl start`   | Submits, schedules and starts unit file| `launched`  | `(unknown)` or `inactive` or `loaded` | `submit+load+start` |
+| `fleetctl stop`    | Stops scheduled unit file | `loaded`  | `launched` | |
+| `fleetctl unload`  | Stops and unschedules unit file from machines' systemd | `inactive`| `launched` or `loaded` | `stop+unload` |
+| `fleetctl destroy` | Stops, unschedules and removes unit file from etcd storage| `(unknown)` | `launched` or `loaded` or `inactive` | `stop+unload+destroy` |
 
 `(unknown)` indicates that the unit has not yet been submitted to fleet at all (or it previously existed in fleet but was destroyed).
```
robszumski: If we used "none", this note probably isn't necessary.

robszumski:
Two comments, but I don't fell too strongly about them.

LGTM

Documentation/states.md
kayrus: Yep. I like `none`

kayrus:
@robszumski fixed

robszumski:
LGTM. I was thinking `none` instead of `(none)` but it's not super critical.

Documentation/states.md
```diff
@@ -1,6 +1,6 @@
 ## Unit states
 
-There are currently three cluster-level states for a unit:
+`fleetctl` uses declarative model which means it sets desired state to units. For example when you wish to start a unit `fleetctl` sets desired state inside etcd storage and `fleet` daemons connected to this storage just perform desired action.  There are currently three cluster-level states for a unit:
```
jonboulle: I don't really like leading with fleetctl here; this is a property of the fleet architecture/model, rather than anything fleetctl specific. For example if someone interacts using just the API, this is all still the case

Also, `fleet` should be `fleetd`

jonboulle: How about:

> fleet uses a _declarative model_ to evaluate unit state. This means that operations to change the state of units (e.g. `fleetctl` commands, or calls to the fleet API) change the desired state, rather than directly performing any state change.

kayrus:
@jonboulle fixed and squashed

jonboulle:
thank you!

PR #1430, opened 2016-02-11
[WIP] test: added test scripts for the functional tests

kayrus:
subj

### DIFF
--- a/build
+++ b/build
@@ -1,41 +1,24 @@
 #!/bin/bash -e
 
-# The -X format changed from go1.4 -> go1.5
-function go_linker_dashX {
-	local version=$(go version)
-	local regex="go([0-9]+).([0-9]+)."
-	if [[ $version =~ $regex ]]; then
-		if [ ${BASH_REMATCH[1]} -eq "1" -a ${BASH_REMATCH[2]} -le "4" ]; then
-			echo "$1 \"$2\""
-		else
-			echo "$1=$2"
-		fi
-	else
-		echo "could not determine Go version"
-		exit 1
-	fi
-}
+CDIR=$(cd `dirname $0` && pwd)
+cd $CDIR
 
 ORG_PATH="github.com/coreos"
 REPO_PATH="${ORG_PATH}/fleet"
 VERSION=$(git describe --dirty)
-GLDFLAGS="-X $(go_linker_dashX github.com/coreos/fleet/version.Version ${VERSION})"
+
+source build-env
 
 if [ ! -h gopath/src/${REPO_PATH} ]; then
-	mkdir -p gopath/src/${ORG_PATH}
-	ln -s ../../../.. gopath/src/${REPO_PATH} || exit 255
+  mkdir -p gopath/src/${ORG_PATH}
+  ln -s ../../../.. gopath/src/${REPO_PATH} || exit 255
 fi
 
-export GOBIN=${PWD}/bin
-export GOPATH=${PWD}/gopath
-
-eval $(go env)
-
 if [ ${GOOS} = "linux" ]; then
-	echo "Building fleetd..."
-	CGO_ENABLED=0 go build -o bin/fleetd -a -installsuffix netgo -ldflags "${GLDFLAGS}" ${REPO_PATH}/fleetd
+  echo "Building fleetd..."
+  CGO_ENABLED=0 go build -o bin/fleetd -a -installsuffix netgo -ldflags "${GLDFLAGS}" ${REPO_PATH}/fleetd
 else
-	echo "Not on Linux - skipping fleetd build"
+  echo "Not on Linux - skipping fleetd build"
 fi
 
 echo "Building fleetctl..."
--- a/functional/README.md
+++ b/functional/README.md
@@ -2,52 +2,101 @@
 
 This functional test suite deploys a fleet cluster using nspawn containers, and asserts fleet is functioning properly.
 
-It shares an instance of etcd deployed on the host machine with each of the nspawn containers.
+It shares an instance of etcd deployed on the host machine with each of the nspawn containers which use `172.18.0.1/16` network, so please make sure this network does not intersect with others.
 
-It's recommended to run this in a virtual machine environment on CoreOS (e.g. using coreos-vagrant). The only dependency for the tests not provided on the CoreOS image is `go`.
+It's recommended to run this in a virtual machine environment on CoreOS (e.g. using [Vagrant][test-in-vagrant]).
 
-The caller must do three things before running the tests:
+Since the tests utilize [`systemd-nspawn`][systemd-nspawn], this needs to be invoked as sudo/root.
 
-1. Ensure an ssh-agent is running and the functional-testing identity is loaded. The `SSH_AUTH_SOCK` environment variable must be set.
+If the tests are aborted partway through, it's currently possible for them to leave residual state as a result of the `systemd-nspawn` operations. This can be cleaned up using the `clean.sh` script.
 
-```
-$ ssh-agent
-$ ssh-add fleet/functional/fixtures/id_rsa
-$ echo $SSH_AUTH_SOCK
-/tmp/ssh-kwmtTOsL7978/agent.7978
-```
-2. Ensure the `FLEETD_BIN` and `FLEETCTL_BIN` environment variables point to the respective fleetd and fleetctl binaries that should be used to drive the actual tests.
+### Run tests in Vagrant
 
-```
-$ export FLEETD_BIN=/path/to/fleetd
-$ export FLEETCTL_BIN=/path/to/fleetctl
+The recommended way to run the tests is to use the provided Vagrantfile, which will set up a single CoreOS instance with a one-member etcd cluster (configuration is applied using `user-data` [Cloud-Config][cloud-config] file located in this directory).
+To do so, simply run the following commands on a system with Vagrant installed (see [Vagrant configuration][configure-vagrant] section of this doc)
+
+```sh
+$ git clone https://github.com/coreos/fleet
+$ cd fleet/functional
+$ ./run-in-vagrant
 ```
 
-3. Make sure etcd is running on the host system.
+Vagrant's provision step includes go binaries download using `functional/provision/install_go.sh` script.
 
-```
-$ systemctl start etcd
-```
+### Run tests inside other CoreOS platforms (QEMU/BareMetal/libvirt/etc)
 
-Then the tests can be run with:
+It's also possible to run the tests on CoreOS on other platforms. The following commands should be run *inside* the CoreOS instance.
 
-```
-# go test github.com/coreos/fleet/functional
+```sh
+$ git clone https://github.com/coreos/fleet
 ```
 
-Since the tests utilize `systemd-nspawn`, this needs to be invoked as sudo/root.
+If you didn't configure etcd2 daemon yet, just run this script:
 
-An example test session using coreos-vagrant follows. This assumes that go is available in `/home/core/go` and the fleet repository in `/home/core/fleet` on the target machine (the easiest way to achieve this is to use shared folders).
-```
-vagrant ssh core-01 -- -A
-export GOROOT="$(pwd)/go"
-export PATH="${GOROOT}/bin:$PATH"
-cd fleet
-ssh-add functional/fixtures/id_rsa
-export GOPATH="$(pwd)/gopath"
-export FLEETD_BIN="$(pwd)/bin/fleetd"
-export FLEETCTL_BIN="$(pwd)/bin/fleetctl"
-sudo -E env PATH=$PATH go test github.com/coreos/fleet/functional -v
+```sh
+$ sudo fleet/functional/start_etcd
 ```
 
-If the tests are aborted partway through, it's currently possible for them to leave residual state as a result of the systemd-nspawn operations. This can be cleaned up using the `clean.sh` script.
+It will configure and start a one-member etcd cluster.
+
+Then run the functional tests (script will download and unpack golang into home directory):
+
+```sh
+$ sudo fleet/functional/test
+```
+
+When `fleet/functional/test` can not find go binaries, it will download them automatically using `functional/provision/install_go.sh` script.
+
+## Configure host environment to run Vagrant
+
+### Debian/Ubuntu
+
+#### Install Vagrant
+
+```sh
+sudo apt-get install -y git nfs-kernel-server
+wget https://releases.hashicorp.com/vagrant/1.8.1/vagrant_1.8.1_x86_64.deb
+sudo dpkg -i vagrant_1.8.1_x86_64.deb
+```
+
+#### Install VirtualBox
+
+```sh
+echo "deb http://download.virtualbox.org/virtualbox/debian $(lsb_release -sc) contrib" | sudo tee /etc/apt/sources.list.d/virtualbox.list
+wget -q https://www.virtualbox.org/download/oracle_vbox.asc -O- | sudo apt-key add -
+sudo apt-get update
+sudo apt-get install -y build-essential dkms
+sudo apt-get install -y VirtualBox-5.0
+#Previous VirtualBox (if you have problems with nested virtualization, more info here: https://www.virtualbox.org/ticket/14965)
+#sudo apt-get install -y VirtualBox-4.3
+```
+
+### CentOS/Fedora
+
+**NOTE**: NFS and Vagrant doesn't work out of the box on CentOS 6.x, so it is recommended to use CentOS 7.x
+
+#### Install Vagrant
+
+```sh
+sudo yum install -y git nfs-utils
+sudo service nfs start
+sudo yum install -y https://releases.hashicorp.com/vagrant/1.8.1/vagrant_1.8.1_x86_64.rpm
+```
+
+#### Install VirtualBox
+
+```sh
+source /etc/os-release
+for id in $ID_LIKE $ID; do break; done
+OS_ID=${id:-rhel}
+curl http://download.virtualbox.org/virtualbox/rpm/$OS_ID/virtualbox.repo | sudo tee /etc/yum.repos.d/virtualbox.repo
+sudo yum install -y make automake gcc gcc-c++ kernel-devel-`uname -r` dkms
+sudo yum install -y VirtualBox-5.0
+#Previous VirtualBox (if you have problems with nested virtualization, more info here: https://www.virtualbox.org/ticket/14965)
+#sudo yum install -y VirtualBox-4.3
+```
+
+[test-in-vagrant]: #run-tests-in-vagrant
+[configure-vagrant]: #configure-host-environment-to-run-vagrant
+[systemd-nspawn]: https://www.freedesktop.org/software/systemd/man/systemd-nspawn.html
+[cloud-config]: https://github.com/coreos/coreos-cloudinit/blob/master/Documentation/cloud-config.md
--- a/functional/Vagrantfile
+++ b/functional/Vagrantfile
@@ -1,0 +1,146 @@
+# -*- mode: ruby -*-
+# # vi: set ft=ruby :
+# Vagrantfile based on official CoreOS Vagranfile https://github.com/coreos/coreos-vagrant with one extra provision string
+
+require 'fileutils'
+
+Vagrant.require_version ">= 1.6.0"
+
+CLOUD_CONFIG_PATH = File.join(File.dirname(__FILE__), "user-data")
+CONFIG = File.join(File.dirname(__FILE__), "config.rb")
+
+# Defaults for config options defined in CONFIG
+$num_instances = 1
+$instance_name_prefix = "core"
+$update_channel = "alpha"
+$image_version = "current"
+$enable_serial_logging = false
+$share_home = false
+$vm_gui = false
+$vm_memory = 1024
+$vm_cpus = 1
+$shared_folders = {}
+$forwarded_ports = {}
+
+# Attempt to apply the deprecated environment variable NUM_INSTANCES to
+# $num_instances while allowing config.rb to override it
+if ENV["NUM_INSTANCES"].to_i > 0 && ENV["NUM_INSTANCES"]
+  $num_instances = ENV["NUM_INSTANCES"].to_i
+end
+
+if File.exist?(CONFIG)
+  require CONFIG
+end
+
+# Use old vb_xxx config variables when set
+def vm_gui
+  $vb_gui.nil? ? $vm_gui : $vb_gui
+end
+
+def vm_memory
+  $vb_memory.nil? ? $vm_memory : $vb_memory
+end
+
+def vm_cpus
+  $vb_cpus.nil? ? $vm_cpus : $vb_cpus
+end
+
+Vagrant.configure("2") do |config|
+  # always use Vagrants insecure key
+  config.ssh.insert_key = false
+
+  config.vm.box = "coreos-%s" % $update_channel
+  if $image_version != "current"
+      config.vm.box_version = $image_version
+  end
+  config.vm.box_url = "https://storage.googleapis.com/%s.release.core-os.net/amd64-usr/%s/coreos_production_vagrant.json" % [$update_channel, $image_version]
+
+  ["vmware_fusion", "vmware_workstation"].each do |vmware|
+    config.vm.provider vmware do |v, override|
+      override.vm.box_url = "https://storage.googleapis.com/%s.release.core-os.net/amd64-usr/%s/coreos_production_vagrant_vmware_fusion.json" % [$update_channel, $image_version]
+    end
+  end
+
+  config.vm.provider :virtualbox do |v|
+    # On VirtualBox, we don't have guest additions or a functional vboxsf
+    # in CoreOS, so tell Vagrant that so it can be smarter.
+    v.check_guest_additions = false
+    v.functional_vboxsf     = false
+  end
+
+  # plugin conflict
+  if Vagrant.has_plugin?("vagrant-vbguest") then
+    config.vbguest.auto_update = false
+  end
+
+  (1..$num_instances).each do |i|
+    config.vm.define vm_name = "%s-%02d" % [$instance_name_prefix, i] do |config|
+      config.vm.hostname = vm_name
+
+      if $enable_serial_logging
+        logdir = File.join(File.dirname(__FILE__), "log")
+        FileUtils.mkdir_p(logdir)
+
+        serialFile = File.join(logdir, "%s-serial.txt" % vm_name)
+        FileUtils.touch(serialFile)
+
+        ["vmware_fusion", "vmware_workstation"].each do |vmware|
+          config.vm.provider vmware do |v, override|
+            v.vmx["serial0.present"] = "TRUE"
+            v.vmx["serial0.fileType"] = "file"
+            v.vmx["serial0.fileName"] = serialFile
+            v.vmx["serial0.tryNoRxLoss"] = "FALSE"
+          end
+        end
+
+        config.vm.provider :virtualbox do |vb, override|
+          vb.customize ["modifyvm", :id, "--uart1", "0x3F8", "4"]
+          vb.customize ["modifyvm", :id, "--uartmode1", serialFile]
+        end
+      end
+
+      if $expose_docker_tcp
+        config.vm.network "forwarded_port", guest: 2375, host: ($expose_docker_tcp + i - 1), auto_correct: true
+      end
+
+      $forwarded_ports.each do |guest, host|
+        config.vm.network "forwarded_port", guest: guest, host: host, auto_correct: true
+      end
+
+      ["vmware_fusion", "vmware_workstation"].each do |vmware|
+        config.vm.provider vmware do |v|
+          v.gui = vm_gui
+          v.vmx['memsize'] = vm_memory
+          v.vmx['numvcpus'] = vm_cpus
+        end
+      end
+
+      config.vm.provider :virtualbox do |vb|
+        vb.gui = vm_gui
+        vb.memory = vm_memory
+        vb.cpus = vm_cpus
+      end
+
+      ip = "172.17.8.#{i+100}"
+      config.vm.network :private_network, ip: ip
+
+      # Uncomment below to enable NFS for sharing the host machine into the coreos-vagrant VM.
+      #config.vm.synced_folder ".", "/home/core/share", id: "core", :nfs => true, :mount_options => ['nolock,vers=3,udp']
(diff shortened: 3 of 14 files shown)


jonboulle:
I don't understand the use case for this. can you help me understand what it provides over just a bash script that runs this? Since we still need etcd, systemd etc running on the host.

kayrus:
It is still WIP. And this script will be executed on coreos.

tixxdz:
You may also check if etcd is up.

jonboulle:
What is the point of doing this in docker at all? Look how much stuff you necessarily need to bring in from the host..

kayrus:
@jonboulle this will allow you to run tests directly on coreos. which already has etcd.

jonboulle:
Why do you need docker for that? The functional tests run fine on coreos
without it

kayrus <notifications@github.com> schrieb am Fr., 12. Feb. 2016 11:50:

kayrus:
@jonboulle there is not go binary inside coreos.

antrik:
To elaborate a bit: according to `functional/README.md`, the official way to run tests is inside a CoreOS VM, typically through Vagrant. However, as CoreOS doesn't provide a Go environment itself, the README suggest mapping it from outside the VM using Vagrant shared folders. But that's neither easy nor a self-contained reproducible environment. In our understanding, the canonical way to get a Go environment inside CoreOS is by running a golang container.

jonboulle:
@antrik thanks for the context, very helpful. I discussed out of band with @mirkoboehm and we'll circulate notes shortly

kayrus:
@jonboulle tests passed. should be squash and merge?

functional/platform/nspawn.go
```diff
@@ -302,14 +302,18 @@ func (nc *nspawnCluster) createMember(id string) (m Member, err error) {
 
 		// minimum requirements for running systemd/coreos in a container
 		fmt.Sprintf("mkdir -p %s/usr", fsdir),
+		//fmt.Sprintf("mkdir -p %s/lib", fsdir), //ubuntu
```
jonboulle: can you explain all these?

functional/platform/nspawn.go
```diff
@@ -395,6 +401,7 @@ UseDNS no
 			return
 		default:
 		}
+		log.Printf("Dialing machine: %s", addr)
```
jonboulle: do we need this?

functional/start_etcd
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash -e
```
jonboulle: what is this script about?

functional/platform/nspawn.go
kayrus: I tried to run these tests in ubuntu without cloud-config. Decided to keep this. If you don't like - i'll remove.

functional/platform/nspawn.go
kayrus: Yes, otherwise we don't have any feedback from tests. btw, we already have `fleetctl list-machines... ` in output. Why don't print `Dialing`?

functional/start_etcd
kayrus: For those who want to run tests in qemu (and any other coreos environments) without specified cloud-config.

build
```diff
@@ -1,29 +1,31 @@
 #!/bin/bash -e
 
+CDIR=$(cd `dirname $0` && pwd)
```
jonboulle: is this working around some problem you encountered?

functional/README.md
```diff
@@ -1,3 +1,63 @@
...
+sudo apt-get update
+sudo apt-get install -y build-essential dkms
+sudo apt-get install -y VirtualBox-5.0
+#Previous VirtualBox (if you have problems with nested virtualization, more info here: https://www.virtualbox.org/ticket/14965)
+#sudo apt-get install -y VirtualBox-4.3
+```
+
+### Centos/Fedora
+
+**NOTE**: NFS doesn't work on CentOS 6.x
```
jonboulle: what does this imply? that they must use >=7?

functional/README.md
```diff
@@ -1,3 +1,63 @@
...
+git clone https://github.com/coreos/fleet
+cd fleet/functional
+# wget https://raw.githubusercontent.com/coreos/coreos-vagrant/master/Vagrantfile
+# we will use default unmodified CoreOS Vagrantfile
+./run-in-vagrant
+```
+
 ## fleet functional tests
 
 This functional test suite deploys a fleet cluster using nspawn containers, and asserts fleet is functioning properly.
```
jonboulle: This intro should still be at the top of the file, otherwise the configure stuff lacks context

functional/Vagrantfile
```diff
@@ -0,0 +1,145 @@
+# -*- mode: ruby -*-
```
jonboulle: is this just a straight port of coreos-vagrant, plus  the provision script?

functional/provision/setup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
+
+USER_ID=${SUDO_UID:-$(id -u)}
+HOME=$(getent passwd "${USER_ID}" | cut -d: -f6)
+
+export GOROOT=${HOME}/go
+export PATH=${HOME}/go/bin:${PATH}
+
+if [ ! -f ${HOME}/go1.5.3.linux-amd64.tar.gz ]; then
```
jonboulle: split out go1.5.3.linux-amd64.tar.gz as a var

functional/provision/setup.sh
```diff
@@ -0,0 +1,14 @@
...
+  # Remove unfinished archive when you press Ctrl+C
```
jonboulle: this seems unnecessary, why not just download it in a mktemp and extract it into HOME?

functional/start_etcd
jonboulle: OK, this needs to be documented, it's not at all self explanatory

build
kayrus: Yes, when you run scripts from home directory for example. `cd ~ ; sudo ./fleet./functional/test`

functional/start_etcd
```diff
@@ -0,0 +1,31 @@
...
+
+USER_ID=${SUDO_UID:-$(id -u)}
+HOME=$(getent passwd "${USER_ID}" | cut -d: -f6)
```
jonboulle: I don't really understand all these machinations. Why do we need HOME? why do we need to ensure we're sudo and not the root user? can't you just put it in temp and run it from there?

functional/README.md
kayrus: Don't want to waste my time trying to debug why it doesn't work in old CentOS 6. So yes.

functional/start_etcd
jonboulle: e.g. maybe they should run coreos-cloudinit --from-file in that case?

functional/Vagrantfile
reneengelhard: yes

kayrus: The only one string we added is `config.vm.provision :shell, :path => "provision/setup.sh", :privileged => false`. The rest is original coreos-vagrant Vagrantfile.

functional/README.md
jonboulle: That's fine, I'm just trying to suggest explicitness

functional/provision/setup.sh
kayrus: Sounds reasonable.

functional/platform/nspawn.go
jonboulle: We definitely shouldn't leave commented lines in here. Are the lines totally superfluous now? The tests all work fine without?

functional/platform/nspawn.go
```diff
@@ -302,14 +302,18 @@ func (nc *nspawnCluster) createMember(id string) (m Member, err error) {
...
+		fmt.Sprintf("echo 'core:x:500:500:CoreOS Admin:/home/core:/bin/bash' > %s/etc/passwd", fsdir),
+		fmt.Sprintf("echo 'core:x:500:' > %s/etc/group", fsdir),
 		fmt.Sprintf("ln -s /proc/self/mounts %s/etc/mtab", fsdir),
 		fmt.Sprintf("ln -s usr/lib64 %s/lib64", fsdir),
 		fmt.Sprintf("ln -s lib64 %s/lib", fsdir),
 		fmt.Sprintf("ln -s usr/bin %s/bin", fsdir),
 		fmt.Sprintf("ln -s usr/sbin %s/sbin", fsdir),
 		fmt.Sprintf("mkdir -p %s/home/core/.ssh", fsdir),
-		fmt.Sprintf("chown -R core:core %s/home/core", fsdir),
+		fmt.Sprintf("chown -R 500:500 %s/home/core", fsdir),
```
jonboulle: is this necessary after you just added the user above?

functional/provision/setup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
```
jonboulle: install_go.sh

functional/test
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash -e
+
+CDIR=$(cd `dirname $0` && pwd)
+USER_ID=${SUDO_UID:-$(id -u)}
+HOME=$(getent passwd "${USER_ID}" | cut -d: -f6)
+
+cd ${CDIR}/../
```
jonboulle: If you're going to the trouble of changing to the root of the repo here, then just use relative paths for the rest of the script. Otherwise it's too much to maintain

functional/start_etcd
kayrus: Temp will be cleaned up after reboot. And after reboot IP address could be also changed. This script also works under root user.

functional/platform/nspawn.go
kayrus: Yes, they do. Will remove ubuntu stuff. But will keep `.. 500 ...` for future support.

functional/provision/setup.sh
reneengelhard: ok, one might (for whatever reason) want to do other stuff in setup.sh, though, so I'll make setup.sh call a new install_go.sh

functional/platform/nspawn.go
kayrus: Yes, otherwise user directory will be owned by root

functional/provision/setup.sh
kayrus: Will fix. And note above too.

functional/start_etcd
kayrus: after we run `setup-network-environment` - yes. will fix that.

functional/README.md
kayrus: True. Will fix.

functional/test
kayrus: Relatives paths will not help when you run scripts from the outside, i.e. `cd ~ ; sudo ./fleet/functional/test`

functional/provision/setup.sh
reneengelhard: nevermind my last comment; did a complete move now anyways as what might make sense in a setup.sh might need privileges which we donÄ't need for install_go.sh and if we ever needed it we can run extra setup.sh

functional/platform/nspawn.go
jonboulle: I mean why can't it be core:core

functional/test
kayrus: Fixed

functional/platform/nspawn.go
kayrus: Because we run this command on host, not inside container and it will not work for non-coreos tests in future.

kayrus:
@jonboulle done, please review once again.

functional/README.md
```diff
@@ -4,50 +4,93 @@ This functional test suite deploys a fleet cluster using nspawn containers, and
...
+$ sudo fleet/functional/start_etcd
 ```
-$ systemctl start etcd
+
+It will configure and start one-member etcd2 daemon.
+
+Then run the functional tests:
+
+```sh
+$ git clone https://github.com/coreos/fleet
```
jonboulle: haven't we already cloned?

functional/README.md
```diff
@@ -4,50 +4,93 @@ This functional test suite deploys a fleet cluster using nspawn containers, and
...
+$ sudo fleet/functional/test
 ```
 
-Then the tests can be run with:
+## configure environment to run Vagrant
```
jonboulle: Configure host environment to run Vagrant

functional/README.md
```diff
@@ -4,50 +4,93 @@ This functional test suite deploys a fleet cluster using nspawn containers, and
...
-```
-2. Ensure the `FLEETD_BIN` and `FLEETCTL_BIN` environment variables point to the respective fleetd and fleetctl binaries that should be used to drive the actual tests.
+### run tests in vagrant
 
+Vagrant will configure CoreOS instance with one-member etcd2 daemon.
+
+```sh
+$ git clone https://github.com/coreos/fleet
+$ cd fleet/functional
+$ # Vagrantfile based on official CoreOS Vagranfile with one extra provision string
```
jonboulle: can this comment instead go in the vagrantfile?

functional/README.md
```diff
@@ -4,50 +4,93 @@ This functional test suite deploys a fleet cluster using nspawn containers, and
 
 It shares an instance of etcd deployed on the host machine with each of the nspawn containers.
 
-It's recommended to run this in a virtual machine environment on CoreOS (e.g. using coreos-vagrant). The only dependency for the tests not provided on the CoreOS image is `go`.
+It's recommended to run this in a virtual machine environment on CoreOS (e.g. using [Vagrant][test-in-vagrant]).
 
-The caller must do three things before running the tests:
+Since the tests utilize `[systemd-nspawn]`, this needs to be invoked as sudo/root.
```
jonboulle: this needs -> they need

functional/README.md
```diff
@@ -4,50 +4,93 @@ This functional test suite deploys a fleet cluster using nspawn containers, and
...
+If the tests are aborted partway through, it's currently possible for them to leave residual state as a result of the `systemd-nspawn` operations. This can be cleaned up using the `clean.sh` script.
 
-```
-$ ssh-agent
-$ ssh-add fleet/functional/fixtures/id_rsa
-$ echo $SSH_AUTH_SOCK
-/tmp/ssh-kwmtTOsL7978/agent.7978
-```
-2. Ensure the `FLEETD_BIN` and `FLEETCTL_BIN` environment variables point to the respective fleetd and fleetctl binaries that should be used to drive the actual tests.
+### run tests in vagrant
```
jonboulle: Run tests in Vagrant

functional/README.md
```diff
@@ -4,50 +4,93 @@ This functional test suite deploys a fleet cluster using nspawn containers, and
...
 
+Vagrant will configure CoreOS instance with one-member etcd2 daemon.
```
jonboulle: The recommended way to run the tests is to use the provided Vagrantfile, which will set up a single CoreOS instance with a one-member etcd cluster.

functional/README.md
```diff
@@ -4,50 +4,93 @@ This functional test suite deploys a fleet cluster using nspawn containers, and
...
+```sh
+$ git clone https://github.com/coreos/fleet
+$ cd fleet/functional
+$ # Vagrantfile based on official CoreOS Vagranfile with one extra provision string
+$ ./run-in-vagrant
 ```
-$ export FLEETD_BIN=/path/to/fleetd
-$ export FLEETCTL_BIN=/path/to/fleetctl
+
+### run tests inside other CoreOS platforms (QEMU/BareMetal/libvirt/etc)
```
jonboulle: Run

functional/README.md
```diff
@@ -4,50 +4,93 @@ This functional test suite deploys a fleet cluster using nspawn containers, and
...
+
```
jonboulle: It's also possible to run the tests on CoreOS on other platforms

functional/README.md
```diff
@@ -4,50 +4,93 @@ This functional test suite deploys a fleet cluster using nspawn containers, and
...
 
-3. Make sure etcd is running on the host system.
+If you didn't configure etcd2 daemon yet, just run this script:
 
+```sh
+$ sudo fleet/functional/start_etcd
 ```
-$ systemctl start etcd
+
+It will configure and start one-member etcd2 daemon.
```
jonboulle: a one-member etcd cluster

functional/config.rb
```diff
@@ -0,0 +1,58 @@
+# Size of the CoreOS cluster created by Vagrant
```
jonboulle: Do we need this configuration? It's just one more moving part. If we have already modified the Vagrantfile, then just do it in there and remove this file.

functional/platform/nspawn.go
jonboulle: Oh, I missed we don't use systemd-nspawn for this, sorry.

functional/platform/nspawn.go
jonboulle: I don't follow?

functional/test-docker
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash -e
```
jonboulle: Do we need this any more?

functional/Vagrantfile
```diff
@@ -0,0 +1,145 @@
...
+      if $share_home
+        config.vm.synced_folder ENV['HOME'], ENV['HOME'], id: "home", :nfs => true, :mount_options => ['nolock,vers=3,udp']
+      end
+
+      if File.exist?(CLOUD_CONFIG_PATH)
+        config.vm.provision :file, :source => "#{CLOUD_CONFIG_PATH}", :destination => "/tmp/vagrantfile-user-data"
+        config.vm.provision :shell, :inline => "mv /tmp/vagrantfile-user-data /var/lib/coreos-vagrant/", :privileged => true
+      end
+
+      config.vm.provision :shell, :path => "provision/install_go.sh", :privileged => false
```
jonboulle: Can't we actually just do this in the cloudinit?

functional/config.rb
kayrus: It is much easier to sync modified Vagrantfile with the original one. So let's use config.rb

functional/platform/nspawn.go
kayrus: Sometimes when your environment has problems, this output message will help you to debug.

functional/test-docker
kayrus: If we already have `build-docker` we have to put `test-docker`.

functional/Vagrantfile
kayrus: Great idea. Our Vagrantfile will be identical to original one. Will do that.

functional/config.rb
kayrus: No, we can not. Because in this case we will have to create a hook which will wait until provisioning has been finished. And we don't control cloudinit execution in vagrant. This will complicate things.

functional/Vagrantfile
kayrus: No, we can not. Because in this case we will have to create a hook which will wait until provisioning has been finished. And we don't control cloudinit execution in vagrant. This will complicate things.

functional/test
```diff
@@ -0,0 +1,32 @@
...
+export VERSION=$(git describe --dirty)
+export GOROOT=${HOME}/go
+export PATH=${HOME}/go/bin:${PATH}
+
+if [ ! -S "$SSH_AUTH_SOCK" ]; then
+  eval $(ssh-agent)
+fi
+
+chmod 600 functional/fixtures/id_rsa
```
tixxdz: Please document why we need this one.

functional/test-env
```diff
@@ -0,0 +1,23 @@
...
+      echo "$1=$2"
+    fi
+  else
+    echo "could not determine Go version"
+    exit 1
+  fi
+}
+
+export GOBIN=${PWD}/bin
+export GOPATH=${PWD}/gopath
```
tixxdz: I would say if some of these environment are shared between the build and test, then just make a single copy in the root directory of fleet, then source it from there. Not sure if currently it's the case but it would be better if we start the functional tests from the root directory too then cd into functional ? at least one would be able to start them from root directory.

Thank you!

functional/user-data
```diff
@@ -0,0 +1,12 @@
...
+---
+coreos:
+  etcd2:
+    advertise-client-urls: http://$private_ipv4:2379
+    listen-client-urls: http://0.0.0.0:2379,http://0.0.0.0:4001
+  units:
+  - name: etcd2.service
+    command: start
+  update:
+    reboot-strategy: off
```
tixxdz: Adding documentation to this file would also help, why it's here...

Thanks!

functional/test-docker
```diff
@@ -0,0 +1,45 @@
...
+echo -e '#!/bin/sh\n LD_LIBRARY_PATH=/opt/lib64 /opt/lib64/ld-linux-x86-64.so.2 /opt/sbin/brctl \$@' > /opt/brctl
+echo -e '#!/bin/sh\n LD_LIBRARY_PATH=/opt/lib64 /opt/lib64/ld-linux-x86-64.so.2 /opt/bin/systemd-nspawn \$@' > /opt/systemd-nspawn
+echo -e '#!/bin/sh\n LD_LIBRARY_PATH=/opt/lib64 /opt/lib64/ld-linux-x86-64.so.2 /opt/bin/systemctl \$@' > /opt/systemctl
+echo -e '#!/bin/sh\n LD_LIBRARY_PATH=/opt/lib64 /opt/lib64/ld-linux-x86-64.so.2 /opt/bin/machinectl \$@' > /opt/machinectl
+chmod +x /opt/brctl /opt/systemd-nspawn /opt/systemctl /opt/machinectl
+source functional/test-env
+eval \$(go env)
+export PATH=\"\${PATH}:/opt:/opt/bin:/opt/sbin\"
+go test github.com/coreos/fleet/functional -ldflags \"\${GLDFLAGS}\" -v
+"
```
tixxdz: Ok it would be better if we separate this logic into its own, it's too long! and if something is shared then make a single copy of it into the root directory. thanks!

functional/platform/nspawn.go
tixxdz: I would suggest to separate these requirements on their own, not hard code them inside the code, just a separate function that does the setup for us in its own file ? @kayrus you could open a PR later for this one perhaps ?

functional/test-env
kayrus: Moved test-env into /build-env

functional/test
kayrus: Done

functional/user-data
kayrus: Done.

functional/platform/nspawn.go
```diff
@@ -211,10 +211,10 @@ func (nc *nspawnCluster) prepCluster() (err error) {
 		return
 	}
 
-	if !strings.Contains(stdout, "172.17.0.1/16") {
-		_, _, err = run("ip addr add 172.17.0.1/16 dev fleet0")
+	if !strings.Contains(stdout, "172.18.0.1/16") {
```
jonboulle: context for this change?

kayrus: If you have running docker daemon with the default configuration in CoreOS (`172.17.42.1`), these networks intersect and tests don't work.

jonboulle: OK, comment to that effect would help for posterity.

functional/Vagrantfile
jonboulle: Why do you need to wait for provisioning to finish? Can't you just write a one-shot that invokes the equivalent of install_go.sh ?

jonboulle: Then we can even get rid of that file too

functional/README.md
```diff
@@ -2,52 +2,98 @@
...
-export FLEETCTL_BIN="$(pwd)/bin/fleetctl"
-sudo -E env PATH=$PATH go test github.com/coreos/fleet/functional -v
+
+#### Install VirtualBox
+
+```sh
+source /etc/os-release
+for id in $ID_LIKE $ID; do break; done
+OS_ID=${id:-rhel}
+curl http://download.virtualbox.org/virtualbox/rpm/$OS_ID/virtualbox.repo | sudo tee /etc/yum.repos.d/virtualbox.repo
```
jonboulle: sudo curl -s -O /etc/....

functional/README.md
```diff
@@ -2,52 +2,98 @@
...
+### Centos/Fedora
 
-An example test session using coreos-vagrant follows. This assumes that go is available in `/home/core/go` and the fleet repository in `/home/core/fleet` on the target machine (the easiest way to achieve this is to use shared folders).
+**NOTE**: NFS and Vagrant doesn't work out of the box on CentOS 6.x
+
+#### Install Vagrant
+
+```sh
+sudo yum install -y git nfs-utils
+sudo reboot
```
jonboulle: why is this necessary? o_O

functional/Vagrantfile
reneengelhard: you mean installing go in the test sceript iself? yeah, we could do and actually that was the case before.

But it's cleaner to seperate the steps. Create the VM (and set it up) and actually running the tests. And installking go belongs to the first part so it was actually moved out of the test to a separate provisioner shell script.

functional/README.md
```diff
@@ -2,52 +2,98 @@
...
+wget -q https://www.virtualbox.org/download/oracle_vbox.asc -O- | sudo apt-key add -
+sudo apt-get update
+sudo apt-get install -y build-essential dkms
+sudo apt-get install -y VirtualBox-5.0
+#Previous VirtualBox (if you have problems with nested virtualization, more info here: https://www.virtualbox.org/ticket/14965)
+#sudo apt-get install -y VirtualBox-4.3
 ```
 
-Since the tests utilize `systemd-nspawn`, this needs to be invoked as sudo/root.
+### Centos/Fedora
```
jonboulle: CentOS

functional/README.md
```diff
@@ -2,52 +2,98 @@
...
 
-An example test session using coreos-vagrant follows. This assumes that go is available in `/home/core/go` and the fleet repository in `/home/core/fleet` on the target machine (the easiest way to achieve this is to use shared folders).
+**NOTE**: NFS and Vagrant doesn't work out of the box on CentOS 6.x
```
jonboulle: hence, CentOS >7.x must be used, or something

functional/Vagrantfile
jonboulle: No, I don't mean in the test script. I mean in the [user-data](https://github.com/coreos/fleet/pull/1430/files#diff-832e3e2830d9fbf1ffb2df429089afa0R1)

functional/config.rb
jonboulle: What do we really need from here? Just the shared folder?

kayrus: channel should be stable, 1 cpu, 512 memory.

functional/platform/nspawn.go
kayrus: Already added this info into README.md

functional/README.md
kayrus: Again, didn't find time to figure out why NFS can not be started. After reboot it works...

functional/README.md
kayrus: do I need to add this text into README.md?

functional/README.md
kayrus: -O doesn't work on centos

functional/README.md
kayrus: fixed

functional/README.md
jonboulle: Just remove this line.

functional/README.md
```diff
@@ -2,52 +2,98 @@
...
+sudo apt-get install -y VirtualBox-5.0
+#Previous VirtualBox (if you have problems with nested virtualization, more info here: https://www.virtualbox.org/ticket/14965)
+#sudo apt-get install -y VirtualBox-4.3
 ```
 
-Since the tests utilize `systemd-nspawn`, this needs to be invoked as sudo/root.
+### CentOS/Fedora
 
-An example test session using coreos-vagrant follows. This assumes that go is available in `/home/core/go` and the fleet repository in `/home/core/fleet` on the target machine (the easiest way to achieve this is to use shared folders).
+**NOTE**: NFS and Vagrant doesn't work out of the box on CentOS 6.x
```
jonboulle: Just change this to "CentOS >= 7.x  is required" or something

functional/Vagrantfile
jonboulle: Ping?

functional/config.rb
jonboulle: I am really struggling to agree that it's simpler to maintain the Vagrantfile in sync with that from coreos-vagrant - [_but not quite_](https://github.com/coreos/fleet/pull/1430/files#diff-a508e808c0fca80a337e9a8d22945e85R142) - rather than just port it to a much simpler Vagrantfile that fulfills our very particular purposes for testing here. 

In case it's not clear why I am pushing back on all these things: every file we add increases the complexity and burden of understanding and maintaining the codebase. I am asking lots of dumb questions here because even though I'm familiar with fleet and CoreOS it's still far from obvious to me why we really need all these parts, and e.g. where I should start making changes if I want to modify things. How can we make this easier for others approaching who have even less context?

jonboulle:
For the sake of moving forward:
i) squash commit history into something sensible
ii) land this
iii) please follow up the comments I've just added

functional/README.md
kayrus: it is not really required. someone can definitely configure this.

functional/Vagrantfile
kayrus: @jonboulle we can write oneshot unit file but vagrant will not wait for its result. `run-in-vagrant` has `vagrant up` and `vagrant ssh core-01 -c "sudo ~/fleet/functional/test"`. At the moment last command will be executed once vagrant up provisioning has been finished. If we will use systemd unit - tests will be started before go binary are unpacked.

functional/README.md
kayrus: Will add `**NOTE**: NFS and Vagrant doesn't work out of the box on CentOS 6.x, so it is recommended to use CentOS 7.x`

functional/README.md
kayrus: Ok

functional/config.rb
kayrus: Well. Let's assume we will add libvirt support for the original Vagrantfile. If we make all modifications (channel, memory, additional cases) it will be harder to merge new changes into this Vagrantfile.

Additional `config.vm.provision` string is an exception. Because without it we have to add more logic into our scripts (i.e. create lock file which will tell `sudo ./test` script that go binaries are downloading and wait until this lock file will be removed, then continue with the tests). If you are ok with these changes - I will do that.

PR #1431, opened 2016-02-12
tests fleetctl destroy behavior

steveeJ:
* Tests `fleetctl destroy` behavior
* Shows unexpected exit code behavior introduced via #1417

/cc @jonboulle

### DIFF
--- a/fleetctl/destroy_test.go
+++ b/fleetctl/destroy_test.go
@@ -1,0 +1,97 @@
+// Copyright 2014 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"testing"
+
+	"github.com/coreos/fleet/client"
+	"github.com/coreos/fleet/job"
+	"github.com/coreos/fleet/machine"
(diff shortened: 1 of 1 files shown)


jonboulle:
Nice, we should have these more generally. @tixxdz can you merge this into your  #1383 followup perhaps?

tixxdz:
@jonboulle yes I'm on it, @steveeJ thank you! but it seems that you have missed the point that users want the output to be 0 not 1 even for non existent units, yes I don't like it too, but the original discussion suggests this https://github.com/coreos/fleet/issues/1383 

So I'll fix up the test and merge it as followup for #1383

tixxdz:
@jonboulle @steveeJ your patch and some others will be queued up later from this branch https://github.com/endocode/fleet/commits/tixxdz/fleet-code-cleaning-v2%2B1 

Thanks

steveeJ:
Actually I wanted to make a point that it doesn't error out as I thought it should, my bad. I'm a supporter of indicating errors but I can see why it would bother people. How about introduce exit codes that represent the kind of error behavior. Similar to the well-established errno style. 

If not, should I update the PR to with the test to expect a 0 or are you picking up the commit?

antrik:
@steveeJ note that the discussion in https://github.com/coreos/fleet/pull/1417 mentions the possibility of adding a command line option for explicitly requesting an error return in case some of the units to be destroyed are missing. If you have an actual use case for that behaviour, I guess it's best to submit a feature request (or PR :-) ) referencing that discussion.

Without any option, the behaviour should indeed be to just ignore missing units (which it was doing in the past, as pointed out in https://github.com/coreos/fleet/issues/1383) -- if there is nothing to destroy, the desired state is already reached, which is not considered an error...

tixxdz:
@steveeJ yes as explained by @antrik if you want that behaviour you can submit a new PR for it!

And thanks for the patch ;-) I've already picked it up here: https://github.com/endocode/fleet/commits/tixxdz/fleet-code-cleaning-v2%2B1  and added more changes on top. With functional tests later we can copy some of these, that would really help us. Thanks again!

PR #1433, opened 2016-02-16
fleetctl: {load|unload|start|stop} and get unit code cleanups

tixxdz:
This patches set aims to make the code more consistent and robust. It does not change the current behaviour just improves the code, consolidate it, more debug logs and code comments.

The main goal is to let stop, start, load and unload to share the same code base and probably avoid any corner case or bug that may related to one of these commands.

* Some logic can be easily shared later when we add stop to destroy command to solve  https://github.com/coreos/fleet/issues/1000  

* There were some TODOs in the code base that I cleaned up with the last two patches. These patches will make it easy for us to implement later https://github.com/coreos/fleet/pull/1295

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -746,6 +746,56 @@ func setTargetStateOfUnits(units []string, state job.JobState) ([]*schema.Unit, 
 	return triggered, nil
 }
 
+// getBlockAttempts gets the correct value of how many attempts to try
+// before giving up on an operation.
+// It returns a negative value which means try forever, if zero is
+// returned then do not make any attempt, and if a positive value is
+// returned then try up to that value
+func getBlockAttempts() int {
+	// By default we wait forever
+	var attempts int = -1
+
+	// Up to BlockAttempts
+	if sharedFlags.BlockAttempts > 0 {
+		attempts = sharedFlags.BlockAttempts
+	}
+
+	// NoBlock we do not wait
+	if sharedFlags.NoBlock {
+		attempts = 0
+	}
+
+	return attempts
+}
+
+// tryWaitForUnitStates tries to wait for units to reach the desired state.
+// It takes 5 arguments, the units to wait for, the desired state, the
+// desired JobState, how many attempts before timing out and a writer
+// interface.
+// tryWaitForUnitStates polls each of the indicated units until they
+// reach the desired state. If maxAttempts is zero, then it will not
+// wait, it will assume that all units reached their desired state.
+// If maxAttempts is negative tryWaitForUnitStates will retry forever, and
+// if it is greater than zero, it will retry up to the indicated value.
+// It returns 0 on success or 1 on errors.
+func tryWaitForUnitStates(units []string, state string, js job.JobState, maxAttempts int, out io.Writer) (ret int) {
+	// We do not wait just assume we reached the desired state
+	if maxAttempts == 0 {
+		for _, name := range units {
+			stdout("Triggered unit %s %s", name, state)
+		}
+		return
+	}
+
+	errchan := waitForUnitStates(units, js, maxAttempts, out)
+	for err := range errchan {
+		stderr("Error waiting for units: %v", err)
+		ret = 1
+	}
+
+	return
+}
+
 // waitForUnitStates polls each of the indicated units until each of their
 // states is equal to that which the caller indicates, or until the
 // polling operation times out. waitForUnitStates will retry forever, or
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -483,6 +483,13 @@ func getChecker() *ssh.HostKeyChecker {
 	return ssh.NewHostKeyChecker(keyFile)
 }
 
+// getUnitFile attempts to get a UnitFile configuration
+// It takes a unit file name as a parameter and tries first to lookup
+// the unit from the local disk. If it fails, it checks if the provided
+// file name may reference an instance of a template unit, if so, it
+// tries to get the template configuration either from the registry or
+// the local disk.
+// It returns a UnitFile configuration or nil; and any error ecountered
 func getUnitFile(file string) (*unit.UnitFile, error) {
 	var uf *unit.UnitFile
 	name := unitNameMangle(file)
@@ -498,14 +505,22 @@ func getUnitFile(file string) (*unit.UnitFile, error) {
 		}
 	} else {
 		// Otherwise (if the unit file does not exist), check if the
-		// name appears to be an instance unit, and if so, check for
-		// a corresponding template unit in the Registry or disk.
+		// name appears to be an instance of a template unit
+		info := unit.NewUnitNameInfo(name)
+		if info == nil {
+			return nil, fmt.Errorf("error extracting information from unit name %s", name)
+		} else if !info.IsInstance() {
+			return nil, fmt.Errorf("unable to find Unit(%s) in Registry or on filesystem", name)
+		}
+
+		// If it is an instance check for a corresponding template
+		// unit in the Registry or disk.
 		// If we found a template unit, later we create a
 		// near-identical instance unit in the Registry - same
 		// unit file as the template, but different name
-		uf, err = getUnitFileFromTemplate(file)
+		uf, err = getUnitFileFromTemplate(info, file)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed getting Unit(%s) from template: %v", file, err)
 		}
 	}
 
@@ -527,21 +542,12 @@ func getUnitFromFile(file string) (*unit.UnitFile, error) {
 	return unit.NewUnitFile(string(out))
 }
 
-// getUnitFileFromTemplate checks if the name appears to be an instance unit
-// and gets its corresponding template unit from the registry or local disk
+// getUnitFileFromTemplate attempts to get a Unit from a template unit that
+// is either in the registry or on the file system
+// It takes two arguments, the template information and the unit file name
 // It returns the Unit or nil; and any error encountered
-func getUnitFileFromTemplate(arg string) (*unit.UnitFile, error) {
+func getUnitFileFromTemplate(uni *unit.UnitNameInfo, fileName string) (*unit.UnitFile, error) {
 	var uf *unit.UnitFile
-	name := unitNameMangle(arg)
-
-	// Check if the name appears to be an instance unit, and if so,
-	// check for a corresponding template unit in the Registry
-	uni := unit.NewUnitNameInfo(name)
-	if uni == nil {
-		return nil, fmt.Errorf("error extracting information from unit name %s", name)
-	} else if !uni.IsInstance() {
-		return nil, fmt.Errorf("unable to find Unit(%s) in Registry or on filesystem", name)
-	}
 
 	tmpl, err := cAPI.Unit(uni.Template)
 	if err != nil {
@@ -549,20 +555,20 @@ func getUnitFileFromTemplate(arg string) (*unit.UnitFile, error) {
 	}
 
 	if tmpl != nil {
-		warnOnDifferentLocalUnit(arg, tmpl)
+		warnOnDifferentLocalUnit(fileName, tmpl)
 		uf = schema.MapSchemaUnitOptionsToUnitFile(tmpl.Options)
 		log.Debugf("Template Unit(%s) found in registry", uni.Template)
 	} else {
 		// Finally, if we could not find a template unit in the Registry,
 		// check the local disk for one instead
-		file := path.Join(path.Dir(arg), uni.Template)
-		if _, err := os.Stat(file); os.IsNotExist(err) {
-			return nil, fmt.Errorf("unable to find Unit(%s) or template Unit(%s) in Registry or on filesystem", name, uni.Template)
+		filePath := path.Join(path.Dir(fileName), uni.Template)
+		if _, err := os.Stat(filePath); os.IsNotExist(err) {
+			return nil, fmt.Errorf("unable to find template Unit(%s) in Registry or on filesystem", uni.Template)
 		}
 
-		uf, err = getUnitFromFile(file)
+		uf, err = getUnitFromFile(filePath)
 		if err != nil {
-			return nil, fmt.Errorf("failed getting template Unit(%s) from file: %v", uni.Template, err)
+			return nil, fmt.Errorf("unable to load template Unit(%s) from file: %v", uni.Template, err)
 		}
 	}
 
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -483,6 +483,36 @@ func getChecker() *ssh.HostKeyChecker {
 	return ssh.NewHostKeyChecker(keyFile)
 }
 
+func getUnitFile(file string) (*unit.UnitFile, error) {
+	var uf *unit.UnitFile
+	name := unitNameMangle(file)
+
+	log.Debugf("Looking up for Unit(%s) or its corresponding template", name)
+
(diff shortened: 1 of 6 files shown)


fleetctl/fleetctl.go
```diff
@@ -482,6 +482,36 @@ func getChecker() *ssh.HostKeyChecker {
 	return ssh.NewHostKeyChecker(keyFile)
 }
 
+func getUnit(file string) (*unit.UnitFile, error) {
```
jonboulle: getUnitFile

fleetctl/fleetctl.go
```diff
@@ -496,6 +526,48 @@ func getUnitFromFile(file string) (*unit.UnitFile, error) {
 	return unit.NewUnitFile(string(out))
 }
 
+// getUnitFromTemplate checks if the name appears to be an instance unit
+// and gets its corresponding template unit from the registry or local disk
+// It returns the Unit or nil; and any error encountered
+func getUnitFromTemplate(arg string) (*unit.UnitFile, error) {
```
jonboulle: getUnitFileFromTemplate

fleetctl/fleetctl.go
```diff
@@ -745,6 +782,62 @@ func setTargetStateOfUnits(units []string, state job.JobState) ([]*schema.Unit,
 	return triggered, nil
 }
 
+// getArgMaxAttempts gets the correct value of how many attempts to try
```
jonboulle: maybe getBlockAttempts

jonboulle: Could just pass it sharedFlags

fleetctl/fleetctl.go
tixxdz: done.

fleetctl/fleetctl.go
tixxdz: done.

fleetctl/fleetctl.go
tixxdz: yes getBlockAttempts() but no arguments, I didn't want to make sharedFlags a type, so it will just access its fields as it is done inside other part of the code inside fleetctl.go. It's better this way.

tixxdz:
PR updated, thanks

fleetctl/fleetctl.go
```diff
@@ -745,6 +782,56 @@ func setTargetStateOfUnits(units []string, state job.JobState) ([]*schema.Unit,
...
 }
 
+// getBlockAttempts gets the correct value of how many attempts to try
+// before giving up on an operation.
+// It returns a negative value which means try forever, if zero is
+// returned then do not make any attempt, and if a positive value is
+// returned then try up to that value
+func getBlockAttempts() int {
+	// By default we wait forever
+	var attempts int = -1
```
jonboulle: Where do you do anything with this?

fleetctl/unload.go
```diff
@@ -36,6 +36,8 @@ func init() {
 }
 
 func runUnloadUnit(args []string) (exit int) {
+	attempts := getBlockAttempts()
```
jonboulle: Here and elsewhere, no need for a separate variable, just inline it in the tryWaitForUnitStates call

fleetctl/fleetctl.go
```diff
@@ -482,6 +482,36 @@ func getChecker() *ssh.HostKeyChecker {
...
+	} else {
+		// Otherwise (if the unit file does not exist), check if the
+		// name appears to be an instance unit, and if so, check for
+		// a corresponding template unit in the Registry or disk
+		uf, err = getUnitFileFromTemplate(file)
+		if err != nil {
+			return nil, err
+		}
+
+		// If we found a template unit, create a near-identical instance unit in
```
jonboulle: ?

jonboulle: this comment has kind of lost the context

fleetctl/fleetctl.go
```diff
@@ -745,6 +782,56 @@ func setTargetStateOfUnits(units []string, state job.JobState) ([]*schema.Unit,
...
+// getBlockAttempts gets the correct value of how many attempts to try
+// before giving up on an operation.
+// It returns a negative value which means try forever, if zero is
+// returned then do not make any attempt, and if a positive value is
+// returned then try up to that value
+func getBlockAttempts() int {
+	// By default we wait forever
+	var attempts int = -1
+
+	// Up to BlockAttempts
```
jonboulle: I think you can just kill this comment and that on 799

fleetctl/fleetctl.go
tixxdz: Yes I set it to negative value to not set it to zero, this way I can return immediately here https://github.com/endocode/fleet/commit/75b60886e6f698845db93e1cd9db89e0ebcb4398#diff-197bb752bf0f52567eb6b0023ea65e40R782

And lalso to allow us to continue with the previous logic inside "waitForUnitStates()" 0 or  any negative value. I didn't touch "waitForUnitStates()" since it's used in some other parts of the code, so I only converted new calls keeping the same semantics from a user experience. Yes the problem is we have two values inside the code blockattempts and noblock that interfere which other that's why I think we should convert this into one value that allows us to be more consistent in each part of the code. Perhaps later other parts can be converted too, but now I only changed fleetctl commands.

fleetctl/unload.go
tixxdz: Ok

fleetctl/fleetctl.go
tixxdz: Improved based on your last comments, two new patches up so it will be easy to followup.

Thank you!

fleetctl/fleetctl.go
tixxdz: This also was handled. thanks!

fleetctl/fleetctl.go
```diff
@@ -482,6 +482,36 @@ func getChecker() *ssh.HostKeyChecker {
 	return ssh.NewHostKeyChecker(keyFile)
 }
 
+func getUnitFile(file string) (*unit.UnitFile, error) {
+	var uf *unit.UnitFile
+	name := unitNameMangle(file)
+
+	log.Debugf("Looking up for Unit(%s) or its corresponding template", name)
```
jonboulle: Do we need these debug logs? do they show up by default?

tixxdz:
The functional tests pass. Thanks!

fleetctl/fleetctl.go
tixxdz: I assume they will show up when debug is set?! just followed up the current behaviour.
I would say yeh we may need them especially with all the concurrency problems, having these would help. Now if they show up by default then yeh that's another problem, that perhaps we should fix in another place ?

fleetctl/fleetctl.go
jonboulle: I was just a bit confused about the inconsistency between BlockAttempts=0 on the command line referring to waiting forever, but getBlockAttempts() returning -1 to indicate the same.

jonboulle: i.e. implicitly relying on https://github.com/coreos/fleet/pull/1433/files#diff-197bb752bf0f52567eb6b0023ea65e40R861 catching both cases. Maybe instead we could standardise on blockattempts=-1 == non blocking and change the check aat https://github.com/coreos/fleet/pull/1433/files#diff-197bb752bf0f52567eb6b0023ea65e40R817

fleetctl/fleetctl.go
jonboulle: Don't assume, test!

It's okay to add more debugging information but it should be more thoughtful - e.g. why should we log both here https://github.com/coreos/fleet/pull/1433/files#diff-197bb752bf0f52567eb6b0023ea65e40R553 and here https://github.com/coreos/fleet/pull/1433/files#diff-197bb752bf0f52567eb6b0023ea65e40R511 if a template lookup succeeds?

fleetctl/fleetctl.go
```diff
@@ -496,6 +526,48 @@ func getUnitFromFile(file string) (*unit.UnitFile, error) {
...
+	var uf *unit.UnitFile
+	name := unitNameMangle(arg)
+
+	// Check if the name appears to be an instance unit, and if so,
+	// check for a corresponding template unit in the Registry
+	uni := unit.NewUnitNameInfo(name)
+	if uni == nil {
+		return nil, fmt.Errorf("error extracting information from unit name %s", name)
+	} else if !uni.IsInstance() {
+		return nil, fmt.Errorf("unable to find Unit(%s) in Registry or on filesystem", name)
```
jonboulle: this error (and that on line 559) are actually bleeding into parent scope - this function is only checking for templates but the error is referencing filesystem looks (which just implicitly happen to have happened and failed before we're called in this scope). Need to clean this up somehow, not sure off the top of my head

fleetctl/fleetctl.go
tixxdz: Ok, so yes we log in both cases since we have two different paths https://github.com/coreos/fleet/pull/1433/files#diff-197bb752bf0f52567eb6b0023ea65e40R550 if the template lookup from the registry succeed we log it otherwise if we fail and take the second path, we end up with the lookup from disk in getUnitFromFile() which will log for us. That's why I had to log inside getUnitFileFromTemplate(), it also falls back to getUnitFromFile() in case the template is not in the registry.
Now why I enclosed the whole logic with more debug messages like this one https://github.com/coreos/fleet/pull/1433/files#diff-197bb752bf0f52567eb6b0023ea65e40R511 the reason is in getUnitFromFile(). That function is used in other places, I don't want to update it and it logs before even succeeding here: https://github.com/coreos/fleet/blob/master/fleetctl/fleetctl.go#L494 but it may fail just in line 496 in NewUnitFile() during deserialize or any other error which is good since we catch both cases: the unit was found in disk but we failed to instantiate it, so we didn't reach this code path at all: https://github.com/coreos/fleet/pull/1433/files#diff-197bb752bf0f52567eb6b0023ea65e40R511

IMHO all these debug messages will be really helpful! thanks you!

fleetctl/fleetctl.go
tixxdz: Hmm, ok so I got you a bit on this one, but yeh not sure what to do, certainly removing errors was not an option for me, so ?!

fleetctl/fleetctl.go
jonboulle: ok, I'm not completely sure about it all together but we can roll forward and clean up later if it turns out to be too noisy. One final thing though: change "Looking up for" to "Looking for" in this message. thanks

fleetctl/fleetctl.go
tixxdz: Yes exactly relying on that one: https://github.com/coreos/fleet/pull/1433/files#diff-197bb752bf0f52567eb6b0023ea65e40R861  to catch both of them.

For your comment "Maybe instead we could standardise on blockattempts=-1 == non blocking and change the check aat https://github.com/coreos/fleet/pull/1433/files#diff-197bb752bf0f52567eb6b0023ea65e40R817" yes I don't have a strong opinion here, I'm just used to negative values meaning retry or block forever for polling functions. Ok will do that change then, thanks!

fleetctl/fleetctl.go
tixxdz: Alright

fleetctl/fleetctl.go
jonboulle: It's less than ideal, but you could create some typed errors like errNoTemplateFoundInRegistry , and return them from this function, then do a switch on the error from the outer scope, and assemble the error message there as appropriate.

fleetctl/fleetctl_test.go
```diff
@@ -117,6 +117,51 @@ func TestUnitNameMangle(t *testing.T) {
 	}
 }
 
+func TestGetBlockAttempts(t *testing.T) {
+	oldNoBlock := sharedFlags.NoBlock
+	oldBlockAttempts := sharedFlags.BlockAttempts
+	sharedFlags.NoBlock = true
```
jonboulle: Please turn this into a test table.

unit/unit.go
```diff
@@ -25,6 +25,16 @@ import (
 	"github.com/coreos/fleet/Godeps/_workspace/src/github.com/coreos/go-systemd/unit"
 )
 
+const (
```
jonboulle: Errors should start lowercase, and I'm not sure what this slice approach buys you. Better:
```
var (
       ErrNameInfo = errors.New("unable to extract ...")
       ...
)
```
But it doesn't make sense to define the errors in this package and return them from the other one..

fleetctl/fleetctl_test.go
tixxdz: done

unit/unit.go
tixxdz: thx @jonboulle put that one on my stupid go skills, so yes I updated it now.

fleetctl/fleetctl.go
```diff
@@ -614,45 +692,12 @@ func lazyCreateUnits(args []string) error {
...
-				warnOnDifferentLocalUnit(arg, tmpl)
-				uf = schema.MapSchemaUnitOptionsToUnitFile(tmpl.Options)
-			}
-
-			// If we found a template unit, create a near-identical instance unit in
-			// the Registry - same unit file as the template, but different name
+		// Assume that the name references a local unit file on
+		// disk or if it is an instance unit and if so get its
+		// corresponding unit
+		uf, err := getUnitFile(arg)
```
jonboulle: Unfortunately this hasn't really solved https://github.com/coreos/fleet/pull/1433#discussion_r53453416 at all. Now look what happens if I do a nonexistent unit:
```
core@coreos ~ $ ./fleetctl  submit nonexistent.file
Error creating units: failed getting template Unit(nonexistent.file.service) info: Not an instance of a template unit
```

tixxdz: @jonboulle ok, will try to improve that, but we have to return some errors messages here, basically this code path also calls into go-systemd too to parse units, and ignoring those errors is IMHO not the best way to go. Will try to make it a bit generic here.

tixxdz: `$ ./bin/fleetctl  submit nonexistent.file
Error creating units: failed getting Unit(nonexistent.file.service) info: unable to find Unit in Registry or on filesystem`

`$ ./bin/fleetctl  submit nonexistent@.file
Error creating units: failed getting Unit(nonexistent@.file.service) from template: unable to find Unit(nonexistent@.service) in Registry or on filesystem`

fleetctl/fleetctl.go
```diff
@@ -496,6 +532,50 @@ func getUnitFromFile(file string) (*unit.UnitFile, error) {
 	return unit.NewUnitFile(string(out))
 }
 
+func getUnitInstanceInfo(name string) (*unit.UnitNameInfo, error) {
```
jonboulle: Let's revert this tot he previous logic which just inlined this at the callsite - then we'll get the old error message back which is a little more direct, rather than wrapping like on https://github.com/coreos/fleet/pull/1433/files#diff-197bb752bf0f52567eb6b0023ea65e40R503

fleetctl/fleetctl.go
```diff
@@ -482,6 +482,42 @@ func getChecker() *ssh.HostKeyChecker {
 	return ssh.NewHostKeyChecker(keyFile)
 }
 
+func getUnitFile(file string) (*unit.UnitFile, error) {
```
jonboulle: Please add a docstring clarifying when exactly this should be used: in particular, after the registry has already been checked for a unit of the given name. This is to mitigate my comment @ https://github.com/coreos/fleet/pull/1433#discussion_r53453416 - i.e. that this function can return an error which references the registry.

tixxdz:
@jonboulle Investigating why semaphore fails, all comments were handled for this one and all functional tests pass here.... but semaphore fails for another error I guess...

PR #1435, opened 2016-02-22
[WIP] fleetctl destroy fix and more tests 

tixxdz:
Fix for fleetctl destroy command and more tests.

@jonboulle when you take a look please start from the patch: "fleetctl: avoid hard coding sleep time values inside calls" since this branch is on top of https://github.com/coreos/fleet/pull/1433

### DIFF
--- a/fleetctl/destroy.go
+++ b/fleetctl/destroy.go
@@ -71,7 +71,7 @@ func runDestroyUnits(args []string) (exit int) {
 				if u == nil {
 					break
 				}
-				time.Sleep(500 * time.Millisecond)
+				time.Sleep(defaultSleepTime)
 			}
 		}
 
--- a/fleetctl/destroy.go
+++ b/fleetctl/destroy.go
@@ -42,6 +42,11 @@ func runDestroyUnits(args []string) (exit int) {
 	for _, v := range units {
 		err := cAPI.DestroyUnit(v.Name)
 		if err != nil {
+			// If unit does not exist do not error out
+			u, _ := cAPI.Unit(v.Name)
+			if u == nil {
+				continue
+			}
 			stderr("Error destroying units: %v", err)
 			exit = 1
 			continue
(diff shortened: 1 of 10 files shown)


jonboulle:
General comment: it would be immensely easier to review and land these more quickly if the PRs were in bite sized chunks. It's unclear to me that this needs to be based on #1433

jonboulle:
e.g.  https://github.com/endocode/fleet/commit/fd173a01ced24065a3824db9919c1a04a3a6a8a5 could be in its own PR, as could say 
https://github.com/endocode/fleet/commit/59fcef31e04b8a9c3e2f0d4719574a476d169548  with https://github.com/endocode/fleet/commit/69f6a1b86632433ea1e95061618905c3a946e7c0

tixxdz:
Alright! noted!

For your comment: "hmm? now we're ignoring errors.."  these are just unit tests some logic could be used later for functional ones, but I wanted to standardize fleectl commands first, or am I missing something ?

Thanks!

jonboulle:
@tixxdz see [here](https://github.com/endocode/fleet/commit/76d2f1a08ec67cebca7b397db221f747eb10edd2#commitcomment-16241104) where the actual comment is. What if this Unit() call returns an error?

tixxdz:
@jonboulle got you! So the new PR is here: https://github.com/coreos/fleet/pull/1439

so yes I added a new function IsErrorUnitNotFound() part of the client package but not an exported cAPI, if later we want to check for more error codes in that path we may consider something IsErrorCode(err error, code int)... and export it.

For the other patches that can be put on there own PR, yes I do agree, but at the same time these changes were done in the same process of improving this code and adding more tests, some of these functions are new and are only used by these new tests.

Thanks!

tixxdz:
Closed in favor of https://github.com/coreos/fleet/pull/1439

PR #1442, opened 2016-02-26
Add --replace flag for load, start and submit

### LINKED ISSUE #760: Redeploy unit in place

klausenbusk:
This commit allow people to "replace" a unit in the registry, even through it already exits.

Dunno if it works, I haven't tested it, also I need some help with the description of the `--force` option.

Fix #760

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -106,6 +106,7 @@ var (
 		BlockAttempts int
 		Fields        string
 		SSHPort       int
+		Replace       bool
 	}{}
 
 	// used to cache MachineStates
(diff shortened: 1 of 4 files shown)


antrik:
This is an interesting approach. All conversation and pull requests regarding this so far have been about adding a separate command -- but adding a --force flag to the existing commands might indeed be preferable?

klausenbusk:
Closed in favor of #1509 by @tixxdz

tixxdz:
@klausenbusk thank you we used the same "--replace" switch but it needed more work.

Issue #1369, opened 2015-10-08
Panic on graceful shutdown (close of closed channel)


balboah:
For fleet version 0.10.2 on CoreOS Stable branch, there seems to be two problems:
- Might close an already closed channel upon graceful shutdown
- The ever growing list of goroutines when having etcd connection issues, almost **two million** in this example

```
... snip snip (a few days of 503) ...

Oct 08 07:52:21 fleetd[127879]: INFO client.go:298: Response 503 Service Unavailable from http://127.0.0.1:4001/ unusable
Oct 08 07:52:21 fleetd[127879]: ERROR client.go:214: Unable to get result for {Create /_coreos.com/fleet/machines/0553e988b3e14a6ebd7d3108a2ab4d8e/object}, retrying in 1s
Oct 08 07:52:22 fleetd[127879]: INFO client.go:298: Response 503 Service Unavailable from http://127.0.0.1:4001/ unusable
Oct 08 07:52:22 fleetd[127879]: ERROR client.go:214: Unable to get result for {Create /_coreos.com/fleet/machines/0553e988b3e14a6ebd7d3108a2ab4d8e/object}, retrying in 1s
Oct 08 07:52:23 fleetd[127879]: INFO client.go:298: Response 503 Service Unavailable from http://127.0.0.1:4001/ unusable
Oct 08 07:52:23 fleetd[127879]: ERROR client.go:214: Unable to get result for {Create /_coreos.com/fleet/machines/0553e988b3e14a6ebd7d3108a2ab4d8e/object}, retrying in 1s
Oct 08 07:53:24 systemd[1]: Stopping fleet daemon...
Oct 08 07:53:24 fleetd[127879]: INFO fleet.go:105: Gracefully shutting down
Oct 08 07:53:24 fleetd[127879]: panic: close of closed channel
Oct 08 07:53:24 fleetd[127879]: goroutine 1 [running]:
Oct 08 07:53:24 fleetd[127879]: main.func·002()
Oct 08 07:53:24 fleetd[127879]: /build/amd64-usr/var/tmp/portage/app-admin/fleet-0.10.2-r1/work/gopath/src/github.com/coreos/fleet/fleetd/fleet.go:106 +0x6b
Oct 08 07:53:24 fleetd[127879]: main.listenForSignals(0xc2080b0ae0)
[...]

PR #1067, opened 2014-12-17
panic due to double-close of channel

jonboulle:
I'm not sure what's going on here, but clearly fleet should never try to close an already-closed channel: From https://github.com/coreos/fleet/issues/1044:

```
Dec 02 18:26:05 ip-172-31-24-115.eu-west-1.compute.internal fleetd[570]: ERROR server.go:169: Server monitor triggered: Monitor timed out before successful heartbeat
Dec 02 18:26:05 ip-172-31-24-115.eu-west-1.compute.internal fleetd[570]: panic: runtime error: close of closed channel
Dec 02 18:26:05 ip-172-31-24-115.eu-west-1.compute.internal fleetd[570]: goroutine 11048 [running]:
Dec 02 18:26:05 ip-172-31-24-115.eu-west-1.compute.internal fleetd[570]: runtime.panic(0x77e400, 0x9bf155)
Dec 02 18:26:05 ip-172-31-24-115.eu-west-1.compute.internal fleetd[570]: /usr/lib/go/src/pkg/runtime/panic.c:279 +0xf5
Dec 02 18:26:05 ip-172-31-24-115.eu-west-1.compute.internal fleetd[570]: github.com/coreos/fleet/server.(*Server).Monitor(0xc208455560)
Dec 02 18:26:05 ip-172-31-24-115.eu-west-1.compute.internal fleetd[570]: /build/amd64-usr/var/tmp/portage/app-admin/fleet-0.8.3/work/fleet-0.8.3/gopath/src/github.com/coreos/fleet/server/server.go:171 +0xfb
Dec 02 18:26:05 ip-172-31-24-115.eu-west-1.compute.internal fleetd[570]: created by github.com/coreos/fleet/server.(*Server).Run
Dec 02 18:26:05 ip-172-31-24-115.eu-west-1.compute.internal fleetd[570]: /build/amd64-usr/var/tmp/portage/app-admin/fleet-0.8.3/work/fleet-0.8.3/gopath/src/github.com/coreos/fleet/server/server.go:152 +0x10e
Dec 02 18:26:05 ip-172-31-24-115.eu-west-1.compute.internal fleetd[570]: goroutine 16 [chan receive]:
Dec 02 18:26:05 ip-172-31-24-115.eu-west-1.compute.internal fleetd[570]: main.listenForSignals(0xc2080a5650)
[...]

PR #1452, opened 2016-03-01
Shutdown fixes/improvements

antrik:
Rebased from https://github.com/coreos/fleet/pull/1067 , and added a test case to verify the fix.

Should fix https://github.com/coreos/fleet/issues/1369

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -57,11 +57,11 @@ func (a *Agent) MarshalJSON() ([]byte, error) {
 
 // Heartbeat updates the Registry periodically with an acknowledgement of the
 // Jobs this Agent is expected to be running.
-func (a *Agent) Heartbeat(stop chan bool) {
+func (a *Agent) Heartbeat(stop <-chan struct{}) {
 	a.heartbeatJobs(a.ttl, stop)
 }
 
-func (a *Agent) heartbeatJobs(ttl time.Duration, stop chan bool) {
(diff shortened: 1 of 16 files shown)


antrik:
Meh, the new test seems to fail intermittently now and then :-(

jonboulle:
How do they fail?

antrik:
@jonboulle in order to set the stage for triggering the condition we want to test, we first need to cut the network connection to etcd, and wait for the "Monitor triggered" message to show up in the journal. Sometimes this just waits forever. Don't know yet why this happens: whether we just miss the message for some reason; or whether fleet sometimes behaves differently, and the message is indeed not generated at all...

antrik:
I should probably note that this seems to happen occasionally both with the original and with the reworked code.

antrik:
OK, it turns out it was only my method of monitoring the journal that was unreliable -- with a different approach it appears to work better.

While at it, I added two extra commits to this PR, that are not strictly necessary, but related: one reduces the agent_ttl for fleet to speed up the new test; the other is a minor cleanup to the fleet code changed in this PR, which I stumbled upon while working on this stuff.

tixxdz:
@antrik could you re-push again so the semaphore runs again ?

antrik:
I guess Semaphore hates this PR :-( Let's try a new one...

tixxdz:
Ok this one was closed in favor of #1496

PR #1457, opened 2016-03-02
Pass on test script arguments to "go test" command

antrik:
This enables running only selected tests -- and maybe other useful things too...


tixxdz:
@jonboulle lgtm

Issue #1443, opened 2016-02-26
usability: show warning for the fleetctl subcommands


kayrus:
Current behavior:

```
$ fleetctl destroy # start/stop/load/any fleet subcommand
# nothing is here
$ echo $?
0
```

Desired behavior:

```
$ fleetctl destroy # start/stop/load/any fleet subcommand
Too few arguments.
$ echo $?
1
```

kayrus:
relates https://github.com/coreos/fleet/issues/1445

tixxdz:
running: `fleetctl journal` 
`One unit file must be provided.`

We should just be consistent and of course tell the user that he must supply the right arguments.

jonboulle:
Following previous discussions about the semantics of `fleetctl destroy <nonexistent unit>`, and for example #1417, it's not _immediately_ obvious to me that `fleetctl destroy` should error. We need to justify making a backwards-incompatible change here

jonboulle:
How about we just write something to stderr saying "No units given"

kayrus:
and 0 exit code? ok.

PR #1460, opened 2016-03-03
fleetctl: more tests for start load command and template path

tixxdz:
Add more tests for fleetctl start and load commands, and make sure that the template unit path also works as expected. While we are it improve fleetctl test helper functions.

PR #1461, opened 2016-03-03
fleetctl: check for empty input unit strings

dongsupark:
If no unit name is given for fleetctl commands like destroy, load,
start, stop, submit, and unload, then return with exit code 0,
printing out a soft warning message to stderr.

Fixes: https://github.com/coreos/fleet/issues/1443

/cc @kayrus @tixxdz @jonboulle

### DIFF
--- a/fleetctl/destroy.go
+++ b/fleetctl/destroy.go
@@ -35,6 +35,11 @@ Destroyed units are impossible to start unless re-submitted.`,
 }
 
 func runDestroyUnits(args []string) (exit int) {
+	if len(args) == 0 {
+		stderr("No units given")
+		return 0
+	}
+
 	units, err := findUnits(args)
 	if err != nil {
 		stderr("%v", err)
--- a/fleetctl/destroy_test.go
+++ b/fleetctl/destroy_test.go
@@ -63,7 +63,7 @@ func TestRunDestroyUnits(t *testing.T) {
 		var wg sync.WaitGroup
 		errchan := make(chan error)
 
-		cAPI = newFakeRegistryForCommands(unitPrefix, len(r.units))
+		cAPI = newFakeRegistryForCommands(unitPrefix, len(r.units), false)
 
 		wg.Add(2)
 		go func() {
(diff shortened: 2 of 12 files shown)


kayrus:
@dongsupark can you put these checks somewhere in one place. these copy-pastes look terrible.

dongsupark:
@kayrus I updated the branch dongsu/error-empty-units.
I wrote a simple wrapper in pkg/args.go, which can be called from each fleetctl command.

tixxdz:
@jonboulle lgtm

jonboulle:
uh, this helper is way overkill. please just revert to the inline check...

jonboulle:
```
	if pkg.IsArgsEmpty(args) {
		stderr("No units given")
		return 0
	}
```
is no simpler than
```
	if len(args) == 0 {
		stderr("No units given")
		return 0
	}
```

kayrus:
@jonboulle is it possible to check arguments inside `init` function?

jonboulle:
Why do you need to check inside init? In the context where you're checking, you've just been passed them.

dongsupark:
@jonboulle all right. I reverted it to the 1st version. So there's no helper in pkg/args.go.

jonboulle:
Thanks! Could you add a test for this? :-)

dongsupark:
@jonboulle Sure, I can.
Actually a PR by @tixxdz https://github.com/coreos/fleet/pull/1460 already contains a set of necessary tests for start and load. I would rather wait for the PR to be first merged, and add my tests on top of that.

fleetctl/destroy.go
```diff
@@ -35,6 +35,13 @@ Destroyed units are impossible to start unless re-submitted.`,
 }
 
 func runDestroyUnits(args []string) (exit int) {
+	// Even if no unit is given by user, it should not exit with error,
```
jonboulle: I don't think the comment is necessary. Please just remove it here and everywhere else.

dongsupark:
Updated the tree as @jonboulle suggested.
I erase the comments as suggested.
I also added simple tests for destroy, load, start, stop, and unload.
Submit test is missing in this PR, because there's no fleetctl/submit_test.go.
So I created a new test submit_test.go that is covered in #1473.

tixxdz:
@dongsupark LGTM, but could you please create a new branch then force push on top of this so semaphore could run again ?

dongsupark:
Closed in favor of #1475

PR #1465, opened 2016-03-04
fleetctl:test: {destroy|stop|unload} if the test fails then return as early as possible

tixxdz:
If the tests fail then return, and avoid printing other confusing error messages


jonboulle:
LGTM

PR #1472, opened 2016-03-07
fleet:functional-test: make Fleetctl() start print stdout and stderr on failures

tixxdz:
If functional tests fail in Fleetctl() start for some units, then just print directly as much information as possible. This will save some time when errors are just obvious ones.


jonboulle:
LGTM

PR #1474, opened 2016-03-07
fleetctl: {load|unload|start|stop} and get unit code cleanups 

tixxdz:
This patches set aims to make the code more consistent and robust. It does not change the current behaviour just improves the code, consolidate it, more debug logs and code comments.

The main goal is to let stop, start, load and unload to share the same code base and probably avoid any corner case or bug that may related to one of these commands.

Some logic can be easily shared later when we add stop to destroy command to solve #1000

There were some TODOs in the code base that I cleaned up with the last two patches. These patches will make it easy for us to implement later #1295

This new PR is to trigger semaphoreci.

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -746,6 +746,56 @@ func setTargetStateOfUnits(units []string, state job.JobState) ([]*schema.Unit, 
 	return triggered, nil
 }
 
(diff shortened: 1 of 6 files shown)


tixxdz:
closing in favor of the original one, since semaphore did trigger again... https://github.com/coreos/fleet/pull/1433

PR #1476, opened 2016-03-07
tests: minor improvements for run-in-qemu script

### LINKED ISSUE #1471: tests: improve run-in-qemu script
1) check coreos signature for the image (https://github.com/coreos/docs/issues/760)
2) handle race condition when ssh daemon is stared but public key was not yet put into /home/core/.ssh/authorized_keys

kayrus:
resolves #1471


kayrus:
Also added qcow2 backup support. This will allow you to run tests inside create image every time.

jonboulle:
lgtm

PR #1485, opened 2016-03-09
use better filename for functional test log

reneengelhard:


reneengelhard:
ignore the first patch, needs more adaptions

tixxdz:
Does the functional/clean.sh also tries to remove this file ? if so then it should also be updated there too...

reneengelhard:
@tixxdz: now yes :)

PR #1486, opened 2016-03-09
tests: remove legacy code

kayrus:
/cc @jonboulle 
I added this code for ubuntu support and it was not implemented. So let's remove.


tixxdz:
lgtm, thanks!

PR #1488, opened 2016-03-09
functional: allow ssh port number and channel name to be overridden

dongsupark:
With these changes, users are now able to override channel name
as well as the local ssh port number. For example:

```
 $ COREOS_SSH_PORT=2245 COREOS_CHANNEL=alpha ./run-in-qemu
```

Additionally BASE_IMAGE should be set according to the channel name,
to avoid potential file naming conflicts.

This PR depends on https://github.com/coreos/fleet/pull/1476

/cc @kayrus

### DIFF
--- a/functional/run-in-qemu
+++ b/functional/run-in-qemu
@@ -2,6 +2,44 @@
 
 BASENAME=$(basename $0)
 
+USAGE="Usage: $0 [options]
+Options:
+    -c|-channel CHANNEL
+                channel name (stable/beta/alpha)               [default: stable]
+    -p|-ssh-port PORT
+                The port on localhost to map to the VM's sshd. [default: 2244]
+    -v          Make verbose
+    -h          this help message
+
+This script is a wrapper around qemu for starting CoreOS virtual machines,
+especially for running functional tests automatically.
+The -c option can be used to specify a channel name: stable, beta, or alpha.
+The -p option may be used to specify a particular ssh port on localhost to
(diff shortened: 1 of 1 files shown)


tixxdz:
@dongsupark lgtm! could you please rebase and update with a git push -f on endocode:dongsu/qemu-override-env-vars same branch ? thanks

dongsupark:
@tixxdz rebased.
And I also updated the naming scheme of BASE_IMAGE, from ``coreos_production_qemu_image_${COREOS_CHANNEL}.img`` to ``coreos_${COREOS_CHANNEL}_production_qemu_image.img``, as suggested by @kayrus.

tixxdz:
@dongsupark actually after some thoughts on this, I think it would be better to just follow the same thing that's provided here: https://coreos.com/os/docs/latest/booting-with-qemu.html  http://stable.release.core-os.net/amd64-usr/current/coreos_production_qemu.sh  and use arguments for port -p and possibly print help only for "--help" 

It seems that scripts is autogenerated, but we won't add much arguments just the port and a possible channel ?  thoughts ?

dongsupark:
Updated.
As @tixxdz suggested, added support for command-line parameters for channel as well as ssh port number, also added help messages.
As @kayrus suggested, changed the naming scheme of ``$IMAGE`` to reflect correct channel name.

dongsupark:
@tixxdz Updated the help message

PR #1489, opened 2016-03-09
tests: removed yet another legacy code

kayrus:


tixxdz:
@kayrus could you please give it a better commit log, which test ? and try to be precise so we won't use the same words for previous or future patchesn thank you!

tixxdz:
lgtm

PR #1491, opened 2016-03-09
tests: kill ssh-agent on exit

kayrus:
in case when you run several tests inside one VM you'd probably see lots of `ssh-agent`'s
/cc @jonboulle @antrik @tixxdz


jonboulle:
lgtm

PR #1493, opened 2016-03-09
Godeps: move to new uuid package home

jonboulle:
code.google.com/p/go-uuid/uuid has moved to github.com/pborman/uuid


reneengelhard:
I actually did already after I saw your commit but that only solves the uuid problem. The other ones still are there.

If I rm -rf Godeps/_workspace and then try again it doesn't work either. Since it still complains about the various stuff not a git checkout(?): So I tried the big hammer and rm -rf Godeps/_workspace/ because restore should do as it said, right.

That at least got them all. And a go get -d github.com/vishvananda/netlink works.  But a
godep save -r ./... still removes everything from Godeps.json...

reneengelhard:
oops, sorry, wrong PR.

PR #1496, opened 2016-03-10
server: fix panic on graceful shutdown

antrik:
Retry of https://github.com/coreos/fleet/pull/1452 , because Semaphore.


antrik:
Weird: Semaphore did check the latest commit, but it doesn't show up in the status box...

tixxdz:
Lgtm for the last 3 patches. I would perhaps argue that we increase the timeout a little bit, but @antrik if you are sure that all tests did run correctly and it does not have any other side effect, then it's probably ok.

@jonboulle I'm a bit busy with the other replace issue, but if you think that we should take a closer look again, ok. Thank you!

Issue #715, opened 2014-07-27
Add functional test coverage for Agent's loss of etcd connectivity


bcwaldon:
Related https://github.com/coreos/fleet/issues/708

We need to add a functional test that ensures that the fleet agent behaves correctly when it loses access to the etcd cluster.

wuqixuan:
+1

themicster:
+1, So it turns out in order to simulate etcd being down as in a case where the etcd machine is dead you have to ignore packets to or from the IP that etcd should be running on. In normal behavior of an up machine you'll receive an ICMP that tells you that the connection is refused from the machine because it is not listening on that port. This results in the client instantly trying the next etcd server, which is great and there is no problem.
But in the case of the machine being dead there is no ICMP packet sent back and the client will wait for normally a long timeout period before trying the next etcd server. Which is where I figured the HeaderTimeoutPerRequest would come in handy. But I agree a test should be created to simulate this condition of non-existent machine in the etcd list.
So, in order to do that properly I have found that you cannot simply stop etcd on a server. You have to drop the packets to/from that server to simulate the whole machine being offline. Or you can put an invalid IP address in the list of etcd machines, but I would think this would be much more difficult, but maybe not. Also keep in mind there is a new pinning system that will pin good servers, so you will need to make sure that it is not just skipping the server altogether. I believe it retries all the servers every 10 seconds or so and then decides which one to pin.

I'm not sure how your test environment is setup, but in my case I have a cumulus switch I can put an iptables rule in to prevent traffic to/from a server. I think you could also simulate this without loosing complete connectivity to the machine by just dropping all ICMP, or find the specific icmp type that gets sent for "connection refused" and block that. Be sure you understand the actual packet flow here because it's been a while since I've read TCP/IP Illustrated, I could be missing something. But I hope this helps.

antrik:
So I think this needs some clarification on what exactly we want to test here.

First of all, AIUI this is about total loss of connectivity to the entire etcd cluster -- failover handling is an entirely different issue, right? As the functional test currently all work with a single etcd instance, this should be more or less straightforward I guess...

Next question is about what exactly "loss of connectivity" means. Should we test the clear-cut case when we get an immediate error response, e.g. when the etcd process went down? Or the case where we don't get replies (but eventually a timeout I guess) when the server process or the network just hangs? Or both?

[...]

PR #1501, opened 2016-03-11
Add test for behaviour on etcd connectivity loss

antrik:
Fixes https://github.com/coreos/fleet/issues/715

### DIFF
--- a/functional/connectivity-loss_test.go
+++ b/functional/connectivity-loss_test.go
@@ -132,19 +132,11 @@ func TestSingleNodeConnectivityLoss(t *testing.T) {
 	}
 
 	// Wait for initial state being reached.
-	timeout := 15 * time.Second
-	alarm := time.After(timeout)
-	ticker := time.Tick(250 * time.Millisecond)
-loop:
-	for {
-		select {
-		case <-alarm:
-			t.Fatalf("Failed to reach expected initial state within %v.", timeout)
-		case <-ticker:
-			if isExpected, _, _ := checkExpectedStates(); isExpected {
-				break loop
-			}
-		}
+	timeout, err := util.WaitForState(
+		func() bool { isExpected, _, _ := checkExpectedStates(); return isExpected },
+	)
+	if err != nil {
+		t.Fatalf("Failed to reach expected initial state within %v.", timeout)
 	}
 
 	// Cut connection to etcd.
--- a/functional/connectivity-loss_test.go
+++ b/functional/connectivity-loss_test.go
@@ -1,0 +1,229 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package functional
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/coreos/fleet/functional/platform"
+	"github.com/coreos/fleet/functional/util"
+)
+
+// Check that units states do not change on loss of connectivity to etcd.
+//
+// Note: this only tests the behaviour of the disconnected node;
+// but not the reaction of the rest of the cluster,
+// nor reconciliation after connectivity is restored.
+func TestSingleNodeConnectivityLoss(t *testing.T) {
+	cluster, err := platform.NewNspawnCluster("smoke")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cluster.Destroy()
+
+	m0, err := cluster.CreateMember()
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = cluster.WaitForNMachines(m0, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Set up some units.
+	stateMapping := map[string]struct {
+		command          []string
+		runState         string
+		systemdFileState string
+		systemdState     []string
+	}{
+		"inactive": {[]string{"submit"}, "", "", nil},
+		"loaded":   {[]string{"load", "--no-block"}, "inactive", "enabled", nil},
+		"launched": {[]string{"start", "--no-block"}, "active", "enabled", []string{"loaded", "active", "running"}},
+	}
+	createUnits := map[string][]string{}
+	expectedUnitFiles := map[string]string{}
+	expectedUnitStates := map[string]string{}
+	expectedSystemdFiles := map[string]string{}
+	expectedSystemdStates := map[string][]string{}
+	for _, service := range []string{"single", "global"} {
+		for state, mapping := range stateMapping {
+			unitName := fmt.Sprintf("%s@%s.service", service, state)
+			unitPath := fmt.Sprintf("fixtures/units/%s", unitName)
+			createUnits[unitName] = append(mapping.command, unitPath)
+
+			expectedUnitFiles[unitName] = state
+
+			if mapping.runState != "" {
+				expectedUnitStates[unitName] = mapping.runState
+			}
+
+			if mapping.systemdFileState != "" {
+				expectedSystemdFiles[unitName] = mapping.systemdFileState
+			}
+
+			if mapping.systemdState != nil {
+				expectedSystemdStates[unitName] = mapping.systemdState
+			}
+		}
+	}
+	for name, command := range createUnits {
+		stdout, stderr, err := cluster.Fleetctl(m0, command...)
+		if err != nil {
+			t.Fatalf("Failed creating unit %s: %v\nstdout: %s\nstderr:%s", name, err, stdout, stderr)
+		}
+	}
+
+	checkExpectedStates := func() (isExpected bool, expected, actual map[string]string) {
+		// First check unit files.
+		// These shouldn't change at all after intital submit -- but better safe than sorry...
+		stdout, _, err := cluster.Fleetctl(m0, "list-unit-files", "--no-legend", "--full", "--fields", "unit,dstate")
+		if err != nil {
+			t.Errorf("Failed listing unit files: %v", err)
+		}
+		stdout = strings.TrimSpace(stdout)
+
+		lines := strings.Split(stdout, "\n")
+		actualUnitFiles := map[string]string{}
+		if stdout != "" {
+			for _, line := range lines {
+				cols := strings.Fields(line)
+				actualUnitFiles[cols[0]] = cols[1]
+			}
+		}
+
+		if !reflect.DeepEqual(actualUnitFiles, expectedUnitFiles) {
+			return false, expectedUnitFiles, actualUnitFiles
+		}
+
+		// Now check the actual unit states.
+		stdout, _, err = cluster.Fleetctl(m0, "list-units", "--no-legend", "--full", "--fields", "unit,active")
+		if err != nil {
+			t.Errorf("Failed listing units: %v", err)
+		}
+		stdout = strings.TrimSpace(stdout)
+
+		lines = strings.Split(stdout, "\n")
+		actualUnitStates := map[string]string{}
+		if stdout != "" {
+			for _, line := range lines {
+				cols := strings.Fields(line)
+				actualUnitStates[cols[0]] = cols[1]
+			}
+		}
+
+		return reflect.DeepEqual(actualUnitStates, expectedUnitStates), expectedUnitStates, actualUnitStates
+	}
+
+	// Wait for initial state being reached.
+	timeout := 15 * time.Second
+	alarm := time.After(timeout)
+	ticker := time.Tick(250 * time.Millisecond)
+loop:
+	for {
+		select {
+		case <-alarm:
+			t.Fatalf("Failed to reach expected initial state within %v.", timeout)
+		case <-ticker:
+			if isExpected, _, _ := checkExpectedStates(); isExpected {
+				break loop
+			}
+		}
+	}
+
+	// Cut connection to etcd.
+	//
+	// We use REJECT here, so fleet knows immediately that it's disconnected, rather than waiting for a timeout.
+	if _, err = cluster.MemberCommand(m0, "sudo", "iptables", "-I", "OUTPUT", "-p", "tcp", "-m", "multiport", "--dports=2379,4001", "-j", "REJECT"); err != nil {
+		t.Fatal(err)
+	}
+
+	// Wait long enough to be reasonably confident that no more state changes will happen.
+	ttl, _ := time.ParseDuration(util.FleetTTL)
+	agentReconcileInterval := 5 * time.Second
+	slack := 2 * time.Second
+
+	time.Sleep(ttl + agentReconcileInterval + slack)
+
+	// Check unit state after connection loss.
+	//
+	// Note: we cannot use fleetctl to check the state here,
+	// as fleet is not available to give us this information...
+	// We have to go deeper, and try to obtain the information from systemd directly.
+	stdout, err := cluster.MemberCommand(m0, "systemctl", "list-unit-files", "-t", "service", "--no-legend", "single@*.service", "global@*.service")
+	stdout = strings.TrimSpace(stdout)
+	if err != nil {
+		t.Fatalf("Failed to retrieve systemd unit file states: %v", err)
+	}
+	actualSystemdFiles := map[string]string{}
+	if stdout != "" {
+		for _, line := range strings.Split(stdout, "\n") {
+			fields := strings.Fields(line)
+			actualSystemdFiles[fields[0]] = fields[1]
+		}
+	}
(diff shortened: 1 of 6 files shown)


antrik:
Whoops, gofmt...

antrik:
Fixed formatting.

functional/connectivity-loss_test.go
```diff
@@ -0,0 +1,229 @@
...
+		stdout = strings.TrimSpace(stdout)
+		if err != nil {
+			t.Errorf("Failed listing unit files: %v", err)
+		}
+
+		lines := strings.Split(stdout, "\n")
+		actualUnitFiles := map[string]string{}
+		if stdout != "" {
+			for _, line := range lines {
+				cols := strings.Split(line, "\t")
```
kayrus: You can use [Fields](https://golang.org/pkg/strings/#Fields) instead of `Split`.

functional/connectivity-loss_test.go
```diff
@@ -0,0 +1,229 @@
...
+		stdout = strings.TrimSpace(stdout)
+		if err != nil {
+			t.Errorf("Failed listing units: %v", err)
+		}
+
+		lines = strings.Split(stdout, "\n")
+		actualUnitStates := map[string]string{}
+		if stdout != "" {
+			for _, line := range lines {
+				cols := strings.Split(line, "\t")
```
kayrus: [Fields](https://golang.org/pkg/strings/#Fields)

functional/connectivity-loss_test.go
```diff
@@ -0,0 +1,229 @@
...
+	slack := 2 * time.Second
+
+	time.Sleep(ttl + agentReconcileInterval + slack)
+
+	// Check unit state after connection loss.
+	//
+	// Note: we cannot use fleetctl to check the state here,
+	// as fleet is not available to give us this information...
+	// We have to go deeper, and try to obtain the information from systemd directly.
+	stdout, err := cluster.MemberCommand(m0, "sudo", "sh", "-c", `'systemctl list-unit-files -t service --no-legend | ( grep "\(single\|global\)@" || true )'`)
```
kayrus: It is not necessary to use sudo to list unit files. Instead of grep you can use:

```sh
systemctl list-unit-files -t service --no-legend --no-pager single@.service global@.service
```

functional/connectivity-loss_test.go
```diff
@@ -0,0 +1,229 @@
...
+		for _, line := range strings.Split(stdout, "\n") {
+			fields := strings.Fields(line)
+			actualSystemdFiles[fields[0]] = fields[1]
+		}
+	}
+	if !reflect.DeepEqual(actualSystemdFiles, expectedSystemdFiles) {
+		t.Fatalf("Units files not in expected state after loosing connectivity.\nExpected: %v\nActual: %v", expectedSystemdFiles, actualSystemdFiles)
+	}
+
+	stdout, err = cluster.MemberCommand(m0, "sudo", "sh", "-c", `'systemctl list-units -t service --no-legend | ( grep "\(single\|global\)@" || true )'`)
```
kayrus: same note as above

functional/connectivity-loss_test.go
```diff
@@ -0,0 +1,229 @@
...
+	// and everything should be back to normal...)
+	time.Sleep(ttl + agentReconcileInterval + slack)
+
+	// Check state after reconnect.
+	if isExpected, expected, actual := checkExpectedStates(); !isExpected {
+		t.Fatalf("Units not in expected state after restoring connectivity.\nExpected: %v\nActual: %v", expected, actual)
+	}
+
+	// Additionally check the logs of all active units for possible temporary state flapping.
+	stdout, err = cluster.MemberCommand(m0, "sudo", "sh", "-c", `journalctl _PID=1 | ( grep 'Stopping \(single\|global\)@' || true )`)
```
kayrus: You can use [ContainsAny](https://golang.org/pkg/strings/#ContainsAny) filter function here instead of grep. Also it is not necessary to use `sudo`

functional/fixtures/units/single@.service
```diff
@@ -0,0 +1,2 @@
+[Service]
+ExecStart=/bin/bash -c "while true; do echo Hello, World!; sleep 1; done"
```
kayrus: It could worth using [specifiers](https://github.com/coreos/fleet/blob/master/Documentation/unit-files-and-scheduling.md#systemd-specifiers), i.e.

```ini
ExecStart=/bin/bash -c "while true; do echo Hello, World %i!; sleep 1; done"
```

functional/fixtures/units/global@.service
```diff
@@ -0,0 +1,5 @@
+[Service]
```
kayrus: Not sure whether global unit should be a template.

functional/connectivity-loss_test.go
antrik: I adapted this code from [`WaitForNActiveUnits()`](https://github.com/coreos/fleet/blob/master/functional/platform/nspawn.go#L107) and [`ParseUnitStates()`](https://github.com/coreos/fleet/blob/master/functional/util/util.go#L99); and it seems to work -- so I didn't see the need to do gratuitous changes...

Note that I used `strings.Fields()` elsewhere, when it was actually necessary.

kayrus: You'd probably have like to see this fix: https://github.com/coreos/fleet/commit/addb71781edad339f6420d84c8cb3678cd1f7e8f

functional/connectivity-loss_test.go
antrik: Well, strictly speaking sudo is not necessary for `systemctl status` either, right? Still, it was used in existing code: https://github.com/coreos/fleet/blob/master/functional/node_test.go#L80 -- so I didn't bother discriminating either. (Note that I did the same in https://github.com/coreos/fleet/blob/master/functional/shutdown_test.go )

If it bothers you, probably best to submit a separate PR that changes this in all existing places -- though it rather seems like unnecessary code churn to me...

Avoiding grep here indeed makes it simpler. Not sure it's really worthwhile; but I don't mind changing it.

functional/connectivity-loss_test.go
antrik: Yes, I could do the filtering in Go; but it would be more complicated, for no real gain as far as I can tell -- so I decided against it.

functional/fixtures/units/global@.service
antrik: Why not?

functional/fixtures/units/single@.service
antrik: Hm... The test doesn't use the output at all; but I guess it might potentially be helpful for debugging.

functional/connectivity-loss_test.go
antrik: Actually, filtering by the template like that doesn't seem to work; I'd probably have to list every possible unit separately, which would be ugly and/or complicated. Feel free to try finding a way -- I for my part don't feel like spending any more time on this. It's not like there is any actual problem with the current approach...

antrik: OK, works with the right syntax.

functional/connectivity-loss_test.go
antrik: Changed it after all: while `strings.ContainsAny()` is *not* the right function, it turns out not too complicated with `strings.Contains()`; and I guess it does improve consistency, now that the other greps are gone as well...

functional/connectivity-loss_test.go
antrik: Oh... In that case, I guess I should adapt it indeed :-)

antrik:
@kayrus OK, addressed all of your comments :-)

functional/connectivity-loss_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Additionally check the logs of all active units for possible temporary state flapping.
+	stdout, err = cluster.MemberCommand(m0, "journalctl", "_PID=1")
+	if err != nil {
+		t.Fatalf("Failed to retrieve journal: %v", err)
+	}
+	if strings.Contains(stdout, "Stopping single@") || strings.Contains(stdout, "Stopping global@") {
+		t.Fatalf("Units were unexpectedly stopped at some point:\n%s", stdout)
+	}
+}
```
tixxdz: Thank you for the work! just a minor comment, could you factor out the function it's a bit longer, maybe move the declaration and the timer loop in their logic if it's possible ?

Thanks!

antrik:
I factored out the common wait logic: https://github.com/endocode/fleet/commit/16e1a1c7a1cbe2cd38efbbcc00b71a75be18d806

I haven't updated the PR yet though, as I'm not entirely convinced this is actually an improvement...

tixxdz:
@antrik as you say in your commit log this logic can be shared so nice, and for every one else who wants later to look into that code it will be a bit easier "hey this parts just checks for waitForState()..." I'm not interested in that part ? so I will check other parts ?!

But the patch that you have linked is not really nice https://github.com/endocode/fleet/commit/16e1a1c7a1cbe2cd38efbbcc00b71a75be18d806

Why you pass two functions as arguments, one should suffice since the timer is triggered anyway, right ? and if you can't return multiple values, just pass a pointer to timeout var where you set there and check back in the caller ?

antrik:
Picking up one of the options I originally discarded, I changed it now to return the error flag accompanied by the timeout value used; and let the caller construct a custom error message from that. I'm still not convinced that this is indeed more elegant; but it does simplify the invocation -- so I guess we can consider it a win...

functional/fixtures/units/global@.service
kayrus: Not sure whether it is a good idea. When you start 3 units in 3-machines cluster, fleet with schedule 3*3 units. Is it necessary in current test case?

```sh
core@coreos1 ~ $ fleetctl start gtemplate@{1..3}.service
Unit gtemplate@1.service 
Unit gtemplate@2.service 
Unit gtemplate@3.service 
Triggered global unit gtemplate@1.service start
Triggered global unit gtemplate@2.service start
Triggered global unit gtemplate@3.service start
core@coreos1 ~ $ fleetctl list-units
UNIT                    MACHINE                 ACTIVE  SUB
gtemplate@1.service     11594cc1.../coreos2     active  running
gtemplate@1.service     5b93e155.../coreos3     active  running
gtemplate@1.service     f6eaabe9.../coreos1     active  running
gtemplate@2.service     11594cc1.../coreos2     active  running
gtemplate@2.service     5b93e155.../coreos3     active  running
gtemplate@2.service     f6eaabe9.../coreos1     active  running
gtemplate@3.service     11594cc1.../coreos2     active  running
gtemplate@3.service     5b93e155.../coreos3     active  running
gtemplate@3.service     f6eaabe9.../coreos1     active  running
```

antrik: This test explicitly uses a single-machine cluster...

If I didn't use a template, I'd have to use three distinct unit files to cover all cases. I don't see how this would be better?

functional/util/config.go
```diff
@@ -74,6 +75,7 @@ type configValues struct {
 	EtcdEndpoint  string
 	EtcdKeyPrefix string
 	FleetAPIPort  int
+	TTL           string
```
jonboulle: FleetTTL or FleetAgentTTL

functional/util/util.go
```diff
@@ -136,3 +137,20 @@ func TempUnit(contents string) (string, error) {
...
 }
+
+func WaitForState(stateCheckFunc func() bool) (time.Duration, error) {
+	timeout := 15 * time.Second
+	alarm := time.After(timeout)
+	ticker := time.Tick(250 * time.Millisecond)
+	for {
+		select {
+		case <-alarm:
+			// Generic message. Callers can build more specific ones using the supplied timeout value.
```
jonboulle: Don't understand this comment - no option to supply timeout value

functional/connectivity-loss_test.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	actualSystemdStates := map[string][]string{}
+	if stdout != "" {
+		for _, line := range strings.Split(stdout, "\n") {
+			fields := strings.Fields(line)
+			actualSystemdStates[fields[0]] = fields[1:4]
+		}
+	}
+	if !reflect.DeepEqual(actualSystemdStates, expectedSystemdStates) {
+		t.Fatalf("Units not in expected state after loosing connectivity.\nExpected: %v\nActual: %v", expectedSystemdStates, actualSystemdStates)
```
jonboulle: %s/loosing/losing/g

functional/connectivity-loss_test.go
```diff
@@ -0,0 +1,229 @@
...
+			fields := strings.Fields(line)
+			actualSystemdFiles[fields[0]] = fields[1]
+		}
+	}
+	if !reflect.DeepEqual(actualSystemdFiles, expectedSystemdFiles) {
+		t.Fatalf("Units files not in expected state after loosing connectivity.\nExpected: %v\nActual: %v", expectedSystemdFiles, actualSystemdFiles)
+	}
+
+	stdout, err = cluster.MemberCommand(m0, "systemctl", "list-units", "-t", "service", "--no-legend", "single@*.service", "global@*.service")
+	stdout = strings.TrimSpace(stdout)
```
jonboulle: nit, move this below the error check

functional/connectivity-loss_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Again, wait long enough to be reasonably confident that no more state changes will happen.
+	//
+	// Here this should cover the time for fleet to realise connectivity is back,
+	// and for the Agent to complete the second run after reconnection.
+	//
+	// (Unlike for the first run immediately after connectivity is back, by the time of the second run,
+	// Engine leadership and Engine reconciliation should have been sorted out,
+	// and everything should be back to normal...)
+	time.Sleep(ttl + agentReconcileInterval + slack)
```
jonboulle: How much time does this test take overall?

functional/connectivity-loss_test.go
```diff
@@ -0,0 +1,229 @@
...
+		"inactive": {[]string{"submit"}, "", "", nil},
+		"loaded":   {[]string{"load", "--no-block"}, "inactive", "enabled", nil},
+		"launched": {[]string{"start", "--no-block"}, "active", "enabled", []string{"loaded", "active", "running"}},
+	}
+	createUnits := map[string][]string{}
+	expectedUnitFiles := map[string]string{}
+	expectedUnitStates := map[string]string{}
+	expectedSystemdFiles := map[string]string{}
+	expectedSystemdStates := map[string][]string{}
+	for _, service := range []string{"single", "global"} {
```
jonboulle: I'm sort of borderline about having this nested loop and constructing the expected results vs just encoding them as fixtures. I appreciate the flexibility/extensibility of the current approach but it makes it harder to read and reason about the tests and expected results.

jonboulle:
A few comments, but looks good to me overall

functional/util/util.go
antrik: What I meant is that the value is supplied to the *caller* by this function, in the form of a return value... I see though how this wording is confusing. For want of a better idea, I guess I'll just s/supplied/returned/

functional/connectivity-loss_test.go
antrik: I lifted this code from the existing `nspawnClust.WaitForNActiveUnits()`; and I also think I read that doing it like this is considered idiomatic Go code... (Though maybe that was only in some more specific context?)

I certainly do not mind changing it though -- in fact I consider it more natural myself :-)

functional/connectivity-loss_test.go
antrik: The entire test takes about 25s, including two waits of 10s each.

It's a bit longer than I would like, but not terribly so (many other tests take > 10s); and I can't think of a way to speed it up without forsaking reliability...

functional/connectivity-loss_test.go
antrik: This code just kinda evolved like that; I haven't consciously considered this choice before... Thinking about it now, I'm not sure which I prefer myself.

When you say "borderline", you mean you have no strong opinion, i.e. you wouldn't mind leaving it as is?

tixxdz:
@antrik thank you! lgtm for the code as it's now, but still need to fix semaphoreci and see why functional tests fail.

functional/connectivity-loss_test.go
jonboulle: Yep.

antrik:
Rebasing on current master fixed the test issues :-)

kayrus:
This PR should cover this issue: https://github.com/coreos/fleet/pull/1067#issuecomment-188917018

@antrik could you update https://github.com/endocode/fleet/blob/master/functional/shutdown_test.go withing this PR and remove unnecessary `sudo` and `grep` from it?

jonboulle:
LGTM after https://github.com/coreos/fleet/pull/1501#issuecomment-204013818 I guess?

antrik:
@kayrus as I already tried to tell you OOB, these are really separate and pretty much orthogonal issues. The other one tests restart after connectivity loss; this one tests regular behaviour on connectivity loss (without restart).

I don't mind doing a *separate* PR for removing all unnecessary sudos (although personally I'm not convinced it's worth the effort...) -- but that's really unrelated.

(As for the grep, it's *not* unnecessary in the other test: doing the wait loop in Go would be *way* more complicated; and I don't see any good reason to change it.)

kayrus:
@antrik I'm ok with that.

jonboulle:
Any objections to this being merged?

tixxdz:
@jonboulle lgtm, and as @kayrus and @antrik agreed they may open a new PR to remove the unnecessary sudos, but it's not related to this PR.

So please go ahead, thank you!

PR #1502, opened 2016-03-11
fleetctl: fixed broken ssh terminal #1499

### LINKED ISSUE #1499: fleetctl: "askToTrustHost" doesn't work correctly with cluster and empty "known_hosts"
Steps to reproduce:

* spawn 3 coreos nodes cluster
* create a template `hello@.service`:

```ini
[Service]
ExecStart=/bin/bash -c "while true; do echo Hello, World %i!; sleep 1; done"
```
* clear `~/.fleetctl/known_hosts`

* start units `fleetctl start hello@{1..10}.service`

* `fleetctl status hello@{1..10}.service` then type "yes" for each request. second request will not work until you type it twice.

kayrus:
fixes #1499

### DIFF
--- a/ssh/ssh.go
+++ b/ssh/ssh.go
@@ -66,10 +66,6 @@ func makeSession(client *SSHForwardingClient) (session *gossh.Session, finalize 
 		return
 	}
 
-	session.Stdout = os.Stdout
-	session.Stderr = os.Stderr
-	session.Stdin = os.Stdin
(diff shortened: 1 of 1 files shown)


kayrus:
Nope, this fix is not correct. `fleetctl ssh hello@5 "bash"` doesn't work, because stdin is not forwarded.

PR #1503, opened 2016-03-14
fleetctl: print error when units were not found in registry

kayrus:
Destroy logic was already [improved](https://github.com/coreos/fleet/commit/ad702480e3a5a404690396d6c6b5e6c8c5d486f4) so this issue could be closed: https://github.com/coreos/fleet/pull/1256
https://github.com/coreos/fleet/issues/710

But we still have to print some kind of error message when units were not found.

### DIFF
--- a/fleetctl/destroy.go
+++ b/fleetctl/destroy.go
@@ -46,6 +46,11 @@ func runDestroyUnits(args []string) (exit int) {
 		return 1
 	}
 
+	if len(units) == 0 {
+		stderr("Units not found in registry")
+		return 0
+	}
+
 	for _, v := range units {
 		err := cAPI.DestroyUnit(v.Name)
 		if err != nil {
--- a/fleetctl/stop.go
+++ b/fleetctl/stop.go
@@ -63,6 +63,11 @@ func runStopUnit(args []string) (exit int) {
 		return 1
 	}
 
+	if len(units) == 0 {
+		stderr("Units not found in registry")
+		return 0
+	}
+
 	stopping := make([]string, 0)
 	for _, u := range units {
 		if !suToGlobal(u) {
--- a/fleetctl/unload.go
+++ b/fleetctl/unload.go
@@ -47,6 +47,11 @@ func runUnloadUnit(args []string) (exit int) {
 		return 1
 	}
 
+	if len(units) == 0 {
+		stderr("Units not found in registry")
+		return 0
+	}
+
 	wait := make([]string, 0)
 	for _, s := range units {
 		if !suToGlobal(s) {


tixxdz:
Hi @kayrus thank you for the patch! but since we agreed on ignoring non-existent units and don't fail we added unit tests for that case, and the patch just triggers those failures., and also findUnits() filters units based on what you have provided, if you did provided non-existent units, it will return an empty list and not an error. The callers assume an empty list if units do not exist, errors are only in case the Units() fail.

kayrus:
@tixxdz updated

tixxdz:
ok thank you could you also please fix gofmt and push again ;-)

While you are it why is this related to #1256 #710 ?? am I missing something here ? when you call with wildcards! I guess it's has nothing to do with fleetctl command or Go, it's you shell which expands automatically and by accident to local valid unit services..., you should know that you are doing or escape ?!  are these really bugs related to fleet ? hmm ?

kayrus:
Btw, should we filter units when we would like to print only one unit? I.e.

```sh
$ systemctl list-unit-files etcd2.service
UNIT FILE     STATE   
etcd2.service disabled

1 unit files listed.
$ systemctl list-units etcd2.service
UNIT          LOAD   ACTIVE SUB     DESCRIPTION
etcd2.service loaded active running etcd2

LOAD   = Reflects whether the unit definition was properly loaded.
ACTIVE = The high-level unit activation state, i.e. generalization of SUB.
SUB    = The low-level unit activation state, values depend on unit type.

1 loaded units listed. Pass --all to see loaded but inactive units, too.
To show all installed unit files use 'systemctl list-unit-files'.
```

```sh
$ fleetctl list-unit-files hello.service
hello.service           3750b3e launched        launched        706e4d01.../coreos1
hello@1.service         79699e3 launched        launched        706e4d01.../coreos1
hello@2.service         79699e3 launched        launched        706e4d01.../coreos1
hello@3.service         79699e3 launched        launched        706e4d01.../coreos1
$ fleetctl list-units hello.service
UNIT                    MACHINE                 ACTIVE  SUB
hello.service           706e4d01.../coreos1     active  running
hello@1.service         706e4d01.../coreos1     active  running
hello@2.service         706e4d01.../coreos1     active  running
hello@3.service         706e4d01.../coreos1     active  running
```

tixxdz:
@kayrus hmm IMO yes it make sense, not sure what others think

I guess the only thing to worry about is the exit codes, since list-unit-files always succeeded, this should not change. If unit tests are not there then this should also be added.

antrik:
I'm not convinced of this change: it seems a bit arbitrary to me that it warns if no units were matched, but not if some of the supplied units were matched while others were not.

At the very least, the warning should be clearer, e.g. "Warning: No units matched."

I'm not sure I understand the remark about listing one unit. (Isn't that a separate issue?...)

dongsupark:
LGTM.
I think there's no reason not to merge this. I'll merge it tomorrow.

PR #1509, opened 2016-03-16
fleet: add replace unit support

tixxdz:
This PR allows units to be replaced with "submit", "load" and "start" commands. Just add the new "--replace" switch.

The previous discussion was about overwrite in this PR https://github.com/coreos/fleet/pull/1295

This PR tries to fix: https://github.com/coreos/fleet/issues/760

### DIFF
--- a/api/units.go
+++ b/api/units.go
@@ -101,15 +101,35 @@ func (ur *unitsResource) set(rw http.ResponseWriter, req *http.Request, item str
 		return
 	}
 
+	newUnit := false
 	if eu == nil {
 		if len(su.Options) == 0 {
 			err := errors.New("unit does not exist and options field empty")
 			sendError(rw, http.StatusConflict, err)
+			return
 		} else if err := ValidateOptions(su.Options); err != nil {
 			sendError(rw, http.StatusBadRequest, err)
+			return
 		} else {
-			ur.create(rw, su.Name, &su)
+			// New valid unit
+			newUnit = true
 		}
+	} else if eu.Name == su.Name && len(su.Options) > 0 {
+		// There is already a unit with the same name that
+		// was submitted before. Check their hashes, if they do
+		// not match then this is probably a new version which
+		// needs its own new unit entry.
+		// In the other case if su.Options == 0 then probably we
+		// don't want to update the Unit options nor its content
+		// but only set the target job state of the
+		// corresponding unit, in this case just ignore.
+		a := schema.MapSchemaUnitOptionsToUnitFile(su.Options)
+		b := schema.MapSchemaUnitOptionsToUnitFile(eu.Options)
+		newUnit = !unit.MatchUnitFiles(a, b)
+	}
+
+	if newUnit {
+		ur.create(rw, su.Name, &su)
 		return
 	}
 
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -671,6 +671,88 @@ func createUnit(name string, uf *unit.UnitFile) (*schema.Unit, error) {
 	return &u, nil
 }
 
+// checkReplaceUnitState checks if the unit should be replaced.
+// It takes a Unit object as a parameter.
+// It returns 0 on success and if the unit should be replaced, 1 if the
+// unit should not be replaced; and any error encountered.
+func checkReplaceUnitState(unit *schema.Unit) (int, error) {
+	// We replace units only for 'submit', 'load' and
+	// 'start' commands.
+	allowedReplace := map[string][]job.JobState{
+		"submit": []job.JobState{
+			job.JobStateInactive,
+		},
+		"load": []job.JobState{
+			job.JobStateInactive,
+			job.JobStateLoaded,
+		},
+		"start": []job.JobState{
+			job.JobStateInactive,
+			job.JobStateLoaded,
+			job.JobStateLaunched,
+		},
+	}
+
+	if allowedJobs, ok := allowedReplace[currentCommand]; ok {
+		for _, j := range allowedJobs {
+			if job.JobState(unit.DesiredState) == j {
+				return 0, nil
+			}
+		}
+		// Report back to caller that we are not allowed to
+		// cross unit transition states
+		stderr("Warning: can not replace Unit(%s) in state '%s', use the appropriate command", unit.Name, unit.DesiredState)
+	} else {
+		// This function should only be called from 'submit',
+		// 'load' and 'start' upper paths.
+		return 1, fmt.Errorf("error: replacing units is not supported in this context")
+	}
+
+	return 1, nil
+}
+
+// checkUnitCreation checks if the unit should be created.
+// It takes a unit file path as a parameter.
+// It returns 0 on success and if the unit should be created, 1 if the
+// unit should not be created; and any error encountered.
+func checkUnitCreation(arg string) (int, error) {
+	name := unitNameMangle(arg)
+
+	// First, check if there already exists a Unit by the given name in the Registry
+	unit, err := cAPI.Unit(name)
+	if err != nil {
+		return 1, fmt.Errorf("error retrieving Unit(%s) from Registry: %v", name, err)
+	}
+
+	// check if the unit is running
+	if unit == nil {
+		if sharedFlags.Replace {
+			log.Debugf("Unit(%s) was not found in Registry", name)
+		}
+		// Create a new unit
+		return 0, nil
+	}
+
+	// if sharedFlags.Replace is not set then we warn in case
+	// the units differ
+	different, err := isLocalUnitDifferent(arg, unit, !sharedFlags.Replace, false)
+
+	// if sharedFlags.Replace is set then we fail for errors
+	if sharedFlags.Replace {
+		if err != nil {
+			return 1, err
+		} else if different {
+			return checkReplaceUnitState(unit)
+		} else {
+			stdout("Found same Unit(%s) in Registry, nothing to do", unit.Name)
+		}
+	} else if different == false {
+		log.Debugf("Found same Unit(%s) in Registry, no need to recreate it", name)
+	}
+
+	return 1, nil
+}
+
 // lazyCreateUnits iterates over a set of unit names and, for each, attempts to
 // ensure that a unit by that name exists in the Registry, by checking a number
 // of conditions and acting on the first one that succeeds, in order of:
@@ -688,14 +770,10 @@ func lazyCreateUnits(args []string) error {
 		arg = maybeAppendDefaultUnitType(arg)
 		name := unitNameMangle(arg)
 
-		// First, check if there already exists a Unit by the given name in the Registry
-		u, err := cAPI.Unit(name)
+		ret, err := checkUnitCreation(arg)
 		if err != nil {
-			return fmt.Errorf("error retrieving Unit(%s) from Registry: %v", name, err)
-		}
-		if u != nil {
-			log.Debugf("Found Unit(%s) in Registry, no need to recreate it", name)
-			warnOnDifferentLocalUnit(arg, u)
+			return err
+		} else if ret != 0 {
 			continue
 		}
 
@@ -732,6 +810,79 @@ func lazyCreateUnits(args []string) error {
 	}
 
 	return nil
+}
+
+// matchUnitFiles compares two unitFiles
+// Returns true if the units match, false otherwise.
+func matchUnitFiles(a *unit.UnitFile, b *unit.UnitFile) bool {
+	if a.Hash() == b.Hash() {
+		return true
+	}
+
+	return false
+}
+
+// matchLocalFileAndUnit compares a file with a Unit
+// Returns true if the contents of the file matches the unit one, false
+// otherwise; and any error encountered.
+func matchLocalFileAndUnit(file string, su *schema.Unit) (bool, error) {
+	result := false
+	a := schema.MapSchemaUnitOptionsToUnitFile(su.Options)
+
+	_, err := os.Stat(file)
+	if err == nil {
+		b, err := getUnitFromFile(file)
+		if err == nil {
+			result = matchUnitFiles(a, b)
+		}
+	}
+
+	return result, err
+}
+
+// isLocalUnitDifferent compares a Unit on the file system with a one
+// provided from the Registry.
+// isLocalUnitDifferent first tries to load the passed Unit from the
+// local file system and compares it with the Unit that is in the
+// Registry. If it fails to load that Unit from the filesystem and
+// fatal was not set, it will check again if that file name is an
+// instance of a template, if so it will load the template Unit and
+// compare it with the provided Unit.
+// It takes four arguments; a path to the local Unit on the file system,
+// the Unit in the registry, a boolean to warn in case the Units differ;
+// and a last boolean to fail in case fatal errors happen.
+// Returns true if the local Unit on file system is different from the
+// one provided, false otherwise; and any error encountered.
+func isLocalUnitDifferent(file string, su *schema.Unit, warnIfDifferent bool, fatal bool) (bool, error) {
+	result, err := matchLocalFileAndUnit(file, su)
+	if err == nil {
+		// Warn in case unit differs from local file
+		if result == false && warnIfDifferent {
+			stderr("WARNING: Unit %s in registry differs from local unit file %s", su.Name, file)
+		}
+		return !result, nil
+	} else if fatal {
+		return false, err
+	}
+
+	info := unit.NewUnitNameInfo(path.Base(file))
+	if info == nil {
+		return false, fmt.Errorf("error extracting information from unit name %s", file)
+	} else if !info.IsInstance() {
+		return false, fmt.Errorf("error Unit %s does not seem to be a template unit", file)
+	}
+
+	templFile := path.Join(path.Dir(file), info.Template)
+	result, err = matchLocalFileAndUnit(templFile, su)
+	if err == nil {
+		// Warn in case unit differs from local template unit file
+		if result == false && warnIfDifferent {
+			stderr("WARNING: Unit %s in registry differs from local template unit file %s", su.Name, info.Template)
+		}
+		return !result, nil
+	}
+
+	return false, err
 }
 
 func warnOnDifferentLocalUnit(loc string, su *schema.Unit) {
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -563,7 +563,7 @@ func getUnitFileFromTemplate(uni *unit.UnitNameInfo, fileName string) (*unit.Uni
 	}
 
 	if tmpl != nil {
-		warnOnDifferentLocalUnit(fileName, tmpl)
+		isLocalUnitDifferent(fileName, tmpl, true, false)
 		uf = schema.MapSchemaUnitOptionsToUnitFile(tmpl.Options)
 		log.Debugf("Template Unit(%s) found in registry", uni.Template)
 	} else {
@@ -883,26 +883,6 @@ func isLocalUnitDifferent(file string, su *schema.Unit, warnIfDifferent bool, fa
 	}
 
 	return false, err
-}
-
-func warnOnDifferentLocalUnit(loc string, su *schema.Unit) {
-	suf := schema.MapSchemaUnitOptionsToUnitFile(su.Options)
-	if _, err := os.Stat(loc); !os.IsNotExist(err) {
-		luf, err := getUnitFromFile(loc)
-		if err == nil && luf.Hash() != suf.Hash() {
-			stderr("WARNING: Unit %s in registry differs from local unit file %s", su.Name, loc)
-			return
-		}
-	}
-	if uni := unit.NewUnitNameInfo(path.Base(loc)); uni != nil && uni.IsInstance() {
-		file := path.Join(path.Dir(loc), uni.Template)
-		if _, err := os.Stat(file); !os.IsNotExist(err) {
-			tmpl, err := getUnitFromFile(file)
-			if err == nil && tmpl.Hash() != suf.Hash() {
-				stderr("WARNING: Unit %s in registry differs from local template unit file %s", su.Name, uni.Template)
-			}
-		}
-	}
 }
 
 func lazyLoadUnits(args []string) ([]*schema.Unit, error) {
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -103,10 +103,14 @@ var (
 		Full          bool
 		NoLegend      bool
 		NoBlock       bool
+		Replace       bool
 		BlockAttempts int
 		Fields        string
 		SSHPort       int
 	}{}
+
+	// current command being executed
+	currentCommand string
 
 	// used to cache MachineStates
 	machineStates map[string]*machine.MachineState
@@ -286,6 +290,10 @@ func main() {
 			os.Exit(1)
 		}
 	}
+
+	// We use this to know in which context we are,
+	// submit, load or start
+	currentCommand = cmd.Name
 
 	os.Exit(cmd.Run(cmd.Flags.Args()))
 
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -812,16 +812,6 @@ func lazyCreateUnits(args []string) error {
 	return nil
 }
 
-// matchUnitFiles compares two unitFiles
-// Returns true if the units match, false otherwise.
-func matchUnitFiles(a *unit.UnitFile, b *unit.UnitFile) bool {
-	if a.Hash() == b.Hash() {
-		return true
-	}
-
-	return false
-}
-
 // matchLocalFileAndUnit compares a file with a Unit
 // Returns true if the contents of the file matches the unit one, false
 // otherwise; and any error encountered.
@@ -833,7 +823,7 @@ func matchLocalFileAndUnit(file string, su *schema.Unit) (bool, error) {
 	if err == nil {
 		b, err := getUnitFromFile(file)
 		if err == nil {
-			result = matchUnitFiles(a, b)
+			result = unit.MatchUnitFiles(a, b)
 		}
 	}
 
--- a/functional/unit_action_test.go
+++ b/functional/unit_action_test.go
@@ -31,6 +31,8 @@ import (
 const (
 	tmpHelloService = "/tmp/hello.service"
 	fxtHelloService = "fixtures/units/hello.service"
+	tmpFixtures     = "/tmp/fixtures"
+	numUnitsReplace = 9
 )
 
 // TestUnitRunnable is the simplest test possible, deplying a single-node
@@ -140,6 +142,10 @@ func TestUnitSubmitReplace(t *testing.T) {
 	if err := replaceUnitCommon("submit"); err != nil {
 		t.Fatal(err)
 	}
+
+	if err := replaceUnitMultiple("submit", numUnitsReplace); err != nil {
+		t.Fatal(err)
+	}
 }
 
 // TestUnitLoadReplace() tests whether a command "fleetctl load --replace
@@ -148,12 +154,20 @@ func TestUnitLoadReplace(t *testing.T) {
 	if err := replaceUnitCommon("load"); err != nil {
 		t.Fatal(err)
 	}
+
+	if err := replaceUnitMultiple("load", numUnitsReplace); err != nil {
+		t.Fatal(err)
+	}
 }
 
 // TestUnitStartReplace() tests whether a command "fleetctl start --replace
 // hello.service" works or not.
 func TestUnitStartReplace(t *testing.T) {
 	if err := replaceUnitCommon("start"); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := replaceUnitMultiple("start", numUnitsReplace); err != nil {
 		t.Fatal(err)
 	}
 }
@@ -574,6 +588,132 @@ func replaceUnitCommon(cmd string) error {
 	if err := WaitForNUnitsCmd(cmd, 0); err != nil {
 		return fmt.Errorf("Failed to get every unit to be cleaned up: %v", err)
 	}
+
+	return nil
+}
+
+// replaceUnitMultiple() tests whether a command "fleetctl {submit,load,start}
+// --replace hello.service" works or not.
+func replaceUnitMultiple(cmd string, n int) error {
+	// check if cmd is one of the supported commands.
+	listCmds := []string{"submit", "load", "start"}
+	found := false
+	for _, ccmd := range listCmds {
+		if ccmd == cmd {
+			found = true
+		}
+	}
+	if !found {
+		return fmt.Errorf("invalid command %s", cmd)
+	}
+
+	cluster, err := platform.NewNspawnCluster("smoke")
+	if err != nil {
+		return fmt.Errorf("%v", err)
+	}
+	defer cluster.Destroy()
+
+	m, err := cluster.CreateMember()
+	if err != nil {
+		return fmt.Errorf("%v", err)
+	}
+	_, err = cluster.WaitForNMachines(m, 1)
+	if err != nil {
+		return fmt.Errorf("%v", err)
+	}
+
+	WaitForNUnitsCmd := func(cmd string, expectedUnits int) (err error) {
+		if cmd == "submit" {
+			_, err = cluster.WaitForNUnitFiles(m, expectedUnits)
+		} else {
+			_, err = cluster.WaitForNUnits(m, expectedUnits)
+		}
+		return err
+	}
+
+	if _, err := os.Stat(tmpFixtures); os.IsNotExist(err) {
+		os.Mkdir(tmpFixtures, 0755)
+	}
+
+	var stdout string
+	var bodiesOrig []string
+	for i := 1; i <= n; i++ {
+		curHelloService := fmt.Sprintf("/tmp/hello%d.service", i)
+		tmpHelloFixture := fmt.Sprintf("/tmp/fixtures/hello%d.service", i)
+
+		// generate a new service derived by fixtures, and store it under /tmp
+		err = util.CopyFile(tmpHelloFixture, fxtHelloService)
+		if err != nil {
+			return fmt.Errorf("Failed to copy a temp fleet service: %v", err)
+		}
+
+		// run a command for a unit and assert it shows up
+		if _, _, err := cluster.Fleetctl(m, cmd, tmpHelloFixture); err != nil {
+			return fmt.Errorf("Unable to %s fleet unit: %v", cmd, err)
+		}
+		if err := WaitForNUnitsCmd(cmd, i); err != nil {
+			return fmt.Errorf("Did not find %d units in cluster: \n%s", i, stdout)
+		}
+
+		helloFilename := path.Base(curHelloService)
+
+		// retrieve content of hello.service, and append to bodiesOrig[]
+		bodyCur, _, err := cluster.Fleetctl(m, "cat", helloFilename)
+		if err != nil {
+			return fmt.Errorf("Failed to run cat %s: %v", helloFilename, err)
+		}
+
+		bodiesOrig = append(bodiesOrig, bodyCur)
+
+		// generate a new service derived by fixtures, and store it under /tmp
+		err = util.GenNewFleetService(curHelloService, fxtHelloService, "sleep 2", "sleep 1")
+		if err != nil {
+			return fmt.Errorf("Failed to generate a temp fleet service: %v", err)
+		}
+	}
+
+	for i := 1; i <= n; i++ {
+		curHelloService := fmt.Sprintf("/tmp/hello%d.service", i)
+
+		// replace the unit and assert it shows up
+		if _, _, err = cluster.Fleetctl(m, cmd, "--replace", curHelloService); err != nil {
+			return fmt.Errorf("Unable to replace fleet unit: %v", err)
+		}
+		if err := WaitForNUnitsCmd(cmd, n); err != nil {
+			return fmt.Errorf("Did not find %d units in cluster: \n%s", n, stdout)
+		}
+
+		helloFilename := path.Base(curHelloService)
+
+		// retrieve content of hello.service, and compare it with the
+		// correspondent entry in bodiesOrig[]
+		bodyCur, _, err := cluster.Fleetctl(m, "cat", helloFilename)
+		if err != nil {
+			return fmt.Errorf("Failed to run cat %s: %v", helloFilename, err)
+		}
+
+		if bodiesOrig[i-1] == bodyCur {
+			return fmt.Errorf("Error. the unit %s has not been replaced.", helloFilename)
+		}
+	}
+
+	// clean up temp services under /tmp
+	for i := 1; i <= n; i++ {
+		curHelloService := fmt.Sprintf("/tmp/hello%d.service", i)
+
+		if _, _, err := cluster.Fleetctl(m, "destroy", curHelloService); err != nil {
+			fmt.Printf("Failed to destroy unit: %v", err)
+			continue
+		}
+
+		os.Remove(curHelloService)
+	}
+
+	if err := WaitForNUnitsCmd(cmd, 0); err != nil {
+		return fmt.Errorf("Failed to get every unit to be cleaned up: %v", err)
+	}
+
+	os.Remove(tmpFixtures)
 
 	return nil
 }
--- a/functional/unit_action_test.go
+++ b/functional/unit_action_test.go
@@ -717,3 +717,113 @@ func replaceUnitMultiple(cmd string, n int) error {
 
 	return nil
 }
+
+// TestReplaceSerialization tests if the ExecStartPre of the new version
+// of the unit when it replaces the old one is excuted after
+// ExecStopPost of the old version.
+// This test is to make sure that two versions of the same unit will not
+// conflict with each other, that the directives are always serialized,
+// and it tries its best to avoid the following scenarios:
+// https://github.com/coreos/fleet/issues/1000
+// https://github.com/systemd/systemd/issues/518
+// Now we can't guarantee that that behaviour will not be triggered by
+// another external operation, but at least from the Unit replace
+// feature context we try to avoid it.
+func TestReplaceSerialization(t *testing.T) {
+	cluster, err := platform.NewNspawnCluster("smoke")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cluster.Destroy()
+
+	m, err := cluster.CreateMember()
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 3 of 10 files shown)


tixxdz:
We need further testing for this one. thanks!

antrik:
Regarding the test issue: I don't know whether this is a bug or expected behaviour; but there seems to be no guarantee that the units have reached the final states by the time the command returns. That's why all the existing tests use [`WaitForNActiveUnits()`](https://github.com/coreos/fleet/blob/master/functional/platform/nspawn.go#L107); and that's also why I created a similar loop in my new connectivity loss test: https://github.com/endocode/fleet/blob/antrik/tests-loss_of_connectivity/functional/connectivity-loss_test.go#L134 (I couldn't use `WaitForNActiveUnits()` in this case, as not all of the units are supposed to become active -- which I think also applies here, depending on the case?) I guess you need something similar.

antrik:
According to the commit message of the first commit (and from a quick glance at the code), a command with `--replace` will only take effect if the unit is already in the respective state. As discussed earlier (out-of-band), I'm pretty sure this is *not* the desired behaviour. The commands should behave exactly the same as they would without `--replace`, except that *if* a matching unit already exists, it is replaced by the new version, rather than retaining the old one. So if a unit is `inactive` for example, `start --replace` should submit the new unit file (replacing the previous one), load it, and start it.

Slightly less clear is what to do for example if a unit is already launched, and we do `submit --replace`. In my understanding reconciliation will automatically reload and restart the updated unit as soon as it's submitted into the registry? That might be slightly unintuitive -- but I guess there is not much we can do about that...

antrik:
Without actually digging into the code, I see some formal issues with this PR from looking through the commit messages. Some of the commits are just fixing things introduced in earlier commits of the same PR (such as the gofmt fix) -- these should be squashed into the respective commit(s) that introduced the problematic code in the first place.

Also, it looks like the order of the commits is wrong: the commit introducing the `--replace` switch for example can't possibly work without some of the internal changes introduced by the other commits -- so it should be committed *after* these changes, not before...

tixxdz:
@antrik yes actually only one commit that gofmt which was fixed, thanks! ok could you please dig then in the code and ping us if we missed something, this is new functionality. On the note on how to replace units, Yes start --replace should overwrite submit and load but not the other way. Later fleet should be enforced to follow up the life cycle and states of units...

@jonboulle we have added functional tests to replace many units, etc, the functionality is there... but to be honest the thing that bothers me is the order of systemd directives... but hey it depends also on the units. Since fleet is declarative, putting StopPost directives that are buggy or do not finish and updating is asking for trouble...

klausenbusk:
> Slightly less clear is what to do for example if a unit is already launched, and we do submit --replace. In my understanding reconciliation will automatically reload and restart the updated unit as soon as it's submitted into the registry?

I think we should provide 2 things.

1. A way to replace a unit and restart it.
2. and a way to replace a unit a reload it (systemctl daemon-reload)

So I think we could use `submit/load` for the latter and `start` for the first, is that a solution? or what do you think.

dongsupark:
@antrik
> Regarding the test issue: I don't know whether this is a bug or expected behaviour; but there seems to be no guarantee that the units have reached the final states by the time the command returns. That's why all the existing tests use WaitForNActiveUnits()

Makes sense. So I added such checks to the functional tests. Maybe during the next update, this patch will be included:
https://github.com/endocode/fleet/pull/21
https://github.com/endocode/fleet/commit/4e4cc9410215458d8df4b2c092680ad3afc67f6a

For now this test seems to work. However, I'm a little careful about adding more tests included in this PR. I'd rather create a new PR afterwards for further comprehensive tests.

tixxdz:
@antrik 
> Slightly less clear is what to do for example if a unit is already launched, and we do submit --replace. In my understanding reconciliation will automatically reload and restart the updated unit as soon as it's submitted into the registry? That might be slightly unintuitive -- but I guess there is not much we can do about that...

submit is to submit units into the cluster, so the desired state is submit and incative:
https://coreos.com/fleet/docs/latest/states.html

Now if you add "--replace" into the game, and if I quote you "will automatically reload and restart" here you are mixing submit, load and start desired states. If you want to achieve that then just do "start --replace" why "--submit --replace" ? we should not mix desired states, and we should follow the life cycle of units and services, otherwise we introduce confusion, not to mention that these operations are also handled by systemd...

So here just use the appropriate command.

tixxdz:
@klausenbusk 

> I think we should provide 2 things.

> A way to replace a unit and restart it.
and a way to replace a unit a reload it (systemctl daemon-reload)
So I think we could use submit/load for the latter and start for the first, is that a solution? or what do you think.

The way to replace a unit and restart it, is what this PR does! we re-execute systemd directives, could you please try it ? edit your unit, then fleetctl start --replace ?

And the second way you mention is also handled and the manager instructs systemd to reload. Now for submit it does not at all involve systemd reload!

Thank you!

antrik:
@klausenbusk AIUI fleet doesn't presently have any notion of reloading a unit? While IIRC there have been requests for that, I don't think we should try to address this question here. This PR is quite complicated as it is...

klausenbusk:
> And the second way you mention is also handled and the manager instructs systemd to reload. Now for submit it does not at all involve systemd reload!

I want a way I can replace a unit without restarting it! But as @antrik said, that should be another pr.
Use-case: I want to replace my Mariadb Galera unit file. Currently I need to do:

1. fleectl destroy mariadb-galera
2. fleetctl load mariadb-galera
3. Bootstrap mariadb-galera manually on one of the nodes (docker run...).
4. Start mariadb with systemctl start mariadb-galera on node 2
5. Wait for it to finish SST and do the same on node 3.
6. Stop the manually started mariadb-galera
7. fleetctl start mariadb-galera

It a rather long process, and it take some time. If I could replace the unit, I could just ssh the every db server and do sudo systemctl restart mariadb-galera (without downtime).

klausenbusk:
> And the second way you mention is also handled and the manager instructs systemd to reload. Now for submit it does not at all involve systemd reload!

I want a way I can replace a unit without restarting it! But as @antrik said, that should be another pr.
Use-case: I want to replace my Mariadb Galera unit file. Currently I need to do:

1. fleectl destroy mariadb-galera
2. fleetctl load mariadb-galera
3. Bootstrap mariadb-galera manually on one of the nodes (docker run...).
4. Start mariadb with systemctl start mariadb-galera on node 2
5. Wait for it to finish SST and do the same on node 3.
6. Stop the manually started mariadb-galera
7. fleetctl start mariadb-galera

It a rather long process, and it take some time. If I could replace the unit, I could just ssh the every db server and do sudo systemctl restart mariadb-galera (without downtime).

tixxdz:
@klausenbusk ok, could you also explain in more details what are 3), 4), 5) and 6) ? and the content of your units ?

It seems you have the case of template units here? actually currently with this PR you could "submit --replace" a template unit, and already started units from the previous template will not be restarted. To do so you have to add an extra "start --replace template@{x..y}", so the question I see here: between this time could you think of a solution for your use case ? 

"fleetctl start --replace" triggers systemd daemon-reload and all the same processes as "systemctl restart" after that.

And a note so we don't forget, we have these transitions:
`inactive   ->    loaded        ->             launched`
`submit    ->  {load|unload} ->  {start | stop}`
https://coreos.com/fleet/docs/latest/states.html

So unless I'm missing something here, if your units are in the launched state and you want to put them in the loaded state it will always involve a stop (Stop and StopPost directives) then daemon-reload, hence a downtime. Perhaps there is a solution for your use case if you could tell us more ?

klausenbusk:
> @klausenbusk ok, could you also explain in more details what are 3), 4), 5) and 6) ? and the content of your units ?

Note: I have 3 db nodes.
3: Galera needs a cluster to connect to, but as I have shutdown the whole cluster, I need to start 1 instance of Galera with `--wsrep_cluster_address=gcomm://` (create new cluster) which I do just with `docker run ..`
4: I start mariadb-galera on node 2 with `sudo systemctl start mariadb-galera`.
5: I wait for it to sync up with Galera, and do the same on node 3.
6: I stop `docker run ..` on node 1.
7: I do `fleetctl start mariadb-galera` so mariadb-galera is automated started if a machine is rebooted.


> It seems you have the case of template units here?

I use a global unit currently, I see a template unit as a stupid workaround.

kayrus:
@klausenbusk why do you use `systemctl` to start units?

Using templates you can ask fleet to start unit on corresponding machine, i.e.

```sh
fleetctl start mariadb-galera-initial@node1
fleetctl start mariadb-galera@node{2..3}
fleetctl stop mariadb-galera-initial@node2
fleetctl start mariadb-galera@node1
```

These steps could be automated with [sidekicks](https://coreos.com/fleet/docs/latest/examples/service-discovery.html)

antrik:
@klausenbusk it seems what you are essentially asking for is a way to temporarily inhibit part of fleet's reconciliation logic, so it will submit (and load) the new unit, but ignore the fact that the existing unit doesn't match the new requested state...

While I can see the use case for this, I have no idea right now whether this could be done without breaking some fundamental assumptions...

klausenbusk:
> Using templates you can ask fleet to start unit on corresponding machine, i.e.

I don't see any big benefit of using template. I only need to ssh to 1 node instant of 3, that the only benefit, and I can't add a new db node without doing `fleetctl start mariadb-galera@<some-new-number>`

The problem is that I need to stop the whole cluster to change the unit file. What I ask for, is a `--replace` which only replace the unint on disk and do `sudo systemctl daemon-reload` without restarting the unit.
Then I can ssh to all my db nodes, and restart mariadb-galera one-by-one, without stopping the cluster.

kayrus:
@klausenbusk what is the problem of using current implementation `fleetctl start --replace mariadb-galera@node1`? It will update only one unit on node1 and restart only one unit which is scheduled on node1.

Then you can run `fleetctl start --replace mariadb-galera@node2` and `fleetctl start --replace mariadb-galera@node3`.

This solution doesn't require to stop the cluster.

antrik:
@dongsupark if you have test cases in mind further testing the functionality introduced in this PR, they clearly belong here, and not into some followup.

klausenbusk:
> @klausenbusk what is the problem of using current implementation fleetctl start --replace mariadb-galera@node1?

Oh, I didn't through it worked that way. :)

I still prefer a global unit, but template units could be a solution until then.
A flag which restart a global unit node-by-node could be useful.

antrik:
> A flag which restart a global unit node-by-node could be useful.

@klausenbusk yeah, that has been requested before. However, as I tried
to explain, this doens't really jibe well with fleet's automatic
reconciliation model. Also, AIUI it is considered somewhat out of scope
for fleet...

fleetctl/fleetctl.go
```diff
@@ -78,6 +78,7 @@ var (
 		Debug   bool
 		Version bool
 		Help    bool
+		currentCommand	string
```
antrik: This is not really a flag -- and I don't think it's a good idea to treat it as one. While it might look slightly easier than explicitly passing through the command to the functions which need to discriminate on it, it's also harder to follow the code. Such things are generally frowned upon -- for good reason.

fleetctl/fleetctl.go
```diff
@@ -75,10 +75,10 @@ var (
...
 	// flags used by all commands
 	globalFlags = struct {
-		Debug   bool
-		Version bool
-		Help    bool
-		currentCommand	string
+		Debug          bool
+		Version        bool
+		Help           bool
+		currentCommand string
```
antrik: Don't include unrelated whitespace changes in code you do not touch in this patch -- this should go into the previous commit.

Except that in this case I don't actually see any point in having these as separate commits at all... And in fact that's true for almost(?) all commits in this PR: they don't really make any sense in isolation -- so they should just be squashed into a single one.

Separate commits only make sense if you have distinct bits of functionality, or if you have functionality-preserving changes to other parts of code that your main changes depend upon. I don't see that here: all these changes are inherently connected. (In fact I'm pretty sure some of the intermediate commits here result in nonsensical behaviour, and probably wouldn't pass the test suite...)

fleetctl/fleetctl.go
```diff
@@ -669,6 +669,81 @@ func createUnit(name string, uf *unit.UnitFile) (*schema.Unit, error) {
 	return &u, nil
 }
 
+// checkReplaceUnitState checks if the unit should be replaced
+// It takes a Unit object as a parameter
+// It returns 0 on success and if the unit should be replaced, 1 if the
+// unit should not be replaced; and any error acountered
+func checkReplaceUnitState(unit *schema.Unit) (int, error) {
```
antrik: Please follow proper orthography: sentences always begin with a capital letter, and end in a full stop.

Note that this is particularly important for comments preceding functions (and other items), as godoc automatically treats them as doc comments, and tries to extract the first sentence for special handling.

fleetctl/fleetctl.go
```diff
@@ -669,6 +669,81 @@ func createUnit(name string, uf *unit.UnitFile) (*schema.Unit, error) {
...
+
+	if allowedJobs, ok := allowedReplace[globalFlags.currentCommand]; ok {
+		for _, j := range allowedJobs {
+			if job.JobState(unit.DesiredState) == j {
+				return 0, nil
+			}
+		}
+		stderr("Warning: can not replace Unit(%s) in state '%s', use the appropriate command", unit.Name, unit.DesiredState)
+	} else {
+		return 1, fmt.Errorf("error 'replace' is not supported for this command")
```
antrik: This is certainly not the right place to check for such obvious invocation errors... It belongs with the rest of the command line handling.

fleetctl/fleetctl.go
antrik: Why does this function (and also `checkUnitCreation()`) return an int, when this is really just a boolean value?...

fleetctl/fleetctl.go
tixxdz: That one is never triggered, since "--replace" is never passed to other commands, but it was put there in case the code changes later, and someone does is by mystake, will just update the message.

fleetctl/fleetctl.go
```diff
@@ -732,6 +803,64 @@ func lazyCreateUnits(args []string) error {
...
 
+// matchUnitFiles compares two unitFiles
+// Returns true if the units match, false otherwise.
+func matchUnitFiles(a *unit.UnitFile, b *unit.UnitFile) bool {
+	if a.Hash() == b.Hash() {
+		return true
+	}
+
+	return false
+}
```
antrik: Putting this in a function makes no sense: the invocation isn't any simpler than the comparison itself; nor does it make the code easier to read I'd say.

It might still be justified if this were encapsulating some internal detail that is very likely to change -- but I don't see that here...

antrik: Also, this kind of change qualifies as refactoring -- so this would be the one bit that should actually go into a separate commit...

tixxdz: Ok that one was introduced in that commit to make it clear what we are doing, then that code was moved into another unit file so it's shared and used by fleetd too.

fleetctl/fleetctl.go
```diff
@@ -732,6 +803,64 @@ func lazyCreateUnits(args []string) error {
...
+		b, err := getUnitFromFile(file)
+		if err == nil {
+			result = matchUnitFiles(a, b)
+		}
+	}
+
+	return result, err
+}
+
+func isLocalUnitDifferent(file string, su *schema.Unit, warnIfDifferent bool, fatal bool) (bool, error) {
```
antrik: This might actually be the least obvious function of the bunch -- yet it has no comment at all?...

TBH, this code seems badly factored to me: between this function and it's caller, they seem to be doing closely related conditional handling, with no obvious separation of concerns. Maybe it's just me -- but I find this code very hard to follow, and nearly impossible to figure out whether the logic is actually sound...

fleetctl/fleetctl.go
```diff
@@ -732,6 +803,64 @@ func lazyCreateUnits(args []string) error {
...
+// matchLocalFileAndUnit compares a file with a Unit
+// Returns true if the contents of the file matches the unit one, false
+// otherwise; and any error ocountered
+func matchLocalFileAndUnit(file string, su *schema.Unit) (bool, error) {
+	result := false
+	a := schema.MapSchemaUnitOptionsToUnitFile(su.Options)
+
+	_, err := os.Stat(file)
+	if err == nil {
+		b, err := getUnitFromFile(file)
```
antrik: You are actually changing the error handling logic here compared to the original code in `warnOnDifferentLocalUnit()`. I'm pretty sure there is a good reason why it was done like that...

api/units.go
```diff
@@ -76,6 +76,7 @@ func (ur *unitsResource) set(rw http.ResponseWriter, req *http.Request, item str
 	}
 
 	var su schema.Unit
+	newUnit := false
```
antrik: Better to introduce the variable closer to its use -- in this case, immediately before the conditional block that sets it.

functional/unit_action_test.go
```diff
@@ -224,3 +226,68 @@ func TestUnitSSHActions(t *testing.T) {
...
+		stdout, _, err = cluster.Fleetctl(m, "list-units", "--no-legend")
+		if err != nil {
+			return fmt.Errorf("Failed to run list-units: %v", err)
+		}
+		if strings.TrimSpace(stdout) == "" || maxAttempts == 0 {
+			found = true
+			break
+		}
+		maxAttempts--
+	}
```
antrik: Why does this run the destroy command multiple times? Shouldn't it just run the command once, and then wait for its outcome?

functional/unit_action_test.go
```diff
@@ -227,6 +257,74 @@ func TestUnitSSHActions(t *testing.T) {
...
+		return fmt.Errorf("Unable to %s fleet unit: %v", cmd, err)
+	}
+	stdout, _, err := cluster.Fleetctl(m, "list-units", "--no-legend")
+	if err != nil {
+		return fmt.Errorf("Failed to run list-units: %v", err)
+	}
+	units := strings.Split(strings.TrimSpace(stdout), "\n")
+	if len(units) != 1 {
+		return fmt.Errorf("Did not find 1 unit in cluster: \n%s", stdout)
+	}
```
antrik: I'm a little confused here: won't `list-units` always remain empty for `submit`?

functional/unit_action_test.go
```diff
@@ -227,6 +257,74 @@ func TestUnitSSHActions(t *testing.T) {
...
+		return fmt.Errorf("Unable to replace fleet unit: %v", err)
+	}
+	stdout, _, err = cluster.Fleetctl(m, "list-units", "--no-legend")
+	if err != nil {
+		return fmt.Errorf("Failed to run list-units: %v", err)
+	}
+	units = strings.Split(strings.TrimSpace(stdout), "\n")
+	if len(units) != 1 {
+		return fmt.Errorf("Did not find 1 unit in cluster: \n%s", stdout)
+	}
```
antrik: AIUI this will only test that the number of units hasn't changed; but not whether the unit has actually been replaced by the new one?

functional/unit_action_test.go
antrik: Hm, probably I got confused between systemctl and fleetctl... Please disregard.

functional/unit_action_test.go
```diff
@@ -313,6 +313,9 @@ func replaceUnitCommon(cmd string) error {
 	if len(units) != 1 {
 		return fmt.Errorf("Did not find 1 unit in cluster: \n%s", stdout)
 	}
+	if err := waitForActiveUnitsReplaceCmds(cluster, m, cmd, 1); err != nil {
+		return err
+	}
```
antrik: @dongsupark I think there has been a misunderstanding here: my point was not to invoke the wait function after already having checked that we have the expected number of units -- that won't make it any more reliable as far as I can tell. Rather, my point was that *instead* of a one-off check directly after the command, it is probably better to use the wait function to see whether the expected state is reached *eventually*... At least that's what all the existing tests I have looked at so far seem to be doing.

(Of course that only applies to cases where we are actually expecting units to become active respectively to cease being active...)

fleetctl/fleetctl.go
antrik: Ah, so this is meant only to be an assertion... That make sense :-) So: yes, please make this more explicit.

fleetctl/fleetctl.go
antrik: Yeah, I'm aware of that. My point was that the code in question is trivial, and wrapping it in a function doesn't really simplify the invocation...

(BTW, the function is way too complicated -- just `return a.Hash() == b.Hash()` would do exactly the same... And it would make it more obvious that the function is trivial.)

fleetctl/fleetctl.go
dongsupark: > Please follow proper orthography: sentences always begin with a capital letter, and end in a full stop.

Basically I agree with the principle. But in this case I'd rather not capitalize the beginning letter, because of confusion during pattern matching in the editor. I'd rather rewrite the sentence, not to start a sentence with a function name.

Besides I would normally just leave comments written by others, given the fact that every comment-writing already requires a fair amount of effort. Think about other bunch of functions without any comment. A wrong-spelled comment is much more valuable than no comment. ;-)

fleetctl/fleetctl.go
dongsupark: @antrik
> You are actually changing the error handling logic here compared to the original code in warnOnDifferentLocalUnit(). I'm pretty sure there is a good reason why it was done like that...

I'm not sure I'm following you. There's no function ``warnOnDifferentLocalUnit()``. This error handling style looks actually different from the normal convention, but at least it's not wrong, is it?

functional/unit_action_test.go
dongsupark: @antrik
> Why does this run the destroy command multiple times? Shouldn't it just run the command once, and then wait for its outcome?

Good question. I tried to do run the destroy command only once, expecting it to eventually succeed. But it resulted in occasional failure. I'm not sure why, and as you might know, that's not the only unexpected error that we see. So I had to write such a hack. I'd actually love to see this hack go away.

functional/unit_action_test.go
dongsupark: @antrik
> AIUI this will only test that the number of units hasn't changed; but not whether the unit has actually been replaced by the new one?

Correct. Let me think about how to actually confirm that each unit content has been replaced.

api/units.go
tixxdz: Not sure why it's better ? if it's just "golang allows it", then for me it's better to have all variables in same place... and this function is not that long to justify this... anyway this is not a big deal, unless you show some optimized generated code or some other official documentation?!

functional/unit_action_test.go
dongsupark: @antrik
> my point was that instead of a one-off check directly after the command, it is probably better to use the wait function to see whether the expected state is reached eventually... At least that's what all the existing tests I have looked at so far seem to be doing.

All right. I made that change, and tested it. It looks fine.
Pushed to my branch: https://github.com/endocode/fleet/tree/dongsu/fleetctl-replace-unit-fxtests
Hopefully that will be merged to this PR.

fleetctl/fleetctl.go
antrik: Sorry, this was not the best place to put this remark: I actually totally agree that the function name should not be capitalised. (And by Go conventions, the sentence actually has to start with the function name... So it basically has to be ugly.)

My remark about sentences starting with a capital letter was actually referring to some of the other comments, where some sentences not starting with an identifier were not capitalised either...

fleetctl/fleetctl.go
antrik: @dongsupark the mentioned function existed in the original code: https://github.com/coreos/fleet/blob/master/fleetctl/fleetctl.go#L729 ; part if it's logic ended up here -- but it was altered in the process. Specifically, the original had explicit handling for `os.IsNotExist(err)`.

functional/unit_action_test.go
tixxdz: @dongsupark you could do "fleetctl cat" and see if the strings contain some new data or check the logs...

api/units.go
antrik: It's easier to find the definition if it's close to the use, end vice versa.

It's not just Go: all modern languages I know allow this -- including even C99...

More to the point, it appears to be a pattern used throughout the existing fleet code -- so you need a good reason for *not* following it.

functional/unit_action_test.go
antrik: Well, if it fails to work on the first try, that's a bug that the test suite should point out rather than ignore...

Also, I wonder whether you actually checked that it doesn't get destroyed eventually (as opposed to immediately) within a reasonable time frame?...

dongsupark: @antrik I'll create a separate issue for this issue. Otherwise we would definitely forget about this bug.

functional/unit_action_test.go
dongsupark: @tixxdz
> you could do "fleetctl cat" and see if the strings contain some new data or check the logs...

Done. See my tree: https://github.com/endocode/fleet/tree/dongsu/fleetctl-replace-unit-fxtests

jonboulle:
let me know when this is ready for review..

dongsupark:
@antrik I added a new commit to change minor things as you wrote today.
@jonboulle I think this tree is ready for review.

fleetctl/fleetctl.go
```diff
@@ -663,6 +671,85 @@ func createUnit(name string, uf *unit.UnitFile) (*schema.Unit, error) {
...
+	if unit == nil {
+		if sharedFlags.Replace {
+			log.Debugf("Unit(%s) was not found in Registry", name)
+		}
+		// Create a new unit
+		return 0, nil
+	}
+
+	// if sharedFlags.Replace is not set then we warn
+	different, err := isLocalUnitDifferent(arg, unit, !sharedFlags.Replace, false)
```
antrik: I think I see now why I find this code so confusing: you have the `fatal` error handling flag in `isLocalUnitDifferent()`, but you ended up not using it at all, and instead doing the error handling here. Please decide on one *or* the other, and clean out the dead code paths...

antrik: Part of the confusion is also because you now do the warning handling inside `isLocalUnitDifferent()`, but the "fatal" error handling outside. If you could somehow shuffle them around to be in the same place (I don't really care which, as long as it's the same), that would certainly improve readability.

api/units.go
```diff
@@ -101,15 +101,33 @@ func (ur *unitsResource) set(rw http.ResponseWriter, req *http.Request, item str
...
+			return
 		} else {
-			ur.create(rw, su.Name, &su)
+			newUnit = true
 		}
+	} else if eu.Name == su.Name && len(su.Options) > 0 {
+		// There is already a unit with the same name
+		// check the hashes if they do not match then we will
+		// create a new unit with the same name and later
+		// the job will be updated to this new unit.
```
antrik: So rather than actually replacing the existing unit in the registry, you are just adding a new one with the same name? Why do you need to do it like this? By my intuition, this is bound to cause problems...

Also, where/how will the job be updated later? And what happens to the old unit? This all needs a lot more explanation...

antrik:
Please correct me if I'm wrong; but as far as I can see, nowhere in the fleetctl code we wait after `--replace` for the unit(s) to actually get updated to the new versions? Shouldn't we do that (if in blocking mode), just as we wait for units to reach the desired run state?...

api/units.go
tixxdz: > So rather than actually replacing the existing unit in the registry, you are just adding a new one with the same name? Why do you need to do it like this? By my intuition, this is bound to cause problems...
How you think this should be done ? what problems ?

> Also, where/how will the job be updated later? And what happens to the old unit? This all needs a lot more explanation...

We instruct etcd to update the job key to point to the right target, and everything else is handled automatically...

The old unit is not referenced anymore, this may relate in part of @dongsupark PR https://github.com/coreos/fleet/pull/1510 , we may insert an explicit destroy unit hash before starting the new unit, but still not sure! IMO this should be fixed later not in this series if it really matters.

tixxdz:
@antrik 

> Please correct me if I'm wrong; but as far as I can see, nowhere in the fleetctl code we wait after --replace for the unit(s) to actually get updated to the new versions? Shouldn't we do that (if in blocking mode), just as we wait for units to reach the desired run state?...

It's automatically handled when we update the job key, we only wait to reach the inactive state then start again...

fleetctl/fleetctl.go
tixxdz: > Part of the confusion is also because you now do the warning handling inside isLocalUnitDifferent(), but the "fatal" error handling outside. If you could somehow shuffle them around to be in the same place (I don't really care which, as long as it's the same), that would certainly improve readability.

I can't see here where "warn if units differ" relates to fatal error ?!
And we check if local unit is either a simple unit or an instance of template unit

functional/unit_action_test.go
tixxdz: OK, thank you!

fleetctl/fleetctl.go
antrik: These code paths are essentially checking the same conditions as far as I can tell; only the reactions differ depending on the mode. Not putting the handling in one place makes it more redundant and much harder to follow IMHO.

tixxdz: Hmm, what same conditions ? "warn if units differ" is not an "error"...  ? I don't follow, sorry.

antrik: I'm talking about the nested case handling here for `sharedFlags.Replace`, `different`, and `err` having considerable overlap with the case handling within `isLocalUnitDifferent()`.

fleetctl/fleetctl.go
```diff
@@ -726,24 +809,65 @@ func lazyCreateUnits(args []string) error {
...
+		return !result, nil
+	} else if fatal {
+		return false, err
+	}
+
+	info := unit.NewUnitNameInfo(path.Base(file))
+	if info == nil {
+		return false, fmt.Errorf("error extracting information from unit name %s", file)
+	} else if !info.IsInstance() {
+		return false, fmt.Errorf("error Unit %s does not seem to be a template unit", file)
```
antrik: It seems you have the same logic error here you had in one of your earlier refactoring patches: there is no reason to assume at this point that the Unit is *supposed* to be a template unit...

api/units.go
antrik: > > So rather than actually replacing the existing unit in the registry, you are just adding a new one with the same name? Why do you need to do it like this? By my intuition, this is bound to cause problems...
> How you think this should be done ?

My thinking was that replacing the unit in the registry should be done in one go -- ideally with an atomic operation; or if etcd doesn't offer such functionality, by deleting the old and adding the new directly after.

Note that I don't have experience with etcd, or how fleet uses it exactly. I might be missing something crucial here -- so I was hoping for some clarification in case I'm totally off...

> what problems ?

My understanding is that until now fleet wasn't ever running into situations where two units with the same name existed in the registry at the same time; and thus existing code probably isn't prepared to deal with this happening...

I might be missing something here as well of course.

> We instruct etcd to update the job key to point to the right target, and everything else is handled automatically...

Where does that happen? I don't see anything related to this in the PR; so I have to assume it happens somewhere in pre-existing code -- but where? Again, I'd appreciate some pointer, so I don't have to dig it up myself...

fleetctl/fleetctl.go
tixxdz: Hmm I thought from your previous comment was about "warn" and "fatal" ?! anyway, you are missing lot of points here... previous code did not even treat errors, so errors are ignored when sharedFlags.Replace is not set, if you do it: 1) you will just break backward compatibility, you can test it...   **they do not overlap**, and 2) previous logic was just printing a warning which we still do for both simple units or instances of template, 3) for Replace we will catch errors since we want a valid file on the filesystem to use for replace... handling of `different` is done in the caller also errors! I really don't see how you mix `warn if units differ`, anyway could you please just be more specific ?

api/units.go
tixxdz: > My thinking was that replacing the unit in the registry should be done in one go -- ideally with an atomic operation; or if etcd doesn't offer such functionality, by deleting the old and adding the new directly after.
Note that I don't have experience with etcd, or how fleet uses it exactly. I might be missing something crucial here -- so I was hoping for some clarification in case I'm totally off...
 what problems ?
My understanding is that until now fleet wasn't ever running into situations where two units with the same name existed in the registry at the same time; and thus existing code probably isn't prepared to deal with this happening...
I might be missing something here as well of course.

I really think that you should **read the code,** the commit logs and **test** it before you comment please!

We are **replacing** the job in the etcd registry, units content is not the same anymore, new operations will refer to the new updated version just after the job is updated to point to the right target and that's where units are fetched from the first place...

For other questions, sorry but I can't clarify something I don't understand, so if you see real problems please point them ? and we will try to fix them.

Thanks!

fleetctl/fleetctl.go
tixxdz: Why ? 

> there is no reason to assume at this point that the Unit is supposed to be a template unit...
Sorry I don't get you at all! that's the same code before, and errors are ignored for `!Replace` cases for backward compatibility... but for replace we want to compare template units otherwise we can't replace them nor their instances...

Sorry but I don't get your comments, could you please be more specific ? Thank you!

api/units.go
antrik: Calm down, I was merely asking for some clarification. The code comment doesn't really make it clear to me how this is supposed to work. I guess that's because I don't know the fleet code and it's etcd interactions by heart -- but then, I think comments should make also make sense to people who don't... Those who know it by heart probably don't even need comments.

tixxdz: > Calm down, I was merely asking for some clarification. The code comment doesn't really make it clear to me how this is supposed to work. I guess that's because I don't know the fleet code and it's etcd interactions by heart -- but then, I think comments should also make sense to people who don't... Those who know it by heart probably don't even need comments.

It's ok, the whole logic was explained in these two patches and the code is commented:
https://github.com/coreos/fleet/pull/1509/commits/0e675d9dca2ebd42d296bde75466e2e49b6daa47
https://github.com/coreos/fleet/pull/1509/commits/cb8fd41c5f3ef0f223d81bb8828c155cdac6d1b2

So that's why I didn't get those comments, and yes please if you think that there are real issues, then sure we are doing something wrong here ;-)

antrik: Well, the commit messages and code changes *seem* to suggest that just submitting a unit with the same name will replace the existing one; which was my original understanding of how this is supposed to work. It's the code comment here that threw me off. I think the wording is *extremely* confusing -- especially the ominous "later"... I suspect what you are actually trying to say is that the reconciler will update the running jobs once the unit file has been replaced in the registry -- but that's not clear at all.

tixxdz: > Well, the commit messages and code changes seem to suggest that just submitting a unit with the same name will replace the existing one; which was my original understanding of how this is supposed to work.

It's clearly commented:
https://github.com/coreos/fleet/pull/1509/commits/0e675d9dca2ebd42d296bde75466e2e49b6daa47#diff-273bb648504b6b04dd506aa86e10e573R118

>  It's the code comment here that threw me off. I think the wording is extremely confusing -- especially the ominous "later"... I suspect what you are actually trying to say is that the reconciler will update the running jobs once the unit file has been replaced in the registry -- but that's not clear at all.

Oh sorry @antrik it seems that you are still with the old idea, yep! so yeh it is still not clear! 

So by later we just mean update the job key to point to the new unit that we have to create anyway, and which will be in the inactive state, which in turn will trigger a unit stop or unload for the old unit if necessary, and make the new unit in the inactive state. So we don't wait for the reconciler at all which will do the same job but might take longer and needs code..., we don't trigger that path, and we don't add more code to update desired and current hashes inside the job keys. All we do: is set job keys manually to the new unit (again that we have to create manually anyway...) and put it into the "inactive" state, and everything else continue to work automatically... and we use the same wait for block logic for inactive, load or start states... and current fleet transition design to the final state is what helped here.

Thank you!

antrik: Not sure what you mean by "the old idea" -- but it doesn't really matter. While I'm still not clear on all the specifics (which I guess is not possible without reading a lot more code), I do think I have a *rough* idea now how it works. It would be good if you could update the code comment with a simliar explanation...

dongsupark:
@tixxdz
As #1544 introduced cluster operations like ``WaitForNUnits``, ``WaitForNUnitFiles``, I had to also update this PR, to avoid conflicts. See https://github.com/endocode/fleet/tree/dongsu/fleetctl-replace-unit-fxtests .
What I did:
* Rebased ``tixxdz/fleetctl-replace-unit-v2`` onto ``dongsu/fleetctl-unit-action-fxtests`` (#1544)
* Changed functional tests for replace option, so that they make use of ``cluster.WaitForNUnits()`` and ``cluster.WaitForNUnitFiles()``.
* Squashed functional test commits into 3 commits in total.

I didn't touch the main patches for the replace option. When you later got time to update this PR, please take also my patches from https://github.com/endocode/fleet/tree/dongsu/fleetctl-replace-unit-fxtests. Thanks!

tixxdz:
@jonboulle it's ready, basically I ended up implementing it in an other way, please see comment: https://github.com/coreos/fleet/pull/1509#discussion_r58524088   and also to make sure that `ExecStartPre` of new unit is serialized after `ExecStopPost` of the old version of the unit, so we don't have further conflicts between the two versions of the unit, at same time it allows us to try to avoid races which may bring us to https://github.com/coreos/fleet/issues/1000 and https://github.com/systemd/systemd/issues/518 . 

Another advantage we don't add replace target hashes nor wait from the client perspective fleetctl for the hashes to be updated to the new version, we just wait for the new unit (new hash) to reach its own desired state: submitted, loaded or started as it's currently done. The old unit will be stopped or unloaded before that when we set the inactive state. So instead of waiting for the reconcile, we just do what it does but without waiting and without adding so much code.

Thank you!

tixxdz:
For the record, patches 9, 10, 11 lgtm, I just added a complex test to cover what I described in this comment: https://github.com/coreos/fleet/pull/1509#issuecomment-211829316  the test is documented and this allows us to make sure that systemd directives are serialized but only in the current replace context, external operations are not our fault.

Waiting for some lgtm, then will merge it, we have been rebasing this branch for some time now...

Thanks!

dongsupark:
``TestReplaceSerialization`` test looks good. Thanks!

kayrus:
@tixxdz there is one issue I've just found. When you replace templated unit, updated unit has being scheduled on different machine. Steps to reproduce:

```sh
fleetct submit hello@.service
fleetct start hello@{1..10}.service
```

Then you'll get the units list:

```
fleetctl list-unit-files                      
hello@.service          4e11fe3 inactive        inactive        -
hello@1.service         4e11fe3 launched        launched        63a6ca6f.../coreos1
hello@10.service        79699e3 launched        launched        5e719f91.../coreos2
hello@2.service         79699e3 launched        launched        5e719f91.../coreos2
hello@3.service         79699e3 loaded          loaded          8c849f40.../coreos3
hello@4.service         79699e3 launched        launched        8c849f40.../coreos3
hello@5.service         79699e3 launched        launched        5e719f91.../coreos2
hello@6.service         79699e3 launched        launched        8c849f40.../coreos3
hello@7.service         79699e3 launched        launched        5e719f91.../coreos2
hello@8.service         79699e3 launched        launched        8c849f40.../coreos3
hello@9.service         79699e3 launched        launched        5e719f91.../coreos2
```

Then make replace:

```
fleetctl start --replace hello\@4.service
Unit hello@4.service inactive
Unit hello@4.service launched on 63a6ca6f.../coreos1
```

And you'll see that unit has been moved from to coreos3 to coreos1:

```
fleetctl list-unit-files                 
hello@.service          4e11fe3 inactive        inactive        -
hello@1.service         4e11fe3 launched        launched        63a6ca6f.../coreos1
hello@10.service        79699e3 launched        launched        5e719f91.../coreos2
hello@2.service         79699e3 launched        launched        5e719f91.../coreos2
hello@3.service         79699e3 loaded          loaded          8c849f40.../coreos3
hello@4.service         4e11fe3 launched        inactive        63a6ca6f.../coreos1
hello@5.service         79699e3 launched        launched        5e719f91.../coreos2
hello@6.service         79699e3 launched        launched        8c849f40.../coreos3
hello@7.service         79699e3 launched        launched        5e719f91.../coreos2
hello@8.service         79699e3 launched        launched        8c849f40.../coreos3
hello@9.service         79699e3 launched        launched        5e719f91.../coreos2
```

Should we document that behavior or should be reschedule unit on the same machine it was started before and move it only when there is a explicit need or requirement? Otherwise when you use extra `X-Fleet` options, it works as expected, i.e.:

```ini
[X-Fleet]
MachineMetadata="hostname=coreos%i"
```

tixxdz:
@kayrus fleetctl submit; load or start deal with the notion of cluster, and not a single specified machine. If we go that path then IMO we will reduce fleet functionality and further changes. If we schedule on the same machine:

1) We are explicitly setting `"[X-Fleet] MachineMetadata="hostname=coreos%i""` for users that do not want it.

2) What happens if a user wants to replace a unit and schedule it based on fleet decisions or any other scheduling decision ? if we do that, then we explicitly chose for him and for sure it will end up re-scheduling all units again and again on the same node. Reducing the notion of migration and automatic replace that we are trying to add here.

3) We will hardcode that inside fleet logic, which I really do not like. At long term it will be a burden and we may see requests to change it.

4) It will conflict with `[X-Fleet] MachineMetadata="hostname=coreos-another-one"` then we have to add code to make one take precedence over the other... 

Now if users want to replace on the same machine, there is already the option:
`[X-Fleet]
MachineMetadata="hostname=coreos%i"`

This allows us to solve previous points:
1) Users who want it, will have to set it.

2) Users who want fleet to do right thing and do not bother are satisfied.

3) Nothing is hardcoded inside the code, everything is optional and one can even move units and services from one cluster with X number of machines to another cluster of Y number of machines.

4) We only have one and unique option we do not add extra logic nor code.

Now for the record we have to update fleetctl help message to let users that his replace a unit in the cluster and not into a specific node, now if you want it to be in node X, then add `MachineMetadata` option, which you can do it where previous versions of units are still running.

Will update the "--replace" help message right now. Thanks!

kayrus:
@tixxdz I've tried to not unload units, but just stop them. This doesn't work as expected. Fleet for some reason hangs. Other solution is to implement "desiredMachine" option, which could be ignored in case when unit could not be replaced on the same machine (i.e. `[X-Fleet]` options were added). But this is another issue and could be resolved in new PR if necessary.

tixxdz:
@kayrus actually yes, we have to unload, free resources, reset status of failed units through systemd manager, and other operations to clean up things.

For the `rescheduleOnSameMachine` that's would be a nice addition not only for this PR, but probably for other parts or PRs.

Thank you!

tixxdz:
So this one looks good, all tests pass, many comments. I'm merging it now to unblock the situation, so it won't be lost nor the upcoming PRs.

For users who want to replace units on the same machine they could use what's already being used:
`[X-Fleet] MachineMetadata="hostname=coreos%i"` 

Later one may add `rescheduleOnSameMachine`.

@jonboulle in all ways, we have to unload then load, hence the rescheduling.

Thank you!

Issue #1456, opened 2016-03-02
usability: garbage collection in /_coreos.com/fleet/job registry directory


kayrus:
steps to reproduce:

1) check etcd registry:

```sh
$ etcdctl ls --recursive /_coreos.com
/_coreos.com/fleet
/_coreos.com/fleet/machines
/_coreos.com/fleet/machines/78b98c5dfe468973b93e098a5a5829ea
/_coreos.com/fleet/machines/78b98c5dfe468973b93e098a5a5829ea/object
/_coreos.com/fleet/machines/f8f4ec94d4193d2f160e8bd22ed360be
/_coreos.com/fleet/machines/f8f4ec94d4193d2f160e8bd22ed360be/object
/_coreos.com/fleet/machines/0903835ba13faf7f62cf5d8ddc7812de
/_coreos.com/fleet/machines/0903835ba13faf7f62cf5d8ddc7812de/object
/_coreos.com/fleet/engine
/_coreos.com/fleet/engine/version
/_coreos.com/fleet/lease
/_coreos.com/fleet/lease/engine-leader
```

2) create template unit:

```ini
[Unit]
Description=My Service

[Service]
TimeoutStartSec=0
ExecStart=/bin/sh -c "trap 'exit 0' INT TERM; while true; do echo Hello World %i; sleep 10; done"
```
3) submit and start units:

```sh
$ fleetctl submit hello@.service          
Unit hello@.service inactive
$ fleetctl start hello@{1..10}.service      
Unit hello@1.service inactive
Unit hello@2.service inactive
Unit hello@3.service inactive
Unit hello@4.service inactive
Unit hello@5.service inactive
Unit hello@6.service inactive
Unit hello@7.service inactive
Unit hello@8.service inactive
Unit hello@9.service inactive
Unit hello@10.service inactive
Unit hello@2.service launched
Unit hello@4.service launched
Unit hello@5.service launched
Unit hello@1.service launched on 78b98c5d.../192.168.122.188
Unit hello@3.service launched on f8f4ec94.../192.168.122.135
Unit hello@6.service launched on f8f4ec94.../192.168.122.135
Unit hello@7.service launched on 78b98c5d.../192.168.122.188
Unit hello@8.service launched on 0903835b.../192.168.122.166
Unit hello@10.service launched on f8f4ec94.../192.168.122.135
Unit hello@9.service launched on 0903835b.../192.168.122.166
```

4) destroy units:

```sh
[...]

PR #1510, opened 2016-03-18
registry: remove units from etcd registry upon DestroyUnit()

dongsupark:
So far each command ``"fleetctl destroy unit"`` has removed job entries from
the etcd registry, under ``/_coreos.com/fleet/job``. But it has not removed
its unit file, under ``/_coreos.com/fleet/unit``. As a result, fleet left
lots of garbages in the etcd registry, so users had to manually clean
them up.

So this patch gets unit contents deleted actually from etcd registry
when ``DestroyUnit()`` gets called. To avoid potential hash collisions,
it first fetches a list of units from registry, to check there's any
duplicated entry. Only if no duplicated unit is found, fleetd actually
deletes the unit from registry.

Fixes: https://github.com/coreos/fleet/issues/1456
Fixes: https://github.com/coreos/fleet/issues/1290
Reference: https://github.com/coreos/fleet/pull/1291

/cc @kayrus @wuqixuan


registry/job.go
```diff
@@ -316,9 +365,81 @@ func (r *EtcdRegistry) DestroyUnit(name string) error {
 	}
 
 	// TODO(jonboulle): add unit reference counting and actually destroying Units
+
```
tixxdz: @jonboulle could you please remember what was the use case that made you add this comment ? "unit reference counting" is there some optimization involved or something else ? thanks :-)

tixxdz:
@dongsupark is a functional test possible for this case ?

registry/job.go
jonboulle: The problem is that the delete operation needs to be atomic (essentially a
compare and delete, "delete this unit file IFF no other units reference
it"), but that information is not stored in one place in etcd at the
moment. That's where a reference count would come in.

I haven't looked at the PR code yet but - how does it deal with this
scenario of a race between destroying a unit and creating a new unit with
an identical unit file hash?

dongsupark:
@tixxdz First of all, I don't think it's possible to test such an extreme case that 2 arbitrary units end up having the same sha1 hash. Theoretically it would be possible, but how would it be practically look like...?
But it's probably possible to write a basic test for submitting & destroying multiple units, to verify that a correct unit has been actually removed. ``functional/unit_action_test.go`` doesn't seem to have such tests.

tixxdz:
@dongsupark yes of course! yeh I was referencing "verify if the unit was actually removed" parts, thanks!

jonboulle:
Creating units with the same hash is easy. echo blabla > foo.service; cp
foo.service bar.service; fleetctl submit foo.service bar.service

tixxdz:
@jonboulle thx!

registry/job.go
dongsupark: @jonboulle
> The problem is that the delete operation needs to be atomic (essentially a compare and delete, "delete this unit file IFF no other units reference it"), but that information is not stored in one place in etcd at the moment. That's where a reference count would come in.
> I haven't looked at the PR code yet but - how does it deal with this scenario of a race between destroying a unit and creating a new unit with an identical unit file hash?

Yeah, I see your point. This PR is basically not about changing the current design of fleet. I don't think I'm capable of doing it. Currently I don't have a better idea about how to do refcounting to prevent such a race between creating and destroying a unit.
But on the other hand, I think that the delete operation needs to be somehow implemented, from the perspective of usability. Thus I had to make a compromise, writing such a lengthy verification function before removing a unit. Of course it's not perfect, but that's probably one of doable things.

dongsupark:
@tixxdz 
> is a functional test possible for this case ? 

I added a new functional test ``TestUnitDestroyFromRegistry``, which does not work correctly. On semaphore, ``etcdctl ls /_coreos.com/fleet/job/hello.service/object`` fails, because there's no such file. Funny enough, everything works fine during my local test. Not sure what's going on.

dongsupark:
So finally I managed to get the new functional test ``TestUnitDestroyFromRegistry`` working. It has failed so far because of a wrong keyspace, which should have been ``/fleet_functional/smoke`` instead of the default ``/_coreos.com/fleet`` Solution was to export ``cluster.Keyspace()`` to make it usable from functional tests correctly. (I also fixed other minor bugs, but I'll not get into the detail...)

Though the Semaphoreci still does not succeed completely. Other tests like ``TestNodeShutdown`` fail randomly, but I think that has nothing to do with my test. That's probably because of the on-going maintenance work from Semaphore.

functional/unit_action_test.go
```diff
@@ -224,3 +229,151 @@ func TestUnitSSHActions(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// submit a unit and assert it shows up
+	if _, _, err := cluster.Fleetctl(m, "submit", "fixtures/units/hello.service"); err != nil {
+		t.Fatalf("Unable to submit fleet unit: %v", err)
+	}
+	var stdout string
+	stdout, _, err = cluster.Fleetctl(m, "list-units", "--no-legend")
```
antrik: Don't define the variable separately: use `stdout, _, err := ...` instead.

functional/unit_action_test.go
```diff
@@ -224,3 +229,151 @@ func TestUnitSSHActions(t *testing.T) {
...
+	units := strings.Split(strings.TrimSpace(stdout), "\n")
+	if len(units) != 1 {
+		t.Fatalf("Did not find 1 unit in cluster: \n%s", stdout)
+	}
+
+	// cat the unit and compare it with the value in etcd registry
+	if stdout, _, err = cluster.Fleetctl(m, "cat", "hello.service"); err != nil {
+		t.Fatalf("Unable to retrieve the fleet unit: %v", err)
+	}
+	unitBody := stdout
```
antrik: Don't use a temporary `stdout` -- just assign the result to `unitBody` directly.

(Note: This will require the assignment to be moved out of the `if` -- but that's the common pattern used throughout the existing codebase...)

functional/unit_action_test.go
```diff
@@ -224,3 +229,151 @@ func TestUnitSSHActions(t *testing.T) {
...
+	if len(stdout) != 0 && len(units) != 1 {
+		t.Fatalf("Did not find 1 unit in cluster: \n%s", stdout)
+	}
+
+	// check for the unit being destroyed from the etcd registry,
+	// /fleet_functional/smoke/unit/.
+	// NOTE: do not check error of etcdctl, as it returns 4 on an empty list.
+	etcdUnitPrefix := path.Join(cluster.Keyspace(), "unit")
+	etcdUnitPath := path.Join(etcdUnitPrefix, hashUnit)
+	stdout, _, err = util.RunEtcdctl("ls", etcdUnitPath)
```
antrik: If you don't check the err, don't assign it in the first place -- just use another `_` instead.

registry/job.go
```diff
@@ -316,9 +365,81 @@ func (r *EtcdRegistry) DestroyUnit(name string) error {
...
+	// Delete unit from the etcd registry
+	if u != nil {
+		var unitMatch bool
+		// check if the unit is really valid. If not, return err.
+		key = r.hashedUnitPath(u.Unit.Hash())
+		if unitMatch, err = r.checkUnitMatch(u); err != nil {
```
antrik: Don't define `unitMatch` separately. (See my other remark about this...)

registry/job.go
```diff
@@ -316,9 +365,81 @@ func (r *EtcdRegistry) DestroyUnit(name string) error {
...
+			err = errors.New("cannot check for the unit is valid")
+			return err
```
antrik: Better wording: "Unit validity check failed", or "Failed checking unit validity".

Also, you probably want to use `fmt.Errorf()` to append the original error message, as done in other places.

And there is no need to assign it to `err` here -- just return the result directly...

registry/job.go
```diff
@@ -316,9 +365,81 @@ func (r *EtcdRegistry) DestroyUnit(name string) error {
...
+		}
+		if unitMatch {
+			err = errors.New("Invalid unit in the etcd registry. Skip deleting the unit.")
```
antrik: I guess this should be something along these lines: "Invalid unit in the etcd registry: not deleting from registry."

(I don't really understand the error condition though -- so not sure whether this is right...)

registry/job.go
```diff
@@ -316,9 +365,81 @@ func (r *EtcdRegistry) DestroyUnit(name string) error {
...
+			return err
+		}
+		_, err = r.kAPI.Delete(r.ctx(), key, opts)
+		if err != nil {
+			if isEtcdError(err, etcd.ErrorCodeKeyNotFound) {
+				err = errors.New("cannot delete unit from the registry. Unit not found.")
+			}
+			return err
```
antrik: Is it actually a failing error if this operation failed? Or maybe it should just print a warning to the log and ignore it?

Also, do you really need the case handling here? Right now all errors are handled the same anyway -- so you can just use `Errorf()` as usual: `return fmt.Errorf("Failed deleting unit from registry: %s", err)`

registry/job.go
antrik: The question here is whether it's acceptable to introduce a race condition potentially leading to hard failure, only to handle something that is merely an annoyance... Doesn't sound like a good tradeoff to me.

dongsupark: > The question here is whether it's acceptable to introduce a race condition potentially leading to hard failure, only to handle something that is merely an annoyance... Doesn't sound like a good tradeoff to me.

If anyone could provide a better patch for this issue, I would be happy to close this PR in favor of the new one. Until then, I'd still think this PR is the only one I can come up with. OTOH given the fact that the 'TODO' comment was written 2 years ago, I don't expect anything to suddenly happen in the near future.

dongsupark:
@antrik ok, I changed minor things as you suggested, and force-pushed.

BTW after rebasing onto the current master, fleet suddenly stopped to build, because commit 64ec3a433c7de606fe46732837b9f66ec5b2cfc0 removed ``HashFromHexString``. So I had to resurrect the function in ``registry/job.go``. Now it's working.

registry/job.go
tixxdz: @dongsu, fair enough! thanks for the patches.

kayrus:
@dongsupark function was restored. that was temporarily behavior. sorry for confusing.

dongsupark:
@kayrus Thanks! :-)
I just removed the temporary commit for ``HashFromHexString`` to use again ``unit.HashFromHexString`` like before, and rebased it onto master.

PR #1512, opened 2016-03-21
tests: added template + metadata functional test

kayrus:
Covers https://github.com/coreos/fleet/issues/1446

PR #1515, opened 2016-03-22
fleetd:registry: improve Units() performance remove extra loops when fetching Jobs list from etcd

tixxdz:
Optimize Units() call when we fetch and process Jobs and Units from etcd.

We tested this with 500 services on 3 nodes where previously it would take the fleetctl client to get the list of units from fleet and etcd ~4secs  to now ~0.9secs.

Actually this will discharge fleet from doing extra sorting logic and two extra loops at the same process. One thing to note is that we request etcd to sort the Job Units for us, so I removed it from fleet, and it seems it is the same sorting logic.

### DIFF
--- a/functional/unit_action_test.go
+++ b/functional/unit_action_test.go
@@ -15,8 +15,11 @@
 package functional
 
 import (
+	"fmt"
 	"io/ioutil"
 	"path"
+	"reflect"
+	"sort"
 	"strings"
 	"testing"
 
@@ -402,5 +405,65 @@ func TestUnitStatus(t *testing.T) {
 	if !strings.Contains(stdout, "Loaded: loaded") {
 		t.Errorf("Could not find expected string in status output:\n%s\nstderr:\n%s",
 			stdout, stderr)
+	}
+}
+
+// TestListUnitFilesOrder simply checks if "fleetctl list-unit-files" returns
+// an ordered list of units
+func TestListUnitFilesOrder(t *testing.T) {
+	cluster, err := platform.NewNspawnCluster("smoke")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cluster.Destroy()
+
+	m, err := cluster.CreateMember()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = cluster.WaitForNMachines(m, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Combine units
+	var units []string
+	for i := 1; i <= 20; i++ {
+		unit := fmt.Sprintf("fixtures/units/hello@%02d.service", i)
+		stdout, stderr, err := cluster.Fleetctl(m, "submit", unit)
+		if err != nil {
+			t.Fatalf("Failed to submit a batch of units: \nstdout: %s\nstder: %s\nerr: %v", stdout, stderr, err)
+		}
+		units = append(units, unit)
+	}
+
+	// make sure that all unit files will show up
+	_, err = cluster.WaitForNUnitFiles(m, 20)
+	if err != nil {
+		t.Fatal("Failed to run list-unit-files: %v", err)
+	}
+
+	stdout, _, err := cluster.Fleetctl(m, "list-unit-files", "--no-legend", "--fields", "unit")
+	if err != nil {
+		t.Fatal("Failed to run list-unit-files: %v", err)
+	}
+
+	outUnits := strings.Split(strings.TrimSpace(stdout), "\n")
+
+	var sortable sort.StringSlice
+	for _, name := range units {
+		n := path.Base(name)
+		sortable = append(sortable, n)
+	}
+	sortable.Sort()
+
+	var inUnits []string
+	for _, name := range sortable {
+		inUnits = append(inUnits, name)
+	}
+
+	if !reflect.DeepEqual(inUnits, outUnits) {
+		t.Fatalf("Failed to get a sorted list of units from list-unit-files")
 	}
 }
--- a/registry/job.go
+++ b/registry/job.go
@@ -90,6 +90,7 @@ func (r *EtcdRegistry) Schedule() ([]job.ScheduledUnit, error) {
 func (r *EtcdRegistry) Units() ([]job.Unit, error) {
 	key := r.prefixed(jobPrefix)
 	opts := &etcd.GetOptions{
+		// We need Job Units to be sorted
 		Sort:      true,
 		Recursive: true,
 	}
@@ -117,7 +118,7 @@ func (r *EtcdRegistry) Units() ([]job.Unit, error) {
 		return unit
 	}
 
-	uMap := make(map[string]*job.Unit)
+	units := make([]job.Unit, 0)
 	for _, dir := range res.Node.Nodes {
 		u, err := r.dirToUnit(dir, unitHashLookupFunc)
 		if err != nil {
@@ -127,18 +128,8 @@ func (r *EtcdRegistry) Units() ([]job.Unit, error) {
 		if u == nil {
 			continue
 		}
-		uMap[u.Name] = u
-	}
 
-	var sortable sort.StringSlice
-	for name, _ := range uMap {
-		sortable = append(sortable, name)
-	}
-	sortable.Sort()
-
-	units := make([]job.Unit, 0, len(sortable))
-	for _, name := range sortable {
-		units = append(units, *uMap[name])
+		units = append(units, *u)
 	}
 
 	return units, nil


tixxdz:
@jonboulle if it's green! and no one objects then it's probably ok to merge it.

One point: I made it into 2 patches since the second one will remove the sorting logic completely, and lets only etcd do it. Maybe there are corner cases of both etcd and fleet sorting algorithms which I didn't  check, if you think it's worth it! I'll do it later.

Btw I was going also to add goroutines there so we fetch job keys and unit keys at the same time, but didn't... Not sure it may add load later on etcd if you extra load or queries...

Thank you!

antrik:
@tixxdz I wonder, have you checked whether the first patch alone actually has any measurable performance impact? I have a hard time believing that just reshuffling the loops like that would have a significant effect...

Avoiding the sort OTOH is a good catch :-)

registry/job.go
```diff
@@ -117,7 +118,9 @@ func (r *EtcdRegistry) Units() ([]job.Unit, error) {
 		return unit
 	}
 
-	uMap := make(map[string]*job.Unit)
+	// make it at least size of Nodes
+	uMap := make(map[string]bool, len(res.Node.Nodes))
+	units := make([]job.Unit, 0)
```
antrik: As you are giving the map an initial capacity, why not the slice as well?...

registry/job.go
```diff
@@ -127,18 +130,14 @@ func (r *EtcdRegistry) Units() ([]job.Unit, error) {
...
-	var sortable sort.StringSlice
-	for name, _ := range uMap {
-		sortable = append(sortable, name)
-	}
-	sortable.Sort()
+		// Just a guard
+		_, ok := uMap[u.Name]
+		if ok {
+			continue
+		}
```
antrik: Note that this slightly changes behaviour: previously, if there was a duplicate name, the last one would win; now it's the first...

I wonder though whether this is something that can actually happen at all? Maybe we can just drop this checking altogether.

Even if the check actually *is* necessary, we don't really need the map either: as we know the input is already sorted by name, we just need to compare each entry against the previous one.

jonboulle: Right, I don't understand - what are we guarding against? in what situation do we get dupes?

jonboulle:
> We tested this with 500 services on 3 nodes where previously it would take the fleetctl client to get the list of units from fleet and etcd ~4secs to now ~0.9secs.

Can you show some numbers or a little more detail? I find it hard to see how a string sort is taking 500 seconds... what else has changed?

registry/job.go
tixxdz: Actually as stated in the commit log, we don't really get dupes but even if we did the old code was filtering them inside a hash, thus we always end up with only one unit name in the final slice returned to the caller., but now theoretically with this change and without that guard we may end up with two same names in the returned slice since the append to slice was moved... and as stated in the commit log and commented this is just a guard, I can remove it.

registry/job.go
tixxdz: the map is local collected, the slice is not.

tixxdz:
> Can you show some numbers or a little more detail? I find it hard to see how a 500-string sort is taking 3 seconds... what else has changed?

Actually you are right, we tested this with anton, but with the reverted patch of this PR: https://github.com/coreos/fleet/pull/1512 which was the one taking all the RPC+I/O time.

Anyway it's pretty same results unless reconcile is triggered then at that time the new version will be discharged from those extra loops.

Since this is a simple patch just timing fleetctl when reconcile is triggered, no need to check fleetd in depth.

Old one:
real    0m1.621s
user    0m0.026s
sys     0m0.013s

real    0m0.721s
user    0m0.021s
sys     0m0.012s

New one:
real    0m1.155s
user    0m0.024s
sys     0m0.028s

real    0m0.665s
user    0m0.025s
sys     0m0.012s

tixxdz:
@jonboulle @antrik patch made simple, if no objections it will be pushed after 0.12, thanks!

jonboulle:
LGTM but do we already have a test covering this?

tixxdz:
> LGTM but do we already have a test covering this?

How many units we get back is already covered by scheduling_tests that uses `fleetctl list-unit-files` to check the matching number, now for the order I'm not sure that we have a cover test, we assume that etcd returns it ordered. Will write a quick one for it then. Thanks for the comment.

tixxdz:
Update: the functional test of this will depend on the new functionality that will be introduced in https://github.com/coreos/fleet/pull/1544 to get the units with list-unit-files and make sure that it's sorted. I can write it my self but that will be just a duplicate where we are already adding the appropriate helpers. Thanks!

tixxdz:
To be more precise on this one: functional test should create several units from a template then wait for them to show up, get them sorted and compare.

functional/unit_action_test.go
```diff
@@ -404,3 +407,63 @@ func TestUnitStatus(t *testing.T) {
...
+
+	_, err = cluster.WaitForNMachines(m, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Combine units
+	var units []string
+	for i := 1; i <= 20; i++ {
+		unit := fmt.Sprintf("fixtures/units/hello@%d.service", i)
```
dongsupark: nit: as the number of units needs 2 digits, ``hello@%02d.service`` would be ideal.

functional/unit_action_test.go
```diff
@@ -404,3 +407,63 @@ func TestUnitStatus(t *testing.T) {
...
+	stdout, _, err := cluster.Fleetctl(m, "list-unit-files", "--no-legend", "--fields", "unit")
+	if err != nil {
+		t.Fatal("Failed to run list-unit-files: %v", err)
+	}
+
+	outUnits := strings.Split(strings.TrimSpace(stdout), "\n")
+
+	var sortable sort.StringSlice
+	for _, name := range units {
+		n := strings.Replace(name, "fixtures/units/", "", 1)
```
dongsupark: alternative is ``"path.Base(name)"``

functional/unit_action_test.go
```diff
@@ -404,3 +407,63 @@ func TestUnitStatus(t *testing.T) {
...
+		units = append(units, unit)
+	}
+
+	// make sure that all unit files will show up
+	_, err = cluster.WaitForNUnitFiles(m, 20)
+	if err != nil {
+		t.Fatal("Failed to run list-unit-files: %v", err)
+	}
+
+	stdout, _, err := cluster.Fleetctl(m, "list-unit-files", "--no-legend", "--fields", "unit")
```
dongsupark: Actually I had thought that it would be redundant to call ``"fleetctl list-unit-files"`` again. But after looking into the cluster operation deeper, I can confirm that actually ``cluster.WaitForNUnitFiles()`` does not return a sorted map. (Hmm, not exactly what I expected - scratching my head.) Of course we could make it return a sorted map, if we would really want it. But for now I think it's fine to simply call ``"fleetctl list-unit-files"`` again like this, as this functional test is not in a critical hot-path.
So LGTM.

tixxdz: Indeed. Thanks!

functional/unit_action_test.go
tixxdz: done

functional/unit_action_test.go
tixxdz: done.

functional/unit_action_test.go
```diff
@@ -404,3 +407,63 @@ func TestUnitStatus(t *testing.T) {
...
+	outUnits := strings.Split(strings.TrimSpace(stdout), "\n")
+
+	var sortable sort.StringSlice
+	for _, name := range units {
+		n := path.Base(name)
+		sortable = append(sortable, n)
+	}
+	sortable.Sort()
+
+	var inUnits []string
```
jonboulle: this is weird, why do you create another slice?

tixxdz: for this just to reflect how units are sorted and where sorted! since inUnits are sent using a counter which does not reflect how units are sorted and returned from etcd and golang.

tixxdz: and to follow up with DeepEqual()

tixxdz: Basically I also moved the logic from the Units() fleetd side to put it here in these lines of the functional test to ensure that Units() returns the same result without these lines....

jonboulle: I'm not quite sure I follow. But I meant, you can just go directly to a slice: http://play.golang.org/p/xhsmv-XHSK

tixxdz: I see ;-D, thank you next time fore sure ;-)

Issue #1190, opened 2015-04-16
Fleet loses socket after SIGHUP


akaspin:
I'm trying to change fleet metadata without restarting fleet daemon under CoreOS. OK (fleet v0.10.0):
```
core-01 core # systemctl kill -s SIGHUP fleet
core-01 core # journalctl -fu fleet
-- Logs begin at Thu 2015-04-16 02:29:34 UTC. --
Apr 16 02:35:37 core-01 fleetd[1192]: INFO fleet.go:87: Reloading configuration from
Apr 16 02:35:37 core-01 fleetd[1192]: INFO fleet.go:157: Using default config file /etc/fleet/fleet.conf
Apr 16 02:35:37 core-01 fleetd[1192]: INFO fleet.go:94: Restarting server components
Apr 16 02:35:37 core-01 fleetd[1192]: INFO client.go:292: Failed getting response from http://127.0.0.1:4001/: cancelled
Apr 16 02:35:37 core-01 fleetd[1192]: ERROR event.go:105: etcd watcher {Watch /_coreos.com/fleet/job} returned error: cancelled
Apr 16 02:35:37 core-01 fleetd[1192]: INFO client.go:292: Failed getting response from http://127.0.0.1:4001/: cancelled
Apr 16 02:35:37 core-01 fleetd[1192]: ERROR event.go:105: etcd watcher {Watch /_coreos.com/fleet/job} returned error: cancelled
Apr 16 02:35:37 core-01 fleetd[1192]: INFO server.go:152: Establishing etcd connectivity
Apr 16 02:35:37 core-01 fleetd[1192]: ERROR server.go:51: Failed serving HTTP on listener: @
...
core-01 core # /opt/bin/fleetctl list-units
Error retrieving list of units from repository: googleapi: Error 503: fleet server currently unavailable
```
Any ideas?

adomenech73:
Are you sure you don't have to renew your discovery token for the cluster? I've seen this message in that situation

akaspin:
@adomenech73 No. I just want to reload fleetd instead restart.
For now we using own solution for sync fleetd metadata with EC2 metadata (instance tags, EIPs e.t.c.). But I need to restart fleetd on change. It's pain.

bcwaldon:
@akaspin 
1. your specific use-case of needing to reload to change metadata will be addressed by #1077 
2. Losing the socket is definitely a bug we need to triage further

dalbani:
[...]

PR #1517, opened 2016-03-23
fleetd: keep a list of listeners across reconfiguring fleetd

dongsupark:
So far fleetd has lost its existing listeners after receiving SIGHUP.
That happened because ``reconfigure()`` simply killed the server and started
a fresh one, without taking consideration into the existing listeners.
That resulted in a strange error like:

>  Failed serving HTTP on listener: @ accept unix @: accept: invalid argument

So let's keep the list of existing listeners before killing the server,
and give it to the new server.

Fixes: https://github.com/coreos/fleet/issues/1190

/cc @kayrus @dalbani

### DIFF
--- a/functional/server_test.go
+++ b/functional/server_test.go
@@ -15,10 +15,12 @@
 package functional
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 
 	"github.com/coreos/fleet/functional/platform"
+	"github.com/coreos/fleet/functional/util"
 )
 
 // TestReconfigureServer checks whether fleetd managed to keep its listeners
@@ -39,30 +41,37 @@ func TestReconfigureServer(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// Check expected state before reconfiguring fleetd
-	stdout, _ := cluster.MemberCommand(m0, "systemctl", "show", "--property=ActiveState", "fleet")
-	if strings.TrimSpace(stdout) != "ActiveState=active" {
-		t.Fatalf("Fleet unit not reported as active: %s", stdout)
-	}
-	stdout, _ = cluster.MemberCommand(m0, "systemctl", "show", "--property=Result", "fleet")
-	if strings.TrimSpace(stdout) != "Result=success" {
-		t.Fatalf("Result for fleet unit not reported as success: %s", stdout)
-	}
-
 	// NOTE: we need to sleep once here to get reliable results.
 	// Without this sleep, the entire fleetd test always ends up succeeding
 	// no matter whether SIGHUP came or not.
 	_, _ = cluster.MemberCommand(m0, "sh", "-c", `'sleep 2'`)
 
-	stdout, _ = cluster.MemberCommand(m0, "sudo", "kill", "-HUP", "$(pidof fleetd)")
+	err = waitForFleetdSocket(cluster, m0)
+	if err != nil {
+		t.Fatalf("Failed to get a list of fleetd sockets: %v", err)
+	}
+
+	// send a SIGHUP to fleetd, and periodically checks if a message
+	// "Reloading configuration" appears in fleet's journal, up to timeout (15) seconds.
+	stdout, _ := cluster.MemberCommand(m0, "sudo", "systemctl", "kill", "-s", "SIGHUP", "fleet")
 	if strings.TrimSpace(stdout) != "" {
 		t.Fatalf("Sending SIGHUP to fleetd returned: %s", stdout)
 	}
 
+	err = waitForReloadConfig(cluster, m0)
+	if err != nil {
+		t.Fatalf("Failed to get log about reconfiguration: %v", err)
+	}
+
+	// check if fleetd is still running correctly, by running fleetctl status
+	_, _, err = cluster.Fleetctl(m0, "list-units")
+	if err != nil {
+		t.Fatalf("Unable to check list-units. Please check for fleetd socket. err:%v", err)
+	}
+
 	// Check for HTTP listener error looking into the fleetd journal
-	stdout, _ = cluster.MemberCommand(m0, "sh", "-c",
-		`'journalctl -u fleet | grep "Failed serving HTTP on listener:"'`)
-	if strings.TrimSpace(stdout) != "" {
+	stdout, _ = cluster.MemberCommand(m0, "journalctl _PID=$(pidof fleetd)")
+	if strings.Contains(strings.TrimSpace(stdout), "Failed serving HTTP on listener:") {
 		t.Fatalf("Fleetd log returned error on HTTP listeners: %s", stdout)
 	}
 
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -115,9 +116,17 @@ func New(cfg config.Config) (*Server, error) {
...
+	if oldListeners == nil || len(oldListeners) == 0 {
+		listeners, err = activation.Listeners(false)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		listeners = make([]net.Listener, len(oldListeners))
+		for i, l := range oldListeners {
+			listeners[i] = l
+		}
```
antrik: Why are you copying the values one by one: can't you just assign `listeners = oldListeners`?...

(This will just copy the reference rather than the actual values -- but I don't see why this would be undesirable?)

antrik:
Looks good aside from the one patch comment. I'd say this should come with a testcase, though?

server/server.go
dongsupark: @antrik 
> can't you just assign listeners = oldListeners?

Correct. Done.

dongsupark:
@antrik 
> I'd say this should come with a testcase, though?

Done. I added a new functional test ``TestReconfigureServer``.

server/server.go
```diff
@@ -115,9 +116,14 @@ func New(cfg config.Config) (*Server, error) {
 
 	e := engine.New(reg, lManager, rStream, mach)
 
-	listeners, err := activation.Listeners(false)
-	if err != nil {
-		return nil, err
+	var listeners []net.Listener
```
jonboulle: A little simpler
`func New(cfg config.Config, listeners []net.Listener) (*Server, error) {`
```
if len(listeners) == 0 {
    listeners, err = activation.Listeners ...
}
```

jonboulle: == nil check is unnecessary, len handles it for you

dongsupark: @jonboulle It's done. Thanks for the suggestion.

jonboulle:
Almost LGTM but perhaps consider https://github.com/coreos/fleet/pull/1528/files ?

dongsupark:
@jonboulle Done. I dropped unnecessary sudo calls, except for ``sudo kill -HUP``.

functional/server_test.go
```diff
@@ -0,0 +1,78 @@
...
+	stdout, _ = cluster.MemberCommand(m0, "systemctl", "show", "--property=Result", "fleet")
+	if strings.TrimSpace(stdout) != "Result=success" {
+		t.Fatalf("Result for fleet unit not reported as success: %s", stdout)
+	}
+
+	// NOTE: we need to sleep once here to get reliable results.
+	// Without this sleep, the entire fleetd test always ends up succeeding
+	// no matter whether SIGHUP came or not.
+	_, _ = cluster.MemberCommand(m0, "sh", "-c", `'sleep 2'`)
+
```
tixxdz: Hmm, would it help after SIGHUP is sent to check the journal if "fleet.*Reloading configuration.*" is really there ?

functional/server_test.go
```diff
@@ -0,0 +1,78 @@
...
+		t.Fatalf("Sending SIGHUP to fleetd returned: %s", stdout)
+	}
+
+	// Check for HTTP listener error looking into the fleetd journal
+	stdout, _ = cluster.MemberCommand(m0, "sh", "-c",
+		`'journalctl -u fleet | grep "Failed serving HTTP on listener:"'`)
+	if strings.TrimSpace(stdout) != "" {
+		t.Fatalf("Fleetd log returned error on HTTP listeners: %s", stdout)
+	}
+
```
tixxdz: The check seems fine, but what worries me is we are checking an error message that might disappear for other reasons! is there a better way to check for sockets in: systemctl status fleetd; grep pid, then inspect /proc/pid/* ?

tixxdz:
I just commented, if there are no easy ways! then don't bother, lgtm.

Thank you!

functional/server_test.go
dongsupark: Yeah, I was just lazy when I wrote this part. I'll try to change it like that.

functional/server_test.go
antrik: AIUI the problem is actually that we need to make sure fleetd is fully initialised before even sending the SIGHUP. I agree though that it would be better to wait for this by checking the logs. (Unless there is some more elegant method for checking fleetd's exact status?...)

Note that a proper check like that will also make the preceding checks using systemctl unnecessary.

functional/server_test.go
antrik: Trying to check sockets sounds complicated, and I'm not convinced it's really a good idea to rely on internals like that...

TBH, I don't think *either* of these approaches is really the right way to test this. Rather, we should try to check whether actual socket connections keep working as expected. I don't have any specific setup in mind for this -- but surely it should be possible to do this?

tixxdz: @antrik indeed, if it's easy to check if the socket is still working in this context then go for it ? my comment was a bit generic...

jonboulle: > Rather, we should try to check whether actual socket connections keep working as expected.

+1, test for the desired behaviour by testing the desired behaviour :-)

functional/server_test.go
dongsupark: @tixxdz 
> Hmm, would it help after SIGHUP is sent to check the journal if "fleet.Reloading configuration." is really there ?

Yes, I updated it like that.
But that change didn't allow me to remove the sleep command before sending SIGHUP. Without the sleep, test always ends up succeeding, no matter what.. :-(

functional/server_test.go
dongsupark: > we should try to check whether actual socket connections keep working as expected

Done.

functional/server_test.go
antrik: @dongsupark actually, I just realised that `WaitForNMachines()` will only succeed once fleet is fully up? So the `systemctl` checks should definitely be redundant.

Also, I still don't understand why the sleep is necessary... But I guess that's not a showstopper.

functional/server_test.go
```diff
@@ -0,0 +1,137 @@
...
+		},
+	)
+	if err != nil {
+		t.Fatalf("Fleetd reconfiguration log not found: %v", err)
+	}
+
+	err = waitForFleetdSocket(cluster, m0)
+	if err != nil {
+		t.Fatalf("Failed to get a list of fleetd sockets: %v", err)
+	}
```
antrik: Do we still need this? I would assume it's no longer necessary, now that we actually check for the socket connection still working?...

dongsupark:
Updated.
* remove unnecessary checks for systemctl show in the beginning
* remove unnecessary checks for fleetd socket after sending the signal
* split out ``waitForReloadConfig`` for better readability

functional/server_test.go
antrik: I see that you use it here now to check for fleet being up, which makes sense I guess. With this in place, do you still need the `sleep` above?...

dongsupark: > I see that you use it here now to check for fleet being up, which makes sense I guess. With this in place, do you still need the sleep above?...

Unfortunately yes, I still need the sleep. I just did another round of tests, but it's still the same.
As mentioned above, without the sleep, the whole test always succeeds. What would be the point testing this at all, if it would always give the same answer? That's where I'm puzzled, but for me it's more important to get the test working at all.

functional/server_test.go
```diff
@@ -0,0 +1,133 @@
...
+func waitForReloadConfig(cluster platform.Cluster, m0 platform.Member) (err error) {
+	_, err = util.WaitForState(
+		func() bool {
+			// NOTE: journalctl should run just simply like "journalctl -u fleet",
+			// without being piped with grep. Doing
+			// "journalctl -u fleet | grep \"Reloading configuration\"" is racy
+			// in a subtle way, so that it sometimes fails only on semaphoreci.
+			// - dpark 20160408
+			stdout, _ := cluster.MemberCommand(m0, "journalctl -u fleet")
+			journalfleet := strings.TrimSpace(stdout)
```
tixxdz: @dongsupark does it make a difference here if you do "journalctl _PID=$(pidof fleetd)" ?

functional/server_test.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	err = waitForReloadConfig(cluster, m0)
+	if err != nil {
+		t.Fatalf("Failed to get log about reconfiguration: %v", err)
+	}
+
+	// check if fleetd is still running correctly, by running fleetctl status
+	_, _, err = cluster.Fleetctl(m0, "status")
+	if err != nil {
```
tixxdz: @dongsu: could you please change this to:
`fleet $ ./bin/fleetctl list-units; echo $?
Error retrieving list of units from repository: googleapi: Error 503: fleet server unable to communicate with etcd
1
`
`fleet $ ./bin/fleetctl status; echo $?
0
`
That's my result of experimenting right now, at least it is the same as in the original reported bug

Thanks you!

dongsupark: @tixxdz this reveals that there's still a bug somewhere. I'll try to investigate soon.

dongsupark: @tixxdz Fixed a bug. Checking with "fleetctl list-units" is now working.

functional/server_test.go
antrik: I don't see why `pidof` would be more reliable than `-u`? Do we expect more than one instance of fleetd to exist during the test? (And if so, is it actually appropriate to consider only the currently active one?...)

(I don't see why using grep wouldn't be reliable in the first place, though... But doing the compare in Go is fine too; so I guess this is a non-issue.)

tixxdz: What are the gains of using `-u` where we just want what the pid logs which is more precise in this context ?

tixxdz: and yeh I forget to say that some developers are also just lazy so they run the tests again and again on the same VM grepping the old bogus journal logs...

tixxdz:
Lgtm, and since this is fixing a bug that most users want I'm merging this series. If any one have a problem with that minor sleep in the functional test feel free to open an issue or a PR. Sure it helps to serialize things when the container and all the services are started and where the logs and extra operation may have some delays...

Thanks!

functional/server_test.go
antrik: @tixxdz there are no "old bogus journal logs" -- each test spawns fresh containers with fresh logs. And I still think the `-u` abstraction offered to us by `journalctl` is more elegant than dealing with low-level details such as PIDs ourselves...

(PIDs are in fact prone to race conditions in general -- though I guess that's not really an issue in this specific case.)

PR #1520, opened 2016-03-29
*: fixed broken template and metadata

### LINKED ISSUE #1514: *: rework "consolidate excessive Hash/Unit requests" PR
We have to rework #1376 PR.

Relates to: #1512, #1446

kayrus:
Resolves #1514

The problem was caused by the [code optimization](https://github.com/coreos/fleet/pull/1376). Before that each unit was stored in its own variable. Then this code was optimized and units became stored in hash map (`getAllUnitsHashMap`). Each hash was assigned to the unit's pointer. And when template unit was checked by `requirements()` function, its content was modified by `values[i] = unitPrintf(v, *uni)` code. Once templated unit was modified, all related units (which have same hash) were modified too, because they are related to one pointer.

Now we don't modify source contents.

### DIFF
--- a/job/job.go
+++ b/job/job.go
@@ -175,11 +175,14 @@ func (j *Job) requirements() map[string][]string {
 		}
 
 		if uni != nil {
+			processedValues := make([]string, len(values))
 			for i, v := range values {
-				values[i] = unitPrintf(v, *uni)
+				processedValues[i] = unitPrintf(v, *uni)
 			}
+			requirements[key] = processedValues
+		} else {
+			requirements[key] = values
 		}
-		requirements[key] = values
 	}
 
 	return requirements
--- a/job/job_test.go
+++ b/job/job_test.go
@@ -195,6 +195,11 @@ Zzz=something
 		}
 	}
 
+	jOrig := NewJob("ssh@2.service", *newUnit(t, contents))
+	// Now ensure that Contents was not modified
+	if !reflect.DeepEqual(jOrig.Unit.Contents, j.Unit.Contents) {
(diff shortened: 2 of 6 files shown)


jonboulle:
Great catch, and gross. `requirements()` shouldn't have any side effects. Can you add a unit test for that?

kayrus:
Added test and it has failed. Now, let's apply the fix...

kayrus:
@jonboulle done

job/job.go
```diff
@@ -174,12 +174,17 @@ func (j *Job) requirements() map[string][]string {
 			requirements[key] = make([]string, 0)
 		}
 
-		if uni != nil {
-			for i, v := range values {
-				values[i] = unitPrintf(v, *uni)
+		processedValues := make([]string, len(values))
+
+		for i, v := range values {
```
tixxdz: @kayrus thanks for the catch! you can put this inside the same previous loop:
`if uni != nil { processedvalues := make()
     for ... { ...}
     requirements[key] = processedvalues
} else { requirements[key] = values } `

Otherwise it's ok too!

tixxdz:
@kayrus perhaps also improve the commit log of the 4th patch (the fix), same wording as for this PR ? otherwise lgtm. Thank you!

job/job.go
kayrus: Fixed

tixxdz:
@jonboulle lgtm, thanks!

jonboulle:
LGTM, thanks - but how about reordering a23ce6f after or before 	de4ba7f ?

kayrus:
@jonboulle I suppose we have to show that the issue is covered by tests, and then fix it. so from my point of view the commit order is fine

PR #1521, opened 2016-03-29
godep: updated golang.org/x/crypto/...

### LINKED ISSUE #565: handshake failed: ssh: unexpected message type 1 (expected 52)
I reported this on coreos/coreos-vagrant#106 but it also happens when trying to manage a cluster on EC2.

It happens when you have more than 7 keys in the ssh-agent. 

My ~/.ssh/config says what key should be used for that host but the ssh tunneling feature seems to ignore that.

kayrus:
resolves #565


tixxdz:
@kayrus @jonboulle it seems that of the library tests (not related to fleet) were removed, where in upstream they are still there. Any hint on this ?

thanks!

kayrus:
I suppose it is godep's behavior.

kayrus:
I use go1.5.3 and godep v60. Even with `-t` godep doesn't preserve test files.

jonboulle:
Not including tests is fine. I'm surprised `-t` doesn't work for you, sounds like a bug..

jonboulle:
Thanks @kayrus ! LGTM

kayrus:
@jonboulle yep, it is a bug https://github.com/tools/godep/issues/405

Issue #464, opened 2014-05-15
Unit does not automatically start Required units


dknell:
I would like to be able to start a single unit and have another required unit start automatically.

This may not be supported by fleet or I'm misunderstanding the systemd documentation, but from what I'm reading in the [systemd.unit documentation](http://www.freedesktop.org/software/systemd/man/systemd.unit.html#Requires=) I should be able to start a unit and it will automatically start any unit that is listed in the `Requires`. Is this possible or am I doing something wrong?

This is a simple test:

busy.service

```
[Unit]
Description=BusyBox Test
After=docker.service
Requires=docker.service
Requires=busy-watch.service

[Service]
ExecStart=/usr/bin/docker run busybox /bin/sh -c "while true; do echo Hello World; sleep 20; done"
```

busy-watch.service

```
[Unit]
Description=BusyBox Test Watcher
BindsTo=busy.service

[Service]
ExecStart=/bin/sh -c "while true; do etcdctl set /tasks/busy/running 'yes' --ttl 60;sleep 45;done"
ExecStop=/usr/bin/etcdctl rm /tasks/busy/running

[X-Fleet]
X-ConditionMachineOf=busy.service
```

Add the units to fleet:

```
fleetctl submit busy*.service
```

Units are available to run:

```
core@core-01 ~ $ fleetctl list-units
UNIT            LOAD    ACTIVE  SUB DESC            MACHINE
busy-watch.service  -   -   -   BusyBox Test Watcher    -
busy.service        -   -   -   BusyBox Test        -
```

By starting busy.service, I would expect busy-watch.service to also start:

```
core@core-01 ~ $ fleetctl start busy.service
Job busy.service scheduled to 4621f512.../172.17.8.102
```

Neither units are running:

```
core@core-01 ~ $ fleetctl list-units
UNIT            LOAD    ACTIVE  SUB DESC            MACHINE
busy-watch.service  -   -   -   BusyBox Test Watcher    -
busy.service        -   -   -   BusyBox Test        -
```

Just to make sure that the units are functional and that there are no syntax errors, I tried removing `Requires=busy-watch.service` and start both of the units together.

```
core@core-01 ~ $ fleetctl start busy.service busy-watch.service
Job busy.service scheduled to 4621f512.../172.17.8.102
Job busy-watch.service scheduled to 4621f512.../172.17.8.102
core@core-01 ~ $ fleetctl list-units
UNIT            LOAD    ACTIVE  SUB DESC            MACHINE
[...]

PR #1523, opened 2016-03-30
fleetd: process dependencies in [Install] section

### LINKED ISSUE #1382: WantedBy doesn't lead to starting service
Hey,

I have 2 services `es` and `es-discovery` ... `es` needs `es-discovery`. If `es` declares `Wants=es.discovery.service` everything works as expected and when I start `es` then `es-discovery` is started too. But if I declare `WantedBy=es` in `es-discovery` service then units are created properly (es unit Wants es-discovery) but `es-discovery` never starts. This is the result of loading both but starting just `es` : http://pastebin.com/raw.php?i=kmATHbGD

`es@.service` : 

```
[Unit]
Description=Elasticsearch Service
Wants=es-data@%i.service log-entries@%i.service
After=es-data@%i.service log-entries@%i.service

[Service]
TimeoutStartSec=0
Restart=always
RestartSec=30
EnvironmentFile=/etc/environment
TimeoutStopSec=30

ExecStartPre=-/usr/bin/docker kill %p-%i
ExecStartPre=-/usr/bin/docker rm %p-%i
ExecStartPre=-/usr/bin/docker pull gwiq/es-prod

ExecStart=/bin/bash -c '\
  curl -f ${COREOS_PRIVATE_IPV4}:4001/v2/keys/services/es; \
  if [ "$?" -eq 0 ]; then \
      UNICAST_HOSTS=$(etcdctl ls --recursive /services/es \
                      | sed "s/\/services\/es\///g" \
                      | sed "s/$/:9300/" \
                      | paste -s -d","); \
  else \
      UNICAST_HOSTS=""; \
  fi; \
  /usr/bin/docker run \
    --name %p-%i \
    --publish 9200:9200 \
    --publish 9300:9300 \
    --volumes-from es-data-%i \
    -e ES_DISCOVERY_UNICAST_HOSTS=$UNICAST_HOSTS \
    -e ES_NETWORK_PUBLISH_HOST=$COREOS_PRIVATE_IPV4 \
    -e ES_NODE_NAME=%p-%i \
    --env-file /opt/coreos-services/project/es/env \
    gwiq/es-prod'

ExecStop=/bin/bash -c '\
    curl -XPOST -fSs ${COREOS_PRIVATE_IPV4}:9200/_cluster/nodes/_local/_shutdown > /dev/null; \
    sleep 10;'

ExecStopPost=/usr/bin/docker stop %p-%i
ExecStopPost=-/usr/bin/docker rm %p-%i

[X-Fleet]
Conflicts=%p@*.service
MachineOf=es-data@%i.service
```

`es-discovery@.service`

```
[Unit]
Description=Elasticsearch Discovery Service
Requires=etcd2.service cluster-init.service
After=etcd2.service cluster-init.service
Before=es@%i.service
BindsTo=es@%i.service

[Service]
Restart=always
EnvironmentFile=/etc/environment

ExecStart=/bin/bash -c '\
  while true; do \
    etcdctl set /services/es/${COREOS_PRIVATE_IPV4} \'{"http_port": 9200, "transport_port": 9300}\' --ttl 30; \
    sleep 20; \
  done'

ExecStop=/usr/bin/etcdctl rm /services/es/${COREOS_PRIVATE_IPV4}

[Install]
WantedBy=es@%i.service

[X-Fleet]
MachineOf=es@%i.service
```

If I just move `WantedBy` from `es-discovery` to `Wants` in `es` then it works... But I need `WantedBy` because `es` service shouldn't know about the discovery...

If I start the service explicitly afterwards, it starts...

I suspect that the problem might be that the ExecStart process blocks because there is never ending while loop.

kayrus:
resolves #1382

### DIFF
--- a/Documentation/unit-files-and-scheduling.md
+++ b/Documentation/unit-files-and-scheduling.md
@@ -57,6 +57,35 @@ To use instance units, simply create a unit file whose name matches the `<name>@
 
 When working with instance units, it is strongly recommended that all units be _entirely homogenous_. This means that any unit created as, say, `foo@1.service`, should be created only from the unit named `foo@.service`. This homogeneity will be enforced by the fleet API in future.
 
+## Definition of the Install Section
+
+Unit files which have an `[Install]` section will be automatically enabled by fleet. This means that the states of such unit files cannot be tracked by fleet. For example, assume we have loaded this `my.service` unit file:
+
+```ini
+[Service]
+ExecStart=/bin/bash -c "while true; do echo my.service unit file; sleep 1; done"
+```
+
+and then loaded an additional [sidekick][sidekick] discovery unit `my_discovery.service`:
+
+```ini
+[Unit]
+BindsTo=my.service
+
+[Service]
+ExecStart=/bin/bash -c "while true; do echo my_discovery.service unit file; sleep 1; done"
+
+[Install]
+WantedBy=my.service
+
+[X-Fleet]
+MachineOf=my.service
+```
+
+fleet will load and enable the `my_discovery.service` unit above because it contains an `[Install]` section. When `my.service` is started, systemd will also start `my_discovery.service`, independent of the desired state defined for `my_discovery.service` in fleet. This can cause confusion between the output of `fleetctl list-units` and `systemctl list-units`, which will not match in this scenario. Use `fleetctl status my_discovery.service` to explicitly identify the service and get its actual unit status.
+
+If systemd can not enable the unit which has `[Install]` section, fleet will interrupt load process and return an error.
+
 ## systemd specifiers
 
 When evaluating the `[X-Fleet]` section, fleet supports a subset of systemd's [specifiers][systemd specifiers] to perform variable substitution. The following specifiers are currently supported:
@@ -88,7 +117,7 @@ For more details on the specific behavior of the engine, read more about [fleet'
 
 For non-global units, several different directives are available to control the engine's scheduling decision.
 
-##### Schedule unit to specific machine
+## Schedule unit to specific machine
 
 The `MachineID` option of a unit file causes the system to schedule a unit to a machine identified by the option's value.
 
@@ -98,7 +127,7 @@ One must use the entire ID when setting `MachineID` - the shortened ID returned 
 fleet depends on its host to generate an identifier at `/etc/machine-id`, which is handled today by systemd.
 Read more about machine IDs in the [official systemd documentation][machine-id].
 
-##### Schedule unit to machine with specific metadata
+## Schedule unit to machine with specific metadata
 
 The `MachineMetadata` option of a unit file allows you to set conditional metadata required for a machine to be elegible.
 
@@ -181,7 +210,7 @@ app.service     fd1d3e94.../10.0.0.1    active  running
 A machine is not automatically configured with metadata.
 A deployer may define machine metadata using the `metadata` [config option][config-option].
 
-##### Schedule unit next to another unit
+## Schedule unit next to another unit
 
 In order for a unit to be scheduled to the same machine as another unit, a unit file can define `MachineOf`.
 The value of this option is the exact name of another unit in the system, which we'll call the target unit.
@@ -193,13 +222,13 @@ Follower units will reschedule themselves around the cluster to ensure their `Ma
 
 Note that currently `MachineOf` _cannot_ be a bidirectional dependency: i.e., if unit `foo.service` has `MachineOf=bar.service`, then `bar.service` must not have a `MachineOf=foo.service`, or fleet will be unable to schedule the units.
 
-##### Schedule unit away from other unit(s)
+## Schedule unit away from other unit(s)
 
 The value of the `Conflicts` option is a [glob pattern][glob-pattern] defining which other units next to which a given unit must not be scheduled. A unit may have multiple `Conflicts` options.
 
 If a unit is scheduled to the system without an `Conflicts` option, other units' conflicts still take effect and prevent the new unit from being scheduled to machines where conflicts exist.
 
-##### Dynamic requirements
+## Dynamic requirements
 
 fleet supports several [systemd specifiers][systemd-specifiers] to allow requirements to be dynamically determined based on a Unit's name. This means that the same unit can be used for multiple Units and the requirements are dynamically substituted when the Unit is scheduled.
 
@@ -221,4 +250,5 @@ would result in an effective `MachineOf` of `foo.socket`. Using the same unit sn
 [glob-pattern]: http://golang.org/pkg/path/#Match
 [unit-scheduling]: #unit-scheduling
 [example-deployment]: examples/example-deployment.md#service-files
+[sidekick]: examples/service-discovery.md
 [systemd-specifiers]: #systemd-specifiers
--- a/functional/install_test.go
+++ b/functional/install_test.go
@@ -1,0 +1,89 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package functional
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/coreos/fleet/functional/platform"
+	"github.com/coreos/fleet/functional/util"
+)
+
+// Load service and discovery units and test whether discovery unit adds itself as a dependency for the service.
+func TestInstallUnit(t *testing.T) {
+	cluster, err := platform.NewNspawnCluster("smoke")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cluster.Destroy()
+
+	// Start with a two-nodes cluster
+	members, err := platform.CreateNClusterMembers(cluster, 2)
+	if err != nil {
+		t.Fatal(err)
+	}
+	m0 := members[0]
+	_, err = cluster.WaitForNMachines(m0, 2)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Load unit files
+	stdout, stderr, err := cluster.Fleetctl(m0, "load", "fixtures/units/hello.service", "fixtures/units/discovery.service")
+	if err != nil {
+		t.Fatalf("Failed loading unit files: \nstdout: %s\nstderr: %s\nerr: %v", stdout, stderr, err)
+	}
+
+	checkState := func(match string) bool {
+		stdout, _, err := cluster.Fleetctl(m0, "--strict-host-key-checking=false", "ssh", "discovery.service", "systemctl show --property=ActiveState discovery.service")
+		if err != nil {
+			t.Error("Failed getting info using remote systemctl")
+		}
+		stdout = strings.TrimSpace(stdout)
+		return stdout == fmt.Sprintf("ActiveState=%s", match)
+	}
+
+	// Verify that discovery.service unit is loaded but not started
+	timeout, err := util.WaitForState(func() bool { return checkState("inactive") })
+	if err != nil {
+		t.Fatalf("discovery.service unit is not reported as inactive within %v: %v", timeout, err)
+	}
+
+	// Start hello.service unit
+	stdout, stderr, err = cluster.Fleetctl(m0, "start", "fixtures/units/hello.service")
+	if err != nil {
+		t.Fatalf("Failed starting unit: \nstdout: %s\nstderr: %s\nerr: %v", stdout, stderr, err)
+	}
+
+	// Verify that discovery.service unit was started
+	timeout, err = util.WaitForState(func() bool { return checkState("active") })
+	if err != nil {
+		t.Fatalf("discovery.service unit is not reported as active within %v:\n%v", timeout, err)
+	}
+
+	// Stop hello.service unit
+	stdout, stderr, err = cluster.Fleetctl(m0, "stop", "fixtures/units/hello.service")
+	if err != nil {
+		t.Fatalf("Failed stopping unit: \nstdout: %s\nstderr: %s\nerr: %v", stdout, stderr, err)
+	}
+
+	// Verify that discovery.service unit was stopped
+	timeout, err = util.WaitForState(func() bool { return checkState("inactive") })
+	if err != nil {
+		t.Fatalf("discovery.service unit is not reported as inactive within %v:\nv", timeout, err)
+	}
+}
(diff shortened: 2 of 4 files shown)


kayrus:
TODO:
* write functional test
* write unit test

systemd/manager.go
```diff
@@ -106,6 +106,13 @@ func (m *systemdUnitManager) Load(name string, u unit.UnitFile) error {
 	if err != nil {
 		return err
 	}
+	if u.HasReverseDependencies() {
```
antrik: I think the idea was to try *always* enabling units on load? This kind of special-casing seems fragile, unnecessary, and confusing...

kayrus: It is not posible to enable unit when it doesn't have `Install` section

antrik: I see. That being so, how about just making it conditional on the presence of this section? It seems a much clearer and less error-prone distinciton...

kayrus: What if this section is empty or it has invalid data?

antrik: Well, you tell me ;-)

I'm assuming if it's empty, systemd will enable it just fine? And if it's invalid, it will surely bomb -- which I'd say is the right thing to do...

kayrus:
Just tested this functionality and noticed few issues. Here is the input data:

`wants@.service`:

```ini
[Service]
ExecStart=/bin/bash -c "while true; do echo wants@%i.service unit file; sleep 1; done"
```

`wanted@.service`:

```ini
[Unit]
Before=wants@%i.service
BindsTo=wants@%i.service

[Service]
ExecStart=/bin/bash -c "while true; do echo wanted@%i.service unit file; sleep 1; done"

[Install]                                                                                                               
WantedBy=wants@%i.service                                                                                               
                                                                                                                        
[X-Fleet]                                                                                                               
MachineOf=wants@%i.service
```

Issue 1

```sh
fleectl submit wanted@.service wants@.service
fleetctl start wants@1.service
```

In this case fleetd doesn't process `WantedBy` of the `wanted@.service` template. And if we would like to start `wanted@1.service` automatically, we have to `load` it first, i.e.:

```sh
fleectl submit wanted@.service wants@.service
fleetctl load wanted@1.service
fleetctl start wants@1.service
```

And here comes Issue 2:

Within this situation fleetd doesn't monitor the `wanted@1.service` status and it still has `loaded` state:

```sh
fleetctl list-unit-files
wanted@.service         2797f9a inactive        inactive        -
wanted@1.service        2797f9a loaded          loaded          6ed03b01.../coreos1
wants@.service          038ee83 inactive        inactive        -
wants@1.service         038ee83 launched        launched        6ed03b01.../coreos1
```

But systemd status is correct:

```sh
systemctl status wanted@1.service                                                                      
● wanted@1.service                                                                                                      
   Loaded: loaded (/run/fleet/units/wanted@1.service; enabled-runtime; vendor preset: disabled)                         
   Active: active (running) since Wed 2016-03-30 16:52:54 UTC; 11min ago                                                
 Main PID: 17765 (bash)                                                                                                 
   CGroup: /system.slice/system-wanted.slice/wanted@1.service                                                           
           ├─17765 /bin/bash -c while true; do echo wanted@1.service unit file; sleep 1; done                           
           └─19108 sleep 1                                                                                              
                                                                                                                        
Mar 30 17:03:50 coreos1 bash[17765]: wanted@1.service unit file                                                         
Mar 30 17:03:51 coreos1 bash[17765]: wanted@1.service unit file                                                         
Mar 30 17:03:52 coreos1 bash[17765]: wanted@1.service unit file
Mar 30 17:03:53 coreos1 bash[17765]: wanted@1.service unit file
Mar 30 17:03:54 coreos1 bash[17765]: wanted@1.service unit file
Mar 30 17:03:55 coreos1 bash[17765]: wanted@1.service unit file
Mar 30 17:03:56 coreos1 bash[17765]: wanted@1.service unit file
Mar 30 17:03:57 coreos1 bash[17765]: wanted@1.service unit file
Mar 30 17:03:58 coreos1 bash[17765]: wanted@1.service unit file
Mar 30 17:03:59 coreos1 bash[17765]: wanted@1.service unit file
```

```sh
systemctl list-dependencies --reverse wants@1.service
wants@1.service
● └─wanted@1.service
```

Should fleet monitor unit status on the systemd side? And what should it do when it will find that desired status doesn't correspond to the real state? Should fleetd change unit's state by itself (no by systemd)?

The answer to the last question could be as follows:

When `wants@.service` unit's state is being changed, fleetd should get all existing units, parse them, and figure out whether there is an additional dependency. If there is `WantedBy` and unit is not loaded - just ignore that. If there is `RequiredBy` and unit is not loaded - we have to load that unit first, and then change `wants@.service` unit's state.

Looks like this issue has lots pitfalls.

kayrus:
@jonboulle any comment?

antrik:
@kayrus I don't think your Issue 1 is actually a problem -- as far as I can see, it's perfectly in line with the conversation in https://github.com/coreos/fleet/issues/1382 .

As for Issue 2... This does sound like a problem -- but isn't it just the same when using "Wants" dependencies instead, which apparently already worked in the existing code base?...

kayrus:
@joshix @robszumski could you review documentation part?

Documentation/unit-files-and-scheduling.md
```diff
@@ -57,6 +57,35 @@ To use instance units, simply create a unit file whose name matches the `<name>@
...
+ExecStart=/bin/bash -c "while true; do echo my_discovery.service unit file; sleep 1; done"
+
+[Install]
+WantedBy=my.service
+
+[X-Fleet]
+MachineOf=my.service
+```
+
+Fleet will load and automatically enable unit above because it contains `[Install]` section. Systemd will create a `my.service.wants/my_discovery.service` symlink to the `my_discovery.service` unit file. And when you start `my.service`, systemd will start `my_discovery.service` independently on `my_discovery.service` desired state defined in fleet. This situation may casue confusion between `fleetctl list-units` and `systemd list-units` outputs. It is recommended to use `fleetctl status my_discovery.service` to get real unit status (`fleetctl status` just passes arguments to `systemctl status` command).
```
jonboulle: Can you elaborate on "this may cause confusion between fleetctl list-units and systemctl list-units"?

jonboulle: It worries me and I'm starting to wonder if this is a bad idea.

robszumski: I think we need to specifically define what "enable" means for the user, ie. starts on boot

kayrus: It doesn't suite fleet behavior. Because fleet stores unit files inside tmpfs and they will not be booted automatically using only systemd routines.

antrik: @jonboulle I agree that this is a problem -- but AIUI it's actually not a *new* problem: according to the discussion in https://github.com/coreos/fleet/issues/1382 , systemd was already handling dependencies expressed by "Wants" relationships even without enabling; it's just "WantedBy" that changes... So unless I'm missing something, enabling units doesn't fundamentally change the situation -- it only makes it manifest in another case. (Which was somewhat inconsistent before.)

AIUI, the issue should be mostly mitigated once fleet implements dependencies such as X-Requires etc. -- though I guess it might be worthwhile to look into the inconsistency problem regardless. But I really think this is a separate issue.

kayrus:
Updated tests. Now semaphoreci works.
I've removed failed golang 1.4.3 test in #1535

functional/install_test.go
```diff
@@ -0,0 +1,117 @@
...
+		if stdout == "ActiveState=inactive" {
+			return true
+		} else {
+			return false
+		}
+	}
+
+	// Verify that discovery.service unit is stopped
+	timeout, err := util.WaitForState(
+		func() bool { return checkInactiveState() },
```
antrik: No need to wrap it in closure like that -- you can just pass `checkInactiveState` directly.

functional/install_test.go
```diff
@@ -0,0 +1,117 @@
...
+	)
+	if err != nil {
+		t.Fatalf("discovery.service unit is not reported as inactive within %v: %v", timeout, err)
```
antrik: Either use the default error message in `err`, *or* print your own error string; but not both -- that would be redundant.

(Admittedly, this is a bit confusing... Which is why I wasn't quite convinced this solution is indeed better than passing a separate callback for the error output.)

systemd/manager.go
```diff
@@ -106,6 +106,14 @@ func (m *systemdUnitManager) Load(name string, u unit.UnitFile) error {
 	if err != nil {
 		return err
 	}
+	if len(u.Contents["Install"]) > 0 {
```
antrik: I wonder, does that skip a present but empty `Install` section? I believe we get least surprise if it *doesn't* skip in this case. (Unless systemd has a problem with that?...)

Either way, it would be good to have a test for this case.

Documentation/unit-files-and-scheduling.md
```diff
@@ -57,6 +57,35 @@ To use instance units, simply create a unit file whose name matches the `<name>@
 
 When working with instance units, it is strongly recommended that all units be _entirely homogenous_. This means that any unit created as, say, `foo@1.service`, should be created only from the unit named `foo@.service`. This homogeneity will be enforced by the fleet API in future.
 
+## Definition of the Install Section
+
+Unit files which have `[Install]` section will be automatically enabled by fleet. This means that states of such unit files may not be tracked by fleet. For example you have loaded `my.service` unit file:
```
joshix: s/which have/which have an/
s/states/the states/
s/may not/cannot/
"For example, assume we have loaded this `my.service` unit file:"

systemd/manager.go
kayrus: It doesn't matter. Current fleet behavior - skip empty sections. If we will keep these sections - systemd will still process such units.

Documentation/unit-files-and-scheduling.md
```diff
@@ -57,6 +57,35 @@ To use instance units, simply create a unit file whose name matches the `<name>@
...
+## Definition of the Install Section
+
+Unit files which have an `[Install]` section will be automatically enabled by fleet. This means that the states of such unit files cannot be tracked by fleet. For example, assume we have loaded this `my.service` unit file:
+
+```ini
+[Service]
+ExecStart=/bin/bash -c "while true; do echo my.service unit file; sleep 1; done"
+```
+
+and then loaded additional [sidekick][sidekick] discovery unit `my_discovery.service`:
```
joshix: s/additional/an additional/

Documentation/unit-files-and-scheduling.md
```diff
@@ -57,6 +57,35 @@ To use instance units, simply create a unit file whose name matches the `<name>@
...
+ExecStart=/bin/bash -c "while true; do echo my_discovery.service unit file; sleep 1; done"
+
+[Install]
+WantedBy=my.service
+
+[X-Fleet]
+MachineOf=my.service
+```
+
+Fleet will load and automatically enable unit above because it contains `[Install]` section. Systemd will create a `my.service.wants/my_discovery.service` symlink to the `my_discovery.service` unit file. And when you start `my.service`, systemd will start `my_discovery.service` independently on `my_discovery.service` desired state defined in fleet. This may cause confusion between `fleetctl list-units` and `systemctl list-units`. It is recommended to use `fleetctl status my_discovery.service` to get real unit status (`fleetctl status` just passes arguments to `systemctl status` command).
```
joshix: fleet will load and enable the `my_discovery.service` unit above because it contains an `[Install]` section. When `my.service` is started, systemd will also start `my_discovery.service`, independent of the desired state defined for `my_discovery.service` in fleet. This can cause confusion between the output of `fleetctl list-units` and `systemctl list-units` output, which will not match in this scenario. Use `fleetctl status my_discovery.service` to explicitly identify the service and get its actual unit status.

kayrus: Done. Removed second **output**

joshix:
docs LGTM for purpose

functional/install_test.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Ensure we can SSH into machine using fleetctl
+	for _, machine := range machines {
+		if stdout, stderr, err := cluster.Fleetctl(m0, "--strict-host-key-checking=false", "ssh", machine, "uptime"); err != nil {
+			t.Fatalf("Unable to SSH into fleet machine: \nstdout: %s\nstderr: %s\nerr: %v", stdout, stderr, err)
+		}
+	}
```
antrik: Why do you need to check this here? We have separate tests for this?...

functional/install_test.go
```diff
@@ -0,0 +1,111 @@
...
+		stdout, _, err := cluster.Fleetctl(m0, "--strict-host-key-checking=false", "ssh", "discovery.service", "systemctl show --property=ActiveState discovery.service")
+		if err != nil {
+			t.Error("Failed getting info using remote systemctl")
+		}
+		stdout = strings.TrimSpace(stdout)
+		if stdout == "ActiveState=active" {
+			return true
+		} else {
+			return false
+		}
```
antrik: You don't need an `if` here: the comparison already produces a bool...

functional/install_test.go
```diff
@@ -0,0 +1,111 @@
...
+		if err != nil {
+			t.Error("Failed getting info using remote systemctl")
+		}
+		stdout = strings.TrimSpace(stdout)
+		if stdout == "ActiveState=inactive" {
+			return true
+		} else {
+			return false
+		}
+	}
```
antrik: These two functions are almost identical... Would be nice to split out the common part.

functional/install_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Verify that discovery.service unit is stopped
```
antrik: I think this comment is a bit misleading here: it sounds as if we are waiting for an active unit to become inactive -- which is the case for the other invocation later on, but not here.

systemd/manager.go
antrik: Well, my concern is that this condition seems misleading at best: it suggests that we are skipping empty sections on purpose. (While only implicitly skipping entirely missing sections.) I don't think that's really an intentional choice?

From your other comments, I understand that the way fleet parses the units currently, empty sections will never show up here; so "is present" and "has a length" are effectively the same for now. However, that is an implementation detail that might change at some point -- and either way, it's always better to be explicit about intentions...

kayrus:
@antrik updated.

antrik:
Looks good to me now :-)

jonboulle:
Punting this to the next milestone as @kayrus thinks it should be reconsidered (?) but is afk

dongsupark:
This PR looks good to me.

Besides I had to think about a couple of open questions:
* Unit states from `"fleetctl list-units"` do not match with those from `"systemctl list-units"`
** : That's not a new issue, as addressed by https://github.com/coreos/fleet/issues/464. In the future we might need to implement `X-Requires` to avoid the inconsistency.
* We have not merged this one, in order to avoid potential conflicts with other PRs like unit state caches (now obsolete) or gRPC (still pending). Though as far as I can tell, gRPC PR doesn't seem to have anything to do with the Install section.

So I think I'll merge this one, probably some time in August.

dongsupark:
Closed via https://github.com/coreos/fleet/pull/1655

Issue #1526, opened 2016-04-01
ssh: golang/crypto/ssh prioritizes server's kex algo instead of parsing known_hosts first


kayrus:
Original issue was reported here: https://github.com/coreos/bugs/issues/1186

The problem is that golang ssh implementation compares client's `ServerHostKeyAlgos` and server's `ServerHostKeyAlgos` and chooses first match depending on the order in `supportedHostKeyAlgos` variable:

```go
// supportedKexAlgos specifies the supported host-key algorithms (i.e. methods
// of authenticating servers) in preference order.
var supportedHostKeyAlgos = []string{
        CertAlgoRSAv01, CertAlgoDSAv01, CertAlgoECDSA256v01,
        CertAlgoECDSA384v01, CertAlgoECDSA521v01,

        KeyAlgoECDSA256, KeyAlgoECDSA384, KeyAlgoECDSA521,
        KeyAlgoRSA, KeyAlgoDSA,
}
```

If remote host had `KeyAlgoRSA` key before and it was written into client's `.fleetctl/known_hosts`, then SSH on the remote host was upgraded (i.e. `ecdsa-sha2-nistp256` support was added), then `crypto/ssh` automatically chooses first match: `KeyAlgoECDSA256` and doesn't take into consideration already existing fingerprints. This causes `Failed building SSH client: ssh: handshake failed: host key mismatch` error message on the fleetctl client's side.

kayrus:
OpenSSH logic:

* load known_hosts keys
* parse them, determine whether remote hostname/IP corresponds to the known_hosts
* set hostkeyalgs order (`order_hostkeyalgs`)
* send `SSH2_MSG_KEXINIT`
* receive host's fingerprint and compare it to the entry inside known_hosts

golang/crypto/ssh logic:

* send `SSH2_MSG_KEXINIT`
* receive host's fingerprint and compare it to the entry inside known_hosts

kayrus:
`ClientConfig` struct contains `HostKeyAlgorithms`. We can `ParseKnownHosts` crypto/ssh function to parse known_hosts and set `HostKeyAlgorithms` inside `sshClientConfig`.

Here is an example on how to parse known_hosts files https://github.com/golang/crypto/blob/master/ssh/keys_test.go#L391

PR #1527, opened 2016-04-01
[WIP] ssh: define known_hosts key algo before handshake

kayrus:
resolves https://github.com/coreos/fleet/issues/1526 and https://github.com/coreos/bugs/issues/1186

TODO:
* write functional test
* test with --tunnel, remote commands, etc.

### DIFF
--- a/ssh/known_hosts.go
+++ b/ssh/known_hosts.go
@@ -87,6 +87,47 @@ func NewHostKeyChecker(m HostKeyManager) *HostKeyChecker {
 	return &HostKeyChecker{m, askToTrustHost}
 }
 
+// Returns public key algo if known_hosts contains an entry for the specified remote host
+func (kc *HostKeyChecker) GetKeyType(addr string) []string {
+	var result []string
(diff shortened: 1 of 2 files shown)


ssh/known_hosts.go
```diff
@@ -87,6 +87,47 @@ func NewHostKeyChecker(m HostKeyManager) *HostKeyChecker {
...
+			if !matchHost(ipAddr, pattern) {
+				continue
+			}
+		}
+		for _, hostKey := range keys {
+			// Return first matching key
+			result = []string{hostKey.Type()}
+			return result
+		}
+	}
```
tixxdz: A first quick look suggests that you should return a list of keys into HostKeyAlgorithms ?

tixxdz:
An updated version is here: https://github.com/coreos/fleet/pull/1529
It was tested with -tunnel, so still to write a functional test for it. In all cases we have to merge that one next hours.

jonboulle:
closed by #1529

PR #1528, opened 2016-04-01
Cleanup: functional tests: Drop gratuitous `sudo` from various commands

antrik:
Diagnostic commands such as `systemctl status` don't need elevated
permissions.

One notable exception is `journalctl`, which is actually privileged
(when accessing system journals), as it might reveal sensitive
information...


tixxdz:
lgtm

jonboulle:
nit, commit message style is "subsystem: fooed a bar"

PR #1529, opened 2016-04-01
ssh: define the list of Key Algorithms of remote hosts before handshake

### LINKED ISSUE #1526: ssh: golang/crypto/ssh prioritizes server's kex algo instead of parsing known_hosts first

tixxdz:
Retrieve remote host Key Algorithms from known_host if they are there
and use them to perform ssh handshake. Otherwise fallback to default
values suggested by remote.

This patch is based from a previous patch written by:
kayrus <kay.diam@gmail.com>

Resolves #1526 and coreos/bugs#1186

### DIFF
--- a/ssh/known_hosts.go
+++ b/ssh/known_hosts.go
@@ -87,6 +87,48 @@ func NewHostKeyChecker(m HostKeyManager) *HostKeyChecker {
 	return &HostKeyChecker{m, askToTrustHost}
 }
 
+// Returns public key algorithms of the remote host that are listed
+// inside known_hosts
+func (kc *HostKeyChecker) GetHostKeyAlgorithms(addr string) []string {
+	var results []string
+	remoteAddr, err := kc.addrToHostPort(addr)
+	if err != nil {
+		log.Errorf("Failed to parse address %v: %v", addr, err)
+		return nil
+	}
+
+	hostKeys, err := kc.m.GetHostKeys()
+
+	_, ok := err.(*os.PathError)
+	if err != nil && !ok {
+		log.Errorf("Failed to read known_hosts file %v: %v", kc.m.String(), err)
+		return nil
+	}
+
+	for pattern, keys := range hostKeys {
+		if !matchHost(remoteAddr, pattern) {
+			remoteIP, err := net.ResolveTCPAddr("tcp", addr)
+			if err != nil {
+				log.Errorf("Failed to resolve TCP address %v: %v", addr, err)
+				continue
+			}
+			ipAddr, err := kc.addrToHostPort(remoteIP.String())
+			if err != nil {
+				log.Errorf("Failed to parse address %v: %v", remoteIP.String(), err)
+				continue
+			}
+			if !matchHost(ipAddr, pattern) {
+				continue
+			}
+		}
+		for _, hostKey := range keys {
+			results = append(results, hostKey.Type())
+		}
+	}
+
+	return results
+}
+
 // Check is called during the handshake to check the server's public key for
 // unexpected changes. The key argument is in SSH wire format. It can be parsed
 // using ssh.ParsePublicKey. The address before DNS resolution is passed in the
--- a/ssh/ssh.go
+++ b/ssh/ssh.go
@@ -171,7 +171,7 @@ func SSHAgentClient() (gosshagent.Agent, error) {
 	return gosshagent.NewClient(agent), nil
 }
 
-func sshClientConfig(user string, checker *HostKeyChecker) (*gossh.ClientConfig, error) {
+func sshClientConfig(user string, checker *HostKeyChecker, addr string) (*gossh.ClientConfig, error) {
 	agentClient, err := SSHAgentClient()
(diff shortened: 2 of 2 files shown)


ssh/known_hosts.go
```diff
@@ -87,6 +87,50 @@ func NewHostKeyChecker(m HostKeyManager) *HostKeyChecker {
...
+	_, ok := err.(*os.PathError)
+	if err != nil && !ok {
+		log.Errorf("Failed to read known_hosts file %v: %v", kc.m.String(), err)
+		return nil
+	}
+
+	for pattern, keys := range hostKeys {
+		if !matchHost(remoteAddr, pattern) {
+			remoteIP, err := net.ResolveTCPAddr("tcp", addr)
+			if err != nil {
```
crawford: Can we log this error?

ssh/known_hosts.go
```diff
@@ -87,6 +87,50 @@ func NewHostKeyChecker(m HostKeyManager) *HostKeyChecker {
...
+				continue
+			}
+			ipAddr, err := kc.addrToHostPort(remoteIP.String())
+			if err != nil {
```
crawford: Same here?

ssh/known_hosts.go
```diff
@@ -87,6 +87,50 @@ func NewHostKeyChecker(m HostKeyManager) *HostKeyChecker {
 	return &HostKeyChecker{m, askToTrustHost}
 }
 
+// Returns public key algorithms of the remote host that are listed
+// inside known_hosts
+func (kc *HostKeyChecker) GetHostKeyAlgorithms(addr string) []string {
+	results := make([]string, 0)
```
crawford:
```go
var results []string
```

ssh/known_hosts.go
```diff
@@ -87,6 +87,50 @@ func NewHostKeyChecker(m HostKeyManager) *HostKeyChecker {
...
+			if !matchHost(ipAddr, pattern) {
+				continue
+			}
+		}
+		for _, hostKey := range keys {
+			results = append(results, hostKey.Type())
+		}
+	}
+
+	if len(results) == 0 {
```
crawford: With the above declaration, you can omit this check.

crawford:
Without this patch:

```
$  bin/fleetctl --tunnel=ec2-54-153-17-9.us-west-1.compute.amazonaws.com list-machines
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
IT IS POSSIBLE THAT SOMEONE IS DOING SOMETHING NASTY!
Someone could be eavesdropping on you right now (man-in-the-middle attack)!
It is also possible that a host key has just been changed.
The fingerprint for the ECDSA key sent by the remote host is
08:b1:9f:a2:50:55:72:1c:df:fa:ae:a1:13:ff:fe:30.
Please contact your system administrator.
Add correct host key in /home/alex/.fleetctl/known_hosts to get rid of this message.
Host key verification failed.
Unable to initialize client: failed initializing SSH client: ssh: handshake failed: host key mismatch
```

With the patch:

```
$ bin/fleetctl --tunnel=ec2-54-153-17-9.us-west-1.compute.amazonaws.com list-machines
MACHINE         IP              METADATA
bb926b38...     172.31.1.5      -
```

LGTM

tixxdz:
@crawford @jonboulle @kayrus  this seems to work and it is a high priority fix, If no one objects I'll merge it tonight. The code doesn't have side effects, on errors we return nil which the upstream ssh library will ignore and fallback to the old behaviour. Thanks!

Issue #1504, opened 2016-03-14
Intermittent failure in TestScheduleOneWayConflict


antrik:
This test sometimes fails (see https://semaphoreci.com/coreos/fleet/branches/pull-request-1501/builds/3 ) due to a race condition in the test case. There is actually a [comment](https://github.com/coreos/fleet/blob/master/functional/scheduling_test.go#L286) in there saying:

```
// TODO(jonboulle): fix this race. Since we no longer immediately
// remove unit state on unit destruction (and instead wait for
// UnitStateGenerator/UnitStatePublisher to clean up), the old unit
// shows up as active for quite some time.
```

I guess it could be fixed by using a similar loop as the one in [WaitForNActiveUnits()](https://github.com/coreos/fleet/blob/master/functional/platform/nspawn.go#L107). I wonder though whether checking the number of known units like that before checking the active units is really necessary? Maybe it could just be moved after the active units check -- or even dropped altogether?..

jonboulle:
@antrik I think you might be right - looks like it's probably just redundant with WaitForNActiveUnits?

antrik:
@jonboulle well, it's not *entirely* redundant, as it checks the total number of units, rather than "only" the active ones. It's not clear to me though whether or not there is much value in doing this...

Note that once https://github.com/coreos/fleet/pull/1501/commits/520a90f269c3bfd554d99f4b39371eb37a157c96 lands, it will be very easy to fix the race while keeping the extra test. I'm fine with either option.

antrik:
@jonboulle in fact, https://github.com/coreos/fleet/pull/1509/commits/8a22f07cf4c842723491dbbf06335fba0e41e9fe (and the followup https://github.com/coreos/fleet/pull/1509/commits/cd3a786cb655e5bb9d8eb46a5ad2283ab87aeb1b ) want to add a dedicated function just for that, which could be used here as well... I could just go ahead and add this function independently.

antrik:
Finding an appropriate wait condition turned out slightly more involved in the end; but it's working now with https://github.com/coreos/fleet/pull/1531.

PR #1531, opened 2016-04-04
Functional tests: Fix race condition in TestScheduleOneWayConflict()

antrik:
Wait for the `destroy` command to take effect using a proper wait
condition (with `WaitForState()`), instead of a fixed sleep period --
which was sometimes insufficient, causing intermittent failure.

This fixes https://github.com/coreos/fleet/issues/1504

The condition function is mostly based on `WaitForNActiveUnits()`, with
some adjustments for the different scenario.

### DIFF
--- a/functional/scheduling_test.go
+++ b/functional/scheduling_test.go
@@ -20,7 +20,6 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/coreos/fleet/functional/platform"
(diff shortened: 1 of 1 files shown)


tixxdz:
Seems ok, but is there a way to make it more readable ? I mean don't pass the whole function as a parameter ? thank you!

jonboulle:
I think passing the whole function is fine in this case. LGTM!

PR #1534, opened 2016-04-06
Add fleet-grpc related godeps packages #1426

hectorj2f:
These Go deps are related to the PR https://github.com/coreos/fleet/pull/1426. As discussed in https://github.com/coreos/fleet/pull/1426#issuecomment-205339036, it should be better to include the go packages in a different PR, instead of in the same PR #1426 .


tixxdz:
/cc @jonboulle @kayrus  

Hi @hectorj2f thanks for the patch, and as you are aware by https://github.com/coreos/fleet/pull/1537 there will be a switch, but after v0.12 is released which is I guess ready (and to not delay it further...) after that #1537 should be merged and you can continue with a clean revision.

@hectorj2f does that sound reasonable ?

Thanks!

hectorj2f:
@tixxdz Yes, ti works for me :). Thanks

kayrus:
@hectorj2f could you also fix commit message? i.e. it should begin with the "vendor: added dependency... etc.."

hectorj2f:
@kayrus Totally, once you merge it ;).

hectorj2f:
@dongsupark We don't use `godeps` anymore. So, should I update this PR with the vendoring or replace it with a new branch-PR ?

dongsupark:
@hectorj2f It doesn't matter. Either way I'm fine.

hectorj2f:
This PR is obsolete as it is now replaced by https://github.com/coreos/fleet/pull/1657

PR #1541, opened 2016-04-08
functional: fleetctl submit multiple units

dongsupark:
Make ``testUnitSubmit()`` also do an additional test for submission of multiple units. It copies multiple units with different names, and verify that they are submitted.

After redestroying all submitted units, it checks for every unit being completely removed, making use of ``util.WaitForState()`` etc.

This was originally implemented as my own spin-off from #1509. Though I got this decoupled from other unmerged commits, using instead ``util.WaitForState()``.

### DIFF
--- a/functional/unit_action_test.go
+++ b/functional/unit_action_test.go
@@ -16,11 +16,19 @@ package functional
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"testing"
 
 	"github.com/coreos/fleet/functional/platform"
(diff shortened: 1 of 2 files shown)


dongsupark:
This is definitely nice-to-have. But now I think this entire code is outdated. I'm going to rewrite this PR based on #1544 , once the #1544 got merged.

dongsupark:
Closed #1541 in favor of https://github.com/coreos/fleet/pull/1552

PR #1542, opened 2016-04-08
tests: move all files' writes into loop

kayrus:
beautifying the code

### DIFF
--- a/functional/platform/nspawn.go
+++ b/functional/platform/nspawn.go
@@ -346,36 +346,33 @@ func (nc *nspawnCluster) createMember(id string) (m Member, err error) {
 		}
 	}
 
-	sshd_config := `# Use most defaults for sshd configuration.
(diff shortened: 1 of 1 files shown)


functional/platform/nspawn.go
```diff
@@ -346,37 +346,34 @@ func (nc *nspawnCluster) createMember(id string) (m Member, err error) {
...
-		return
-	}
-
 	filesContents := []struct {
 		path     string
 		contents string
 		mode     os.FileMode
 	}{
 		{
+			"/etc/ssh/sshd_config",
```
tixxdz: It seems the fsdir path is missing.. guess it doesn't change the where files are written ?!  otherwise lgtm, thanks!

kayrus: fsdir is defined earlier.

PR #1544, opened 2016-04-11
functional: more fleetctl unit action tests

dongsupark:
In order to make functional tests cover more fleetctl commands, do the following:

* define ``util.UnitFileState`` for handling output of list-unit-files
* aadd new cluster operations ``WaitForNUnits`` and ``WaitForNUnitFiles``, wrappers based on ``util.WaitForState``.
* introduce a new test ``TestUnitCat`` for "``fleetctl cat``".
* introduce a new test ``TestUnitStatus`` for "``fleetctl status``".
* improve an existing test ``TestUnitSubmit``.
* introduce a new test ``TestUnitLoad`` for ``fleetctl load``.

### DIFF
--- a/functional/platform/nspawn.go
+++ b/functional/platform/nspawn.go
@@ -128,6 +128,47 @@ func (nc *nspawnCluster) FleetctlWithInput(m Member, input string, args ...strin
 	return util.RunFleetctlWithInput(input, args...)
 }
 
+// WaitForNUnits runs fleetctl list-units to verify the actual number of units
+// matched with the given expected number. It periodically runs list-units
+// waiting until list-units actually shows the expected units.
+func (nc *nspawnCluster) WaitForNUnits(m Member, expectedUnits int) (map[string][]util.UnitState, error) {
+	var nUnits int
+	retStates := make(map[string][]util.UnitState)
+	checkListUnits := func() bool {
+		outListUnits, _, err := nc.Fleetctl(m, "list-units", "--no-legend", "--full", "--fields", "unit,active,machine")
+		if err != nil {
+			return false
+		}
+		// NOTE: There's no need to check if outListUnits is expected to be empty,
+		// because ParseUnitStates() implicitly filters out such cases.
+		// However, in case of ParseUnitStates() going away, we should not
+		// forget about such special cases.
+		units := strings.Split(strings.TrimSpace(outListUnits), "\n")
+		allStates := util.ParseUnitStates(units)
+		nUnits = len(allStates)
+		if nUnits != expectedUnits {
+			return false
+		}
+
+		for _, state := range allStates {
+			name := state.Name
+			if _, ok := retStates[name]; !ok {
+				retStates[name] = []util.UnitState{}
+			}
+			retStates[name] = append(retStates[name], state)
+		}
+		return true
+	}
+
+	timeout, err := util.WaitForState(checkListUnits)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find %d units within %v (last found: %d)",
+			expectedUnits, timeout, nUnits)
+	}
+
+	return retStates, nil
+}
+
 func (nc *nspawnCluster) WaitForNActiveUnits(m Member, count int) (map[string][]util.UnitState, error) {
 	var nactive int
 	states := make(map[string][]util.UnitState)
@@ -163,6 +204,49 @@ func (nc *nspawnCluster) WaitForNActiveUnits(m Member, count int) (map[string][]
 	}
 
 	return states, nil
+}
+
+// WaitForNUnitFiles runs fleetctl list-unit-files to verify the actual number of units
+// matched with the given expected number. It periodically runs list-unit-files
+// waiting until list-unit-files actually shows the expected units.
+func (nc *nspawnCluster) WaitForNUnitFiles(m Member, expectedUnits int) (map[string][]util.UnitFileState, error) {
+	var nUnits int
+	retStates := make(map[string][]util.UnitFileState)
+
+	checkListUnitFiles := func() bool {
+		outListUnitFiles, _, err := nc.Fleetctl(m, "list-unit-files", "--no-legend", "--full", "--fields", "unit,dstate,state")
+		if err != nil {
+			return false
+		}
+		// NOTE: There's no need to check if outListUnits is expected to be empty,
+		// because ParseUnitFileStates() implicitly filters out such cases.
+		// However, in case of ParseUnitFileStates() going away, we should not
+		// forget about such special cases.
+		units := strings.Split(strings.TrimSpace(outListUnitFiles), "\n")
+		allStates := util.ParseUnitFileStates(units)
+		nUnits = len(allStates)
+		if nUnits != expectedUnits {
+			// retry until number of units matched
+			return false
+		}
+
+		for _, state := range allStates {
+			name := state.Name
+			if _, ok := retStates[name]; !ok {
+				retStates[name] = []util.UnitFileState{}
+			}
+			retStates[name] = append(retStates[name], state)
+		}
+		return true
+	}
+
+	timeout, err := util.WaitForState(checkListUnitFiles)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find %d units within %v (last found: %d)",
+			expectedUnits, timeout, nUnits)
+	}
+
+	return retStates, nil
 }
 
 func (nc *nspawnCluster) WaitForNMachines(m Member, count int) ([]string, error) {
--- a/functional/unit_action_test.go
+++ b/functional/unit_action_test.go
@@ -15,6 +15,8 @@
 package functional
 
 import (
+	"io/ioutil"
+	"path"
 	"strings"
 	"testing"
 
@@ -222,5 +224,54 @@ func TestUnitSSHActions(t *testing.T) {
 
 	if !strings.Contains(stdout, "Hello, World!") {
 		t.Errorf("Could not find expected string in journal output:\n%s", stdout)
+	}
+}
+
+// TestUnitCat simply compares body of a unit file with that of a unit fetched
+// from the remote cluster using "fleetctl cat".
+func TestUnitCat(t *testing.T) {
+	cluster, err := platform.NewNspawnCluster("smoke")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cluster.Destroy()
+
+	m, err := cluster.CreateMember()
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = cluster.WaitForNMachines(m, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// read a sample unit file to a buffer
+	unitFile := "fixtures/units/hello.service"
+	fileBuf, err := ioutil.ReadFile(unitFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+	fileBody := strings.TrimSpace(string(fileBuf))
+
+	// submit a unit and assert it shows up
+	_, _, err = cluster.Fleetctl(m, "submit", unitFile)
+	if err != nil {
+		t.Fatalf("Unable to submit fleet unit: %v", err)
+	}
+	// wait until the unit gets submitted up to 15 seconds
+	_, err = cluster.WaitForNUnitFiles(m, 1)
+	if err != nil {
+		t.Fatalf("Failed to run list-units: %v", err)
+	}
+
+	// cat the unit file and compare it with the original unit body
+	stdout, _, err := cluster.Fleetctl(m, "cat", path.Base(unitFile))
+	if err != nil {
+		t.Fatalf("Unable to submit fleet unit: %v", err)
+	}
+	catBody := strings.TrimSpace(stdout)
+
+	if strings.Compare(catBody, fileBody) != 0 {
+		t.Fatalf("unit body changed across fleetctl cat: \noriginal:%s\nnew:%s", fileBody, catBody)
 	}
 }
--- a/functional/unit_action_test.go
+++ b/functional/unit_action_test.go
@@ -129,6 +129,81 @@ func TestUnitSubmit(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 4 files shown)


functional/unit_action_test.go
```diff
@@ -73,48 +76,47 @@ func TestUnitSubmit(t *testing.T) {
...
 
+	unitFile := "fixtures/units/hello.service"
+
 	// submit a unit and assert it shows up
-	if _, _, err := cluster.Fleetctl(m, "submit", "fixtures/units/hello.service"); err != nil {
+	_, _, err = cluster.Fleetctl(m, "submit", unitFile)
+	if err != nil {
 		t.Fatalf("Unable to submit fleet unit: %v", err)
 	}
-	stdout, _, err := cluster.Fleetctl(m, "list-units", "--no-legend")
```
tixxdz: @dongsupark thanks for the patches!

Hmm something is wrong here, my understanding with fleet is submit only sends the units into the cluster, so it will show up when you do "fleetctl list-unit-files" but not with "fleetctl list-units" since the unit is not active.

Was this test wrong from the beginning ? we should not check the number of printed lines, we should just grep for the name of the unit, thoughts ?

Thanks!

dongsupark: @tixxdz Yes, this part has been wrong since the day 0. (5f1340b3)
I'm not sure how the test has always succeeded. I'll try to improve it.

jonboulle: It wasn't "wrong", just not specific enough, right?

tixxdz: @jonboulle yes! by "wrong" I'm only referring to the functional tests and that perhaps there is a bug somewhere that allowed it to pass. For fleetctl commands they are clear.

tixxdz: @jonboulle yes not specific enough ;-)

tixxdz: @dongsu capturing the context, maybe we need a function `runCheckListUnits()` , `runCheckListUnitFiles()` and if you want to grep submitted units since they won't show up in other commands and you are not sure about the number of returned units... then make runGetListUnitFiles() return the units and grep it into the caller ? just quick thoughts.

thanks

dongsupark:
Updated.
* add new utilities ``util.SliceContains`` and ``util.GetListUnitNames``.
* add a new helper ``runListUnitFiles``, a wrapper for ``fleetctl list-unit-files``.
* improve ``TestUnitSubmit`` to correctly check the output running by list-unit-files instead of list-units, as well as to parse output to compare the unit name with the expected name.
* improve ``TestUnitLoad`` to parse output to compare the unit name with the expected name.

functional/unit_action_test.go
```diff
@@ -69,47 +76,132 @@ func TestUnitSubmit(t *testing.T) {
...
 	}
 
+	unitFile := "fixtures/units/hello.service"
+	inputUnitFiles := make([]string, 0)
+	inputUnitFiles = append(inputUnitFiles, path.Base(unitFile))
+
 	// submit a unit and assert it shows up
-	if _, _, err := cluster.Fleetctl(m, "submit", "fixtures/units/hello.service"); err != nil {
+	_, _, err = cluster.Fleetctl(m, "submit", unitFile)
+	if err != nil {
```
antrik: Why did you change the syntax here? There is no reason to have the assignment outside the `if`.

functional/unit_action_test.go
antrik: @jonboulle no, it was indeed completely wrong: `list-units` returns an empty string; but that's still counts as one line, so the test "passed"...

antrik: Note that most of the existing tests only check the number of units, except for cases where there is actually a possibility for ambiguity -- so checking only the count would be fine here as well. It just needs to be fixed to use `list-unit-files`; and to properly check for a non-empty string...

functional/unit_action_test.go
```diff
@@ -69,47 +76,132 @@ func TestUnitSubmit(t *testing.T) {
...
+	inputUnits := make([]string, 0)
+	inputUnits = append(inputUnits, path.Base(unitFile))
+
+	// load a unit and assert it shows up
+	_, _, err = cluster.Fleetctl(m, "load", unitFile)
+	if err != nil {
+		t.Fatalf("Unable to load fleet unit: %v", err)
+	}
+
+	// wait until the unit gets unloaded up to 15 seconds
```
antrik: s/unloaded/loaded/

functional/unit_action_test.go
```diff
@@ -224,3 +316,139 @@ func TestUnitSSHActions(t *testing.T) {
...
+	m, err := cluster.CreateMember()
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = cluster.WaitForNMachines(m, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// submit a unit and assert it shows up
```
antrik: Eh?

functional/unit_action_test.go
```diff
@@ -224,3 +316,139 @@ func TestUnitSSHActions(t *testing.T) {
...
+	_, _, err = cluster.Fleetctl(m, "status")
+	if err != nil {
+		t.Fatalf("Unable to get fleet status: %v", err)
+	}
+}
+
+// runListUnits runs fleetctl list-units to verify the actual number of units
+// matched with the given expected number. It periodically runs list-units
+// waiting until list-units actually shows the expected units.
+func runListUnits(cluster platform.Cluster, m platform.Member, expectedUnits int) ([]string, error) {
```
antrik: Can't you just use the `WaitForNUnits()` function you introduced in another PR? Seems to do the same more or less...

functional/unit_action_test.go
```diff
@@ -224,3 +316,139 @@ func TestUnitSSHActions(t *testing.T) {
...
+	}
+
+	return units, nil
+}
+
+// runListUnitFiles runs fleetctl list-unit-files to verify the actual number
+// of units matched with the given expected number. It periodically runs
+// list-unit-files waiting until list-unit-files actually shows the expected
+// number of unit files.
+func runListUnitFiles(cluster platform.Cluster, m platform.Member, expectedUnits int) ([]string, error) {
```
antrik: For consistency, this probably should be `WaitForNUnitFiles()`, with similar semantics to `WaitForNUnits()` and `WaitForNActiveUnits()`.

functional/util/util.go
```diff
@@ -154,3 +154,32 @@ func WaitForState(stateCheckFunc func() bool) (time.Duration, error) {
 		}
 	}
 }
+
+// SliceContains returns true if every string in needle slice exists in any
+// slice in haystack.
+func SliceContains(haystack []string, needle []string) bool {
```
antrik: You should probably check that the list of units is *identical* to the expected ones, not that it's a superset...

But for the actual test cases in question, it's probably overkill anyway -- see my first comment, about just checking the count...

functional/util/util.go
```diff
@@ -154,3 +154,32 @@ func WaitForState(stateCheckFunc func() bool) (time.Duration, error) {
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// GetListUnitNames returns the 1st field of the given string, which is
+// a single line of output of "fleetctl list-{units,unit-files}".
+func GetListUnitNames(listUnitFiles []string) []string {
```
antrik: Note that when you go with a `WaitForNUnits()` and `WaitForNUnitFiles()` along the lines of the existing `WaitForNActiveUnits()` -- i.e. returning a map -- you won't need this helper...

functional/unit_action_test.go
antrik: I just saw that one of the existing tests in this very file (namely `TestUnitRunnable()`) actually test both the count *and* the name -- so feel free to copy that if you really think it's worthwhile...

dongsupark: @antrik 
> no, it was indeed completely wrong: list-units returns an empty string; but that's still counts as one line, so the test "passed"...

Yes, that was exactly what I wanted to say. That's why I said yesterday that this part has been wrong since the beginning. I think the underlying reason was that ``len(strings.Split("", "\n"))`` returns 1, not 0 as expected.

So my solution to get it working was adding additional conditions like "``expectedUnits == 0 && len(outListUnitFiles) == 0``" in each check function. Still somehow nasty, but it's working for now. Afterwards, we should do some code auditing in the entire tree to find such errors. But anyway that would be over the scope of this PR.

functional/unit_action_test.go
tixxdz: That's not a good idea we explicitly want to know from a caller perspective that this command is running `fleetctl list-units` so we use it where it's appropriate.

functional/unit_action_test.go
tixxdz: Same thing here, I've a functional test waiting #1515 that will use the runListUnitFiles() which is `fleetctl list-unit-files` and I want the returned output, so I could do the parsing. So WaitForNUnit() should be either inside these functions or just on its own. Functional tests should just be easy to write I want to run this function and get its output...

functional/util/util.go
tixxdz: That's a good point! so since it's just functional maybe return a map unless some other points that we are not aware of...

tixxdz:
So a couple of changes then maybe merge this one ?

dongsupark:
Updated.
* define ``util.UnitFileState`` for handling output of list-unit-files
* add new cluster operations ``WaitForNUnits`` and ``WaitForNUnitFiles``
* make ``TestUnitSubmit()`` use ``cluster.WaitForNUnitFiles``
* make ``TestUnitLoad()`` use ``cluster.WaitForNUnits``
* fix a bug in ``TestUnitStatus()``, loading a unit before getting status.
* drop unnecessary helpers like ``SliceContains``, ``runListUnits``, etc.
* fix minor things like comments or coding style

I assume this PR should be merged prior to the existing #1509.
So I might also have to update functional tests in #1509 to avoid conflicts.

functional/unit_action_test.go
```diff
@@ -69,47 +74,133 @@ func TestUnitSubmit(t *testing.T) {
...
 		t.Fatalf("Expected no failure when double-submitting unit, got this: %v", err)
 	}
 
 	// destroy the unit and ensure it disappears from the unit list
-	if _, _, err := cluster.Fleetctl(m, "destroy", "fixtures/units/hello.service"); err != nil {
+	if _, _, err := cluster.Fleetctl(m, "destroy", unitFile); err != nil {
 		t.Fatalf("Failed to destroy unit: %v", err)
 	}
-	stdout, _, err = cluster.Fleetctl(m, "list-units", "--no-legend")
+	// wait until the unit gets submitted up to 15 seconds
```
antrik: I think this comment is still wrong here...

functional/unit_action_test.go
antrik: This one is still wrong too.

functional/unit_action_test.go
```diff
@@ -224,3 +315,97 @@ func TestUnitSSHActions(t *testing.T) {
...
+	unitFile := "fixtures/units/hello.service"
+
+	// Load a unit and print out status.
+	// Without loading a unit, it's impossible to run fleetctl status
+	_, _, err = cluster.Fleetctl(m, "load", unitFile)
+	if err != nil {
+		t.Fatalf("Unable to load a fleet unit: %v", err)
+	}
+
+	// wait until the unit gets unloaded up to 15 seconds
```
antrik: And another wrong comment...

functional/unit_action_test.go
```diff
@@ -224,3 +315,97 @@ func TestUnitSSHActions(t *testing.T) {
...
+	_, err = cluster.WaitForNUnits(m, 1)
+	if err != nil {
+		t.Fatalf("Failed to run list-units: %v", err)
+	}
+
+	// NOTE: cluster.Fleetctl() cannot be used due to endpoint being messed up.
```
antrik: Huh? I don't understand this comment...

functional/platform/nspawn.go
```diff
@@ -128,6 +128,46 @@ func (nc *nspawnCluster) FleetctlWithInput(m Member, input string, args ...strin
...
+	var nUnits int
+	retStates := make(map[string][]util.UnitState)
+	checkListUnits := func() bool {
+		outListUnits, _, err := nc.Fleetctl(m, "list-units", "--no-legend", "--full", "--fields", "unit,active,machine")
+		if err != nil {
+			return false
+		}
+		if expectedUnits == 0 && len(outListUnits) == 0 {
+			return true
+		}
```
antrik: Note that `ParseUnitStates()` implicitly filters out an empty line -- so the explicit check shouldn't be necessary here.

functional/platform/nspawn.go
```diff
@@ -165,6 +205,48 @@ func (nc *nspawnCluster) WaitForNActiveUnits(m Member, count int) (map[string][]
...
+	retStates := make(map[string][]util.UnitFileState)
+
+	checkListUnitFiles := func() bool {
+		outListUnitFiles, _, err := nc.Fleetctl(m, "list-unit-files", "--no-legend", "--full", "--fields", "unit,dstate,state")
+		if err != nil {
+			return false
+		}
+		if expectedUnits == 0 && len(outListUnitFiles) == 0 {
+			return true
+		}
```
antrik: Same as above: no need to special-case `0` here.

functional/unit_action_test.go
dongsupark: @antrik Aha, this was my mistake. I fixed it into using ``cluster.Fleetctl()``. Thanks.

functional/platform/nspawn.go
dongsupark: @antrik 
> Note that ParseUnitStates() implicitly filters out an empty line

Yeah, I just got cold feet. Actually I would not like to rely on such implicit behavior. Anyway I removed checking for empty line, and wrote a comment about that.

dongsupark:
Done.
* Fixed a bug in ``TestUnitStatus``, by calling ``cluster.Fleetctl()`` correctly.
* Removed unnecessary checking in ``cluster.WaitForN{Units,UnitFiles}()``
* Fixed comments

tixxdz:
lgtm! thanks all.

Issue #1547, opened 2016-04-11
test: TestParseFilepath failed in docker env


kayrus:
because of using `-u $(id -u):$(id -g)` in docker command environment inside the container seems to be broken and `TestParseFilepath` test fails with the following messages:

```
DEBUG filepath.go:49: Failed to get current home directory: user: unknown userid 1000
DEBUG filepath.go:49: Failed to get current home directory: user: unknown userid 1000
DEBUG filepath.go:60: Failed to get ~'s home directory: user: unknown user ~
--- FAIL: TestParseFilepath (0.07s)
        filepath_test.go:45: Failed parsing out prefix / for ~/a
        filepath_test.go:45: Failed parsing out prefix / for ~
FAIL
```

Here is how docker env looks like:

```sh
I have no name!@84167d7c92c3:/opt/fleet$ env
GOLANG_VERSION=1.5.3
HOSTNAME=84167d7c92c3
TERM=xterm
PATH=/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/opt/fleet
GOLANG_DOWNLOAD_SHA256=43afe0c5017e502630b1aea4d44b8a7f059bf60d7f29dfd58db454d4e4e0ae53
SHLVL=1
HOME=/
GOPATH=/go
GOLANG_DOWNLOAD_URL=https://golang.org/dl/go1.5.3.linux-amd64.tar.gz
OLDPWD=/go
_=/usr/bin/env
```

PR #1548, opened 2016-04-12
tests: fix TestParseFilepath where it is run under an unknown user or $HOME dir

tixxdz:
Make sure that ParseFilepath() returned the same input since it was not
able to determin the current user or the $HOME env variable.

This fixes a bug where tests are under these conditions but at the same
time it adds unit checks for ParseFilepath() to make sure that it is
doing the right thing.

This fixes https://github.com/coreos/fleet/issues/1547

### DIFF
--- a/pkg/filepath_test.go
+++ b/pkg/filepath_test.go
@@ -16,6 +16,7 @@ package pkg
 
 import (
 	"os"
+	"os/user"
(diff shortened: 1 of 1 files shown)


kayrus:
LGTM. both `./test` and `./test-docker` work.

PR #1549, opened 2016-04-12
build-aci: no need to set the docker uid

tixxdz:
This reverts parts of 98f185061c0085 which bumped go version but at the
same time added docker switches to set the uid in order to build
binaries under that user, this is supposed to be used as a build aci
where fleetd is under root.


kayrus:
@tixxdz it doesn't make sense, becasue rkt will run fleetd under root user anyway. it doesn't matter whether fleetd is owned not by root

tixxdz:
closed in favor #1550

PR #1550, opened 2016-04-12
build-aci: improved build-aci script

kayrus:
supersedes #1549

* since actool started to support --owner-root flag ([v0.7.1](https://github.com/appc/spec/commit/d26e9449fdc4533dcf16e25fdb6d5f616fa4436c))
   we can use it to create tar archive with the root owned files.
* added bash trap instead of rm -rf $tgt
* added double quotes to protect aci filename when user passes
  tag with the whitespaces

/cc @tixxdz @jonboulle

### DIFF
--- a/build-aci
+++ b/build-aci
@@ -1,5 +1,4 @@
-#!/bin/bash
-set -e
+#!/bin/bash -e
 
 if [ $# -ne 1 ]; then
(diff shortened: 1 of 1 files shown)


jonboulle:
This LGTM. Should really move to acbuild at some point but is fine for now.

PR #1552, opened 2016-04-15
functional: support testing multiple units

dongsupark:
Make ``TestUnit{Submit,Load,Start}()`` capable of testing multiple units. After launching multiple units for activation, they become deactivated via correspondent commands: ``"submit"`` and ``"destroy"``, ``"load"`` and ``"unload"``, ``"start"`` and ``"stop"`` are coupled respectively. These are based on a common function ``doMultipleUnitsCmd()`` to be used by different tests.

This PR depends on #1544.


dongsupark:
Oops, a wrong branch name. Closed #1552 in favor of #1553.

Issue #614, opened 2014-07-03
fleetctl submit - name conflict


porjo:
When I submit a unit file with the same name as one already in the registry, but where content differs, I get the following message:

```
$ fleetctl submit xxxx.service
WARNING: Job(xxxx.service) in Registry differs from local Unit(xxxx.service)
```

I find this message misleading. My expecation is that the unit file has been updated with the new content, however that is not the case. Running `fleetctl cat xxxx.service` shows the old content.

A clearer message would be:

```
ERROR: Job(xxxx.service) in Registry differs from local Unit(xxxx.service). No change made.
```

Alternatively, a 'force' flag could be added for those occasions when you really do want to overwrite the unit file. In that case the WARNING message would be appropriate.

jonboulle:
Let's just do a simple UX fix improvement here.

antrik:
This UI change seems interconnected with the upcoming addition of the --replace flag (following up on https://github.com/coreos/fleet/pull/1295 and https://github.com/coreos/fleet/pull/1442 ) -- maybe it should in fact be done as part of that?

tixxdz:
@antrik indeed, maybe it should just warn the user "please use --replace" ?!

antrik:
Exact wording TBD; but basically that's what I meant indeed.

tixxdz:
After #1509 is merged this one can just be fixed with this branch: https://github.com/endocode/fleet/commits/tixxdz/fleetctl-warn-user-aboutt-replace

Thanks!

PR #1560, opened 2016-04-19
fleetctl: inform the user about the '-replace' switch in case the units differ

tixxdz:
If the unit in the registry is different from the one that is on disk,
and the user did not provide the '--replace' switch flag, then instead
of just printing a warning, let him know about the 'replace' flag.

Closes: https://github.com/coreos/fleet/issues/614

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -563,7 +563,7 @@ func getUnitFileFromTemplate(uni *unit.UnitNameInfo, fileName string) (*unit.Uni
 	}
 
 	if tmpl != nil {
-		isLocalUnitDifferent(fileName, tmpl, true, false)
+		isLocalUnitDifferent(fileName, tmpl, false)
 		uf = schema.MapSchemaUnitOptionsToUnitFile(tmpl.Options)
 		log.Debugf("Template Unit(%s) found in registry", uni.Template)
 	} else {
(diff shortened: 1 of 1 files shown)


tixxdz:
This PR depends on https://github.com/coreos/fleet/pull/1509

dongsupark:
Looks good. A simple test also works.

antrik:
I think it would be better to just merge this into the other PR.

Also, this surely should be covered by some functional test? As you didn't have to change the expected result of any existing tests, I assume there is none yet for this case, i.e. a new one needs to be added...

tixxdz:
@antrik, sorry but that's not a good idea, each PR on its own now. Other PRs have already been delayed for several days for such cases or many comments without concluding lgtm or comments who lacked the right context of what the PR does... So I'm not adding more patches on top to complicate things.

antrik:
@tixxdz let's not overgeneralise. In this case it really doesn't make sense to have it separate, as it just changes the code introduced by that other PR; and the change itself should be non-controversial -- so having it separate is just unnecessary code churn and reviewer burden. Just implement the desirable behaviour right away.

(Note that the other PR has *not* been reviewed by Jon yet -- so it's not like you would be loosing anything by adding improvements to it...)

dongsupark:
@antrik
When I first saw this PR, I actually thought about adding functional tests for checking the warning message. But I didn't do it, because that looked like just one of minor things or nice-to-have. Moreover, think about what would happen if we change the error message in fleetctl. We would then have to update the functional test again.

Apart from all of that, if you'd really like to see a new functional test for the warning message, sure, I could write it, as I'm also responsible for the functional test for the replace option.

I don't care, whether this PR should be a separate one or not. That's just a matter of taste. I'd just like to see the entire replace option implementation finally merged.

PR #1288, opened 2015-07-03
fleetd: Detecting the existing machine-id

wuqixuan:
Now support detecting the existing machine-id on startup.

Fixes #1241 #615

Issue #1241, opened 2015-06-05
fleet should detect machines with the same machine-id


kelseyhightower:
fleet could provide a better error message when it detects machines with the same machine-id. I have 5 nodes with the same machine-id which cause fleet to log the following errors:

```
Jun 05 04:49:49 localhost fleet[11544]: ERROR engine.go:218: Engine leadership lost, renewal failed: 101: Compare failed ([260 != 261]) [261]
Jun 05 04:49:53 localhost fleet[11544]: ERROR engine.go:218: Engine leadership lost, renewal failed: 101: Compare failed ([262 != 264]) [264]
Jun 05 04:49:57 localhost fleet[11544]: ERROR engine.go:218: Engine leadership lost, renewal failed: 101: Compare failed ([265 != 266]) [269]
Jun 05 04:50:01 localhost fleet[11544]: ERROR engine.go:218: Engine leadership lost, renewal failed: 101: Compare failed ([271 != 272]) [272]
Jun 05 04:50:05 localhost fleet[11544]: ERROR engine.go:218: Engine leadership lost, renewal failed: 101: Compare failed ([273 != 274]) [274]
Jun 05 04:50:09 localhost fleet[11544]: ERROR engine.go:218: Engine leadership lost, renewal failed: 101: Compare failed ([275 != 277]) [277]
Jun 05 04:50:13 localhost fleet[11544]: ERROR engine.go:218: Engine leadership lost, renewal failed: 101: Compare failed ([278 != 279]) [283]
```

If you look at the output from the fleet `list-machines` command you see the root cause:

```
$ /opt/bin/fleetctl --endpoint=http://10.20.132.53:2379 list-machines
MACHINE     IP      METADATA
3f44ca14... 10.20.132.53    -
core@localhost /opt/bin $ /opt/bin/fleetctl --endpoint=http://10.20.132.53:2379 list-machines
MACHINE     IP      METADATA
3f44ca14... 10.20.132.56    -
core@localhost /opt/bin $ /opt/bin/fleetctl --endpoint=http://10.20.132.53:2379 list-machines
MACHINE     IP      METADATA
3f44ca14... 10.20.132.57    -
core@localhost /opt/bin $ /opt/bin/fleetctl --endpoint=http://10.20.132.53:2379 list-machines
MACHINE     IP      METADATA
3f44ca14... 10.20.132.55    -
core@localhost /opt/bin $ /opt/bin/fleetctl --endpoint=http://10.20.132.53:2379 list-machines
[...]

PR #1561, opened 2016-04-19
fleetd: detect the existing machine ID

dongsupark:
Detect the existing machine ID when fleetd starts up, to avoid uncomfortable errors when creating multiple machines in a cluster. In addition to the fix in fleetd, this PR also contains a functional test to cover such a case.

Functional test was suggested by @antrik .
/cc @wuqixuan 
Based on https://github.com/coreos/fleet/pull/1288
Fixes: https://github.com/coreos/fleet/issues/615
See also https://github.com/coreos/fleet/issues/1241

### DIFF
--- a/functional/node_test.go
+++ b/functional/node_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/coreos/fleet/functional/platform"
+	"github.com/coreos/fleet/functional/util"
 )
 
 // Simulate the shutdown of a single fleet node
@@ -80,5 +81,104 @@ func TestNodeShutdown(t *testing.T) {
 	stdout, _ = cluster.MemberCommand(m0, "systemctl", "status", "hello.service")
 	if !strings.Contains(stdout, "Active: inactive") {
 		t.Fatalf("Unit hello.service not reported as inactive:\n%s\n", stdout)
+	}
+}
+
+// TestDetectMachineId checks for etcd registration failing on a duplicated
+// machine-id on different machines.
+// First it creates a cluster with 2 members, m0 and m1. Then make their
+// machine IDs the same as each other, by explicitly setting the m1's ID to
+// the same as m0's. Test succeeds when an error returns, while test fails
+// when nothing happens.
+func TestDetectMachineId(t *testing.T) {
+	cluster, err := platform.NewNspawnCluster("smoke")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cluster.Destroy()
+
+	members, err := platform.CreateNClusterMembers(cluster, 2)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	m0 := members[0]
+	m1 := members[1]
+	_, err = cluster.WaitForNMachines(m0, 2)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	machineIdFile := "/etc/machine-id"
+
+	// Restart fleet service, and check if its systemd status is still active.
+	restartFleetService := func(m platform.Member) error {
+		stdout, err := cluster.MemberCommand(m, "sudo", "systemctl", "restart", "fleet.service")
+		if err != nil {
+			return fmt.Errorf("Failed to restart fleet service\nstdout: %s\nerr: %v", stdout, err)
+		}
+
+		stdout, _ = cluster.MemberCommand(m, "systemctl", "show", "--property=ActiveState", "fleet")
+		if strings.TrimSpace(stdout) != "ActiveState=active" {
+			return fmt.Errorf("Fleet unit not reported as active: %s", stdout)
+		}
+		stdout, _ = cluster.MemberCommand(m, "systemctl", "show", "--property=Result", "fleet")
+		if strings.TrimSpace(stdout) != "Result=success" {
+			return fmt.Errorf("Result for fleet unit not reported as success: %s", stdout)
+		}
+		return nil
+	}
+
+	stdout, err := cluster.MemberCommand(m0, "cat", machineIdFile)
+	if err != nil {
+		t.Fatalf("Failed to get machine-id\nstdout: %s\nerr: %v", stdout, err)
+	}
+	m0_machine_id := strings.TrimSpace(stdout)
+
+	// If the two machine IDs are different with each other,
+	// set the m1's ID to the same one as m0, to intentionally
+	// trigger an error case of duplication of machine ID.
+	stdout, err = cluster.MemberCommand(m1,
+		"echo", m0_machine_id, "|", "sudo", "tee", machineIdFile)
+	if err != nil {
+		t.Fatalf("Failed to replace machine-id\nstdout: %s\nerr: %v", stdout, err)
+	}
+
+	if err := restartFleetService(m1); err != nil {
+		t.Fatal(err)
+	}
+
+	// fleetd should actually be running, but failing to list machines.
+	// So we should expect a specific error after running fleetctl list-machines,
+	// like "googlapi: Error 503: fleet server unable to communicate with etcd".
+	stdout, stderr, err := cluster.Fleetctl(m1, "list-machines", "--no-legend")
+	if err != nil {
+		if !strings.Contains(err.Error(), "exit status 1") ||
+			!strings.Contains(stderr, "fleet server unable to communicate with etcd") {
+			t.Fatalf("m1: Failed to get list of machines. err: %v\nstderr: %s", err, stderr)
+		}
+		// If both conditions are satisfied, "exit status 1" and
+		// "...unable to communicate...", then it's an expected error. PASS.
+	} else {
+		t.Fatalf("m1: should get an error, but got success.\nstderr: %s", stderr)
+	}
+
+	// Trigger another test case of m0's ID getting different from m1's.
+	// Then it's expected that m0 and m1 would be working properly with distinct
+	// machine IDs, after having restarted fleet.service both on m0 and m1.
+	stdout, err = cluster.MemberCommand(m0,
+		"echo", util.NewMachineID(), "|", "sudo", "tee", machineIdFile)
+	if err != nil {
+		t.Fatalf("m0: Failed to replace machine-id\nstdout: %s\nerr: %v", stdout, err)
+	}
+
+	// Restart fleet service on m0, and see that it's still working.
+	if err := restartFleetService(m0); err != nil {
+		t.Fatal(err)
+	}
+
+	stdout, stderr, err = cluster.Fleetctl(m0, "list-machines", "--no-legend")
+	if err != nil {
+		t.Fatalf("m0: error: %v\nstdout: %s\nstderr: %s", err, stdout, stderr)
 	}
 }
--- a/functional/platform/nspawn.go
+++ b/functional/platform/nspawn.go
@@ -29,7 +29,6 @@ import (
 	"time"
 
 	"github.com/coreos/fleet/Godeps/_workspace/src/github.com/coreos/go-systemd/dbus"
-	"github.com/coreos/fleet/Godeps/_workspace/src/github.com/pborman/uuid"
 
 	"github.com/coreos/fleet/functional/util"
 )
@@ -384,14 +383,9 @@ func (nc *nspawnCluster) CreateMember() (m Member, err error) {
 	return nc.createMember(id)
 }
 
-func newMachineID() string {
-	// drop the standard separators to match systemd
-	return strings.Replace(uuid.New(), "-", "", -1)
-}
-
 func (nc *nspawnCluster) createMember(id string) (m Member, err error) {
 	nm := nspawnMember{
-		uuid: newMachineID(),
+		uuid: util.NewMachineID(),
 		id:   id,
 		ip:   fmt.Sprintf("172.18.1.%s", id),
 	}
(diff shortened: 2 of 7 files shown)


functional/node_test.go
```diff
@@ -82,3 +82,76 @@ func TestNodeShutdown(t *testing.T) {
...
+	stdout, err = cluster.MemberCommand(m1, "cat", machineIdFile)
+	if err != nil {
+		t.Fatalf("Failed to get machine-id\nstdout: %s\nerr: %v", stdout, err)
+	}
+	m1_machine_id := strings.TrimSpace(stdout)
+
+	// If the two machine IDs are different with each other,
+	// set the m1's ID to the same one as m0, to intentionally
+	// trigger an error case of duplication of machine ID.
+	if strings.Compare(m0_machine_id, m1_machine_id) != 0 {
```
antrik: Just set the ID of m1 unconditionally -- no need to get and compare it first...

functional/node_test.go
```diff
@@ -82,3 +82,76 @@ func TestNodeShutdown(t *testing.T) {
...
+		stdout, err = cluster.MemberCommand(m1,
+			"echo", m0_machine_id, "|", "sudo", "tee", machineIdFile)
+		if err != nil {
+			t.Fatalf("Failed to replace machine-id\nstdout: %s\nerr: %v", stdout, err)
+		}
+	}
+
+	stdout, err = cluster.MemberCommand(m1, "sudo", "systemctl", "restart", "fleet.service")
+	if err != nil {
+		t.Fatalf("Failed to replace machine-id\nstdout: %s\nerr: %v", stdout, err)
```
antrik: I think you forgot to adapt the error message here :-)

functional/node_test.go
```diff
@@ -82,3 +82,76 @@ func TestNodeShutdown(t *testing.T) {
...
+	}
+	stdout, _ = cluster.MemberCommand(m1, "systemctl", "show", "--property=Result", "fleet")
+	if strings.TrimSpace(stdout) != "Result=success" {
+		t.Fatalf("Result for fleet unit not reported as success: %s", stdout)
+	}
+
+	// expect an error after running fleetctl list-machines
+	stdout, stderr, err := cluster.Fleetctl(m1, "list-machines", "--no-legend")
+	if err != nil {
+		t.Skipf("Got an expected error: %v\nstdout: %s\nstderr: %s", err, stdout, stderr)
```
antrik: Just proceed if there is *no* error here -- it's the expected success condition.

Could you check for the expected error more specifically, though, rather than just any error?

functional/node_test.go
```diff
@@ -82,3 +82,76 @@ func TestNodeShutdown(t *testing.T) {
...
+	}
+
+	stdout, _ = cluster.MemberCommand(m1, "systemctl", "show", "--property=ActiveState", "fleet")
+	if strings.TrimSpace(stdout) != "ActiveState=active" {
+		t.Fatalf("Fleet unit not reported as active: %s", stdout)
+	}
+	stdout, _ = cluster.MemberCommand(m1, "systemctl", "show", "--property=Result", "fleet")
+	if strings.TrimSpace(stdout) != "Result=success" {
+		t.Fatalf("Result for fleet unit not reported as success: %s", stdout)
+	}
```
antrik: So fleet should actually be running, but failing to perform any operations? I guess that makes sense in a way (the conflicting machine might go away later) -- but it's a bit surprising at first. Maybe add a comment explaining this?

(I wonder, is this behaviour actually mentioned in the fleetd documentation?...)

antrik:
@dongsupark "suggested by antrik" is a bit confusing: I only suggested how to perform the functional test -- not the original fix :-)

functional/node_test.go
tixxdz: @antrik: yes as discussed I'm surprised too.. could you please check the documentation and open a PR for that one ?

functional/node_test.go
```diff
@@ -82,3 +82,76 @@ func TestNodeShutdown(t *testing.T) {
...
+
+	// expect an error after running fleetctl list-machines
+	stdout, stderr, err := cluster.Fleetctl(m1, "list-machines", "--no-legend")
+	if err != nil {
+		t.Skipf("Got an expected error: %v\nstdout: %s\nstderr: %s", err, stdout, stderr)
+	} else {
+		t.Fatalf("Error is expected, but got success.\nstderr: %s", stderr)
+	}
+}
```
tixxdz: Looks promising! so on top of all the comments, it would be nice to add after this the following:

stdout, err = cluster.MemberCommand(m0, "echo", new_uniq_machine_id, "|", "sudo", "tee", machineIdFile)
on machine 0, then restart fleet on machine 0 so we break the loop here https://github.com/coreos/fleet/pull/1561/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bR179 on machine 1 and fleet continue to works properly after of course all your previous tests.

Now if restarting fleet on m0 doesn't work, then just kill machine 0 and let machine 1 up to do its stuff register its self cleanly and follow up with a last `stdout, stderr, err := cluster.Fleetctl(m1, "list-machines", "--no-legend")` ?

Thank you!

dongsupark:
Updated
* In ``TestDetectMachineId``, add more test cases of m0's ID getting different from m1's. In that case, m0 and m1 are expected to work gracefully.
* From Cluster interface, export ``NewMachineID()`` to be used for functional tests.
* Split ``"restart fleet + systemctl show"`` calls into a separate function ``restartFleetService()`` 
* Remove an unnecessary comparison between m0 and m1 before setting the same machine IDs.
* When intentionally triggering an error, handle error of ``"fleetctl list-machines"`` more precisely.
* Fixed comments

Thanks!

functional/node_test.go
```diff
@@ -82,3 +82,119 @@ func TestNodeShutdown(t *testing.T) {
...
+	// So we should expect a specific error after running fleetctl list-machines,
+	// like "googlapi: Error 503: fleet server unable to communicate with etcd".
+	stdout, stderr, err := cluster.Fleetctl(m1, "list-machines", "--no-legend")
+	if err != nil {
+		if strings.Contains(err.Error(), "exit status 1") &&
+			strings.Contains(stderr, "fleet server unable to communicate with etcd") {
+			// This is an expected error. PASS.
+		} else {
+			t.Fatalf("m1: Failed to get list of machines. err: %v\nstderr: %s", err, stderr)
+		}
```
antrik: Empty `then` blocks are usually considered bad style... Just reverse the condition :-)

(Keep the comment though to make things clear.)

Alternatively, you could merge the inner and outer if conditions into an if + elseif construct... Not sure which approach is more readable in this case.

dongsupark:
It looks like this PR somehow causes ``TestSingleNodeConnectivityLoss`` to fail. Of course it's hard to say, why the unrelated test fails after having fixed a bug w.r.t. the machine ID.
I suspect a missing commit which will be implemented in a PR #1563 has something to do with the error. Will investigate.

functional/node_test.go
```diff
@@ -82,3 +82,119 @@ func TestNodeShutdown(t *testing.T) {
...
+	} else {
+		t.Fatalf("m1: should get an error, but got success.\nstderr: %s", stderr)
+	}
+
+	// Trigger another test case of m0's ID getting different from m1's.
+	// Then it's expected that m0 and m1 would be working properly with distinct
+	// machine IDs, after having restarted fleet.service both on m0 and m1.
```
antrik: Why do you want to restart on both? AIUI, changing the ID of m0 and restarting this one should be sufficient to resolve the conflict, so m1 should automatically become functional as well?...

functional/node_test.go
```diff
@@ -82,3 +82,119 @@ func TestNodeShutdown(t *testing.T) {
...
+	stdout, err = cluster.MemberCommand(m0,
+		"echo", cluster.NewMachineID(), "|", "sudo", "tee", machineIdFile)
+	if err != nil {
+		t.Fatalf("m0: Failed to replace machine-id\nstdout: %s\nerr: %v", stdout, err)
+	}
+
+	// Restart fleet service on m0.
+	// Even in an error case, m1 should still work.
+	// So destroy m0 and continue testing on m1.
```
antrik: I don't see a point in continuing the tests once something unexpected happens... Restarting m0 should just work?

antrik:
@dongsupark is the failure reproducible or intermittent?

dongsupark:
> is the failure reproducible or intermittent?

It's almost always reproducible. A persistent regression.
This regression has nothing to do with the functional test. Culprit is only the fleetd fix.

functional/node_test.go
tixxdz: @dongsupark could you update on top of your patch with the following
https://gist.github.com/tixxdz/14fbe081594613ba17209900b6052097

?

If it works for you then just let us do it this way, and continue to investigate with the other issue of TestSingleNodeConnectivityLoss  test.

Thanks!

dongsupark:
I think what's happening is like that:

* ``TestSingleNodeConnectivityLoss`` tries to block connections to etcd for a while, and to recover the connections again, to see that fleetd gets reattached to etcd. This mechanism is based on fleetd monitor, which invokes ``Server.run()`` in the restart context, immediately after everything gets ready. Then ``Server.run()`` first of all enters a heartbeat loop.
* Before this PR, in the restart context, the heartbeat loop only called ``Heart.Beat()``. That had succeeded in a short period, as ``Beat()`` already handled both cases even if the machine was already registered.
* With the fleetd fix in this PR, however, the heartbeat loop calls ``Heart.Register()`` that returns always an error like ``"key already exists"``, because the machine was already registered. Then it retries in the heartbeat loop, retrying the registration. That's normally harmless, as the monitor is supposed to kick it out of the loop after timeout expires. However, it's unfortunate that the functional test cannot wait for so long time until it recovers.

Solution could be:

1. First try to ``Register()`` the machine, and ``Beat()`` starting from the second attempt. This is done by #1563. But now I think it's sub-optimal, as it fails at the first attempt anyway.
2. Distinguish the restart context from the normal start context. Then call ``Register()`` only for the start context, while calling ``Beat()`` for the restart context.

I'll update this PR according to the solution 2.

dongsupark:
Updated.
* Improved the heartbeat logic to distinguish the restart context from other ones.
* Updated the functional test as suggested.

antrik:
@dongsupark well, if it's really just a timeout issue, we could just increase the timeout in the test case by another TTL length or so...

Regardless, we need to think about whether it's more correct for restart to try a new registration or update the existing one. I tend to agree that updating (as you implemented it now) is probably better -- but I haven't thought much about it...

functional/platform/nspawn.go
```diff
@@ -384,14 +384,14 @@ func (nc *nspawnCluster) CreateMember() (m Member, err error) {
 	return nc.createMember(id)
 }
 
-func newMachineID() string {
+func (nc *nspawnCluster) NewMachineID() string {
```
antrik: Why did you turn it into a method on `nc`, if it doesn't need `nc` for anything?...

functional/node_test.go
```diff
@@ -82,3 +82,97 @@ func TestNodeShutdown(t *testing.T) {
...
+	machines, err := cluster.WaitForNMachines(m1, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// m1 should have get the first ID now
+	if machines[0] != m0_machine_id {
+		t.Fatalf("Error: m1 failed to register its self with the previous ID of m0: %s", m0_machine_id)
+	}
+}
```
antrik: I think there was a bit of a misunderstanding here: I actually do think it was good to test that both m0 and m1 can serve fleet commands after the ID change and restart -- it was only the part about restarting m1 after the ID change that was unnecessary in your original test...

dongsupark:
> well, if it's really just a timeout issue, we could just increase the timeout in the test case by another TTL length or so...

Maybe yes, we could solve it by tuning timeout. But as far as I have observed, the original change in fleetd affected not only the connectivity test, but other tests. Not always, but sometimes. So I cannot prove it right now. Though I'm not surprised, as it's not the first time for me to see occasional random failures in functional tests.
Thus I tend to rely on a fix in fleetd, instead of some tuning in functional tests. Of course that's only if my fix has no logical errors or any other downsides.

> Regardless, we need to think about whether it's more correct for restart to try a new registration or update the existing one. I tend to agree that updating (as you implemented it now) is probably better -- but I haven't thought much about it...

Updating through ``Beat()`` is better, as that already performs conditional actions, "create or update".

functional/platform/nspawn.go
dongsupark: > Why did you turn it into a method on nc, if it doesn't need nc for anything?

No reason. I just thought that was an ideal way for a method to be exported from the Cluster interface. Of course we could make it a general helper in functional/util or whatever.

antrik:
@dongsupark well, I didn't mean to say that working around by increasing the timeout is necessarily a good idea -- just pointing out that it would be an option if it was the *only* problem... But if it uncovers a real problem, it's better to solve the real problem of course :-)

As for the create vs update question, I was just wondering whether a reload should be treated more like a fresh start, i.e. trying to acquire the machine ID again. But that of course can't work if the old entry isn't purged from the registry yet -- so as I already said, your current approach is *probably* correct :-)

functional/platform/nspawn.go
antrik: Oh, right, this *might* be specific to a certain cluster type... I was just thinking of exporting is as a function from where it is, but that obviously would conflict with other implementations. Sorry for the noise.

tixxdz:
lgtm, thank you!

PR #1562, opened 2016-04-19
functional: silence go vet reports on functional test

tixxdz:


antrik:
I'm confused: why did you open the PR, and then close it again without comment?...

Anyway, I think "silence" is not the right word to use here: as far as I can see you actually fixed the issues, rather then just silencing the report?...

Issue #750, opened 2014-08-06
Monitor heartbeat attempt can time out, but succeed


bcwaldon:
An attempt to heartbeat machine presence can succeed, but timeout client-side. This causes all subsequent attempts to fail as the presence is created using prevExist=false.

```
Aug 06 06:43:21 node-02 fleet[668]: I0806 06:43:21.360308 00668 monitor.go:54] Monitor heartbeat function returned err, retrying in 2.5s: timeout reached
…
Aug 06 06:43:25 node-02 fleet[668]: I0806 06:43:25.354149 00668 monitor.go:54] Monitor heartbeat function returned err, retrying in 2.5s: 105: Key already exists (/_coreos.com/fleet/machines/00b7e1bbb2464ed9a4ef4282b11f2161/object)
```

wuqixuan:
@jonboulle  @bcwaldon  It's not totally same as #615. #615 is prevent other machine with the same id. This is for allow the existing same id when you second time heartbeat. 
So heartbeat code should cover the two case.

wuqixuan:
There are 3 case:

kayrus:
Does anyone know how to reproduce this issue?

dongsupark:
> 1. prevent other machine with the same id (#615)
2. allow the existing same id when you second time heartbeat ( #750)
3. prevent another daemon in the same host. 

The 1st one, preventing other machines from being registered with the same ID, was addressed by #1561.
I'm not sure about what we can do about the 3rd one, preventing another daemon in the same host.

Let me get the 2nd one straight, ``"allowing the existing same id when doing heartbeat for the second time"``. AFAIK no PR addressed this issue.
Could it be simply done, if we would do ``s.hrt.Register()`` only for the 1st attempt, and do ``s.hrt.Beat()`` starting from the 2nd attempt? Then this issue could be relatively simply addressed.

Though I'm not sure if this is supposed to be done inside the milestone v0.13.

antrik:
At first I thought 3) would be addressed by the fix for 1) too -- but I realised now that it is indeed a separate issue: if you start two fleetds on the same machine, they will be "fighting" for the machine ID... Not sure this is really a problem that needs to be fixed though -- at least not an urgent one.

Anyway, this issue here is clearly about case 2.

dongsupark:
> 1. prevent other machine with the same id (#615)
> 2. allow the existing same id when you second time heartbeat ( #750)
> 3. prevent another daemon in the same host.

[...]

PR #1563, opened 2016-04-20
[WIP] fleetd: register heartbeat only for the first attempt

dongsupark:
When beating the Heart for the 1st time, call ``Heart.Register()`` to avoid such a case of registering machine with the same ID. Starting from the next heartbeat, however, call ``Heart.Beat()`` to allow registration with the same ID. That way fleetd can handle the machine presence in a graceful way.
    
Suggested-by: @wuqixuan 
Fixes: https://github.com/coreos/fleet/issues/750

This PR depends on #1561.
NOTE: this is still experimental. do not merge it yet please!

### DIFF
--- a/functional/node_test.go
+++ b/functional/node_test.go
@@ -82,3 +82,119 @@ func TestNodeShutdown(t *testing.T) {
 		t.Fatalf("Unit hello.service not reported as inactive:\n%s\n", stdout)
 	}
 }
+
+// TestDetectMachineId checks for etcd registration failing on a duplicated
+// machine-id on different machines.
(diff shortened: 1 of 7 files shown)


antrik:
@dongsupark this sounds wrong to me: doesn't it mean that the followup attempts will "steal" the ID from another instance if there is a conflict?...

(Haven't really thought it through in detail though -- so I might be something here...)

dongsupark:
A bugfix for a regression in fleetd was already included in #1561. Anyway this PR #1563 does not seem to correctly fix the original issue.
Closing.

Issue #1025, opened 2014-11-14
fleetctl start != systemctl start


bcwaldon:
The `fleetctl start` command sets the desired state of a unit to "launched" (that's a fleet state, not a systemd state) and waits for the unit to be scheduled and the target machine to publish systemd state. fleetctl does not look at the contents of this systemd state.

The `systemctl start` command acts more like an RPC command, sending a `StartUnit` message over dbus and waiting for the created systemd job to finish.

We have a clear usability problem here. Users expect `fleetctl start` to be equivalent to `systemctl start` (#1019). This inconsistency is also preventing us from moving forward on adding a `fleetctl restart` command, which should reasonably act like `systemctl restart` (#961).

Assuming others agree that this inconsistency is a problem, I see only one real path forward. `fleetctl start` needs to watch for the unit to actually start successfuly in systemd.

chrisfarms:
:+1:

anpieber:
I'm also a big :+1:  on this one to finally get the inconsistency solved. Anything we can provide to help with this one?

xied75:
:+1:

PR #1566, opened 2016-04-25
fleetctl: check also systemd states after starting units

dongsupark:
To resolve inconsistency between ``"systemctl start"`` and ``"fleetctl start"``, fleetctl needs to check systemd states after starting units. If the systemd state is not a desired one, wait until it gets an active state.

Fixes: https://github.com/coreos/fleet/issues/1025

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -33,6 +33,7 @@ import (
 	"github.com/spf13/cobra"
 
 	etcd "github.com/coreos/etcd/client"
+	sd_dbus "github.com/coreos/go-systemd/dbus"
 
 	"github.com/coreos/fleet/api"
 	"github.com/coreos/fleet/client"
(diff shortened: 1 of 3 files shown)


dongsupark:
Updated. (after several hiccups)
* Handle return value of ``tryWaitForSystemdActiveState`` carefully. It should not return 1 only because of a missing systemd dbus connection. Otherwise it will break unit tests based on Travis, as that environment cannot correctly set up a systemd dbus connection.
* Added a missing nil checking before ``conn.Close()``. Also call ``conn.Close()`` after checking for nil.
* Wrote comments etc.

dongsupark:
I'm going to merge it soon, unless anyone has other opinions.

fleetctl/fleetctl.go
```diff
@@ -1060,6 +1061,87 @@ func assertUnitState(name string, js job.JobState, out io.Writer) (ret bool) {
...
+// active state, making use of the systemd's DBUS interface. First it takes one
+// or more units as input, and ensures that every unit in the []units must be
+// in the active state. If yes, return 0. Otherwise return 1.
+//
+// NOTE: To avoid unnecessary conflicts with unit tests in start_test.go, we
+// should return 0 for the case of sd_dbus.New() returning error. In that case,
+// it just means that the unit test environment based on Travis is not capable
+// of setting up a DBUS connection of systemd, which would actually work fine
+// for production systems or functional tests.
+func tryWaitForSystemdActiveState(units []string) (ret int) {
```
jonboulle: imho the ret value is pretty unidiomatic; rather, return an error, and have the callers decide to log the error or not

fleetctl/fleetctl.go
```diff
@@ -1060,6 +1061,87 @@ func assertUnitState(name string, js job.JobState, out io.Writer) (ret bool) {
...
+	conn, err := sd_dbus.New()
+	if err != nil {
+		stderr("Failed to get systemd-dbus conn: %v", err)
+		return 0
+	}
+	if conn == nil {
```
jonboulle: can this happen?

fleetctl/fleetctl.go
```diff
@@ -1060,6 +1061,87 @@ func assertUnitState(name string, js job.JobState, out io.Writer) (ret bool) {
...
+	for attempt := 0; attempt < maxAttempts; attempt++ {
+		found, errAssert = assertSystemdActiveState(conn, unitName)
+		if found && errAssert == nil {
+			break
+		}
+		time.Sleep(sleep)
+	}
+
+	if found {
+		err = nil
```
jonboulle: unnecessary

fleetctl/fleetctl.go
```diff
@@ -1060,6 +1061,87 @@ func assertUnitState(name string, js job.JobState, out io.Writer) (ret bool) {
...
+// assertSystemdActiveState determines if a given systemd unit is actually
+// in the active state, making use of systemd's DBUS connection.
+func assertSystemdActiveState(conn *sd_dbus.Conn, unitName string) (found bool, err error) {
+	listUnits, err := conn.ListUnits()
+	if err != nil {
+		return false, fmt.Errorf("Failed to list systemd unit: %v", err)
+	}
+
+	for _, u := range listUnits {
+		if strings.Compare(u.Name, unitName) != 0 {
```
jonboulle: == works fine for strings

fleetctl/start.go
```diff
@@ -87,5 +87,10 @@ func runStartUnit(cCmd *cobra.Command, args []string) (exit int) {
 		return exitVal
 	}
 
+	exitVal = tryWaitForSystemdActiveState(starting)
```
jonboulle: This needs to consider block-attempts and no-block flags right?

jonboulle:
I think this is wrong; it's checking against a local systemd instance over dbus, but that will ONLY be right if `fleetctl` happens to be invoked on the same system on which the units are scheduled, right? Surely we need to check against the API, as we do with the other states?

fleetctl/fleetctl.go
dongsupark: Agreed. I'll try to make it return an error.

fleetctl/fleetctl.go
dongsupark: Correct. I'll just remove `err = nil`.

fleetctl/fleetctl.go
dongsupark: Right. I'll just use `==`.

fleetctl/start.go
dongsupark: Geez, you're right. I'll pass `getBlockAttempts(cCmd)` to that function to check the flags, just like `tryWaitForUnitStates()`.

dongsupark:
@jonboulle Ok. I'll try to take a different approach as you suggested. From my understanding, `schema.UnitState` has already its own fields like `SystemdActiveState`, `SystemdLoadState`, and `SystemdSubState`. So we should make use of `cAPI.UnitStates()` to retrieve these states, to make sure that these are the expected states. Is that correct?

jonboulle:
Yep, that sounds about right!

PR #1569, opened 2016-04-26
functional: reduce time spent during unit actions

dongsupark:
To reduce time spent during unit actions, especially ``"fleetctl {load,start} --replace"``, do some optimizations:

* combine ``replaceUnitCommon`` with ``replaceUnitMultiple``
* refactor ``replaceUnitCommon`` to improve test speed
* refactor ``unitStartCommon``
* reduce number of replaced units for tests

### DIFF
--- a/functional/unit_action_test.go
+++ b/functional/unit_action_test.go
@@ -35,6 +35,12 @@ const (
 	numUnitsReplace = 9
 )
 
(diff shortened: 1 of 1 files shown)


tixxdz:
lgtm thanks!

Issue #1453, opened 2016-03-02
usability: fleetctl / fleetd help output with double hyphens


kayrus:
According to this [issue](https://github.com/coreos/docs/issues/650) `fleetctl -%someinvalidcommand%` should print commands list with the double hyphens.

```sh
$ fleetd -foo
flag provided but not defined: -foo
fleetd is the server component of fleet, a simple orchestration system for scheduling systemd units in a cluster.
Usage of fleetd:
  -config="": Path to config file. Fleet will look for a config at /etc/fleet/fleet.conf by default.
  -version=false: Print the version and exit
$ fleetctl --foo
flag provided but not defined: -foo
Usage of fleetctl:
  -ca-file="": Location of TLS CA file used to secure communication with the fleet API or etcd
  -cert-file="": Location of TLS cert file used to secure communication with the fleet API or etcd
  -debug=false: Print out more debug information to stderr
  -driver="API": Adapter used to execute fleetctl commands. Options include "API" and "etcd".
  -endpoint="unix:///var/run/fleet.sock": Location of the fleet API if --driver=API. Alternatively, if --driver=etcd, location of the etcd API.
  -etcd-cafile="": hidden
  -etcd-certfile="": hidden
  -etcd-key-prefix="/_coreos.com/fleet/": Keyspace for fleet data in etcd (development use only!)
  -etcd-keyfile="": hidden
  -experimental-api=true: hidden
  -h=false: Print usage information and exit
  -help=false: Print usage information and exit
  -key-file="": Location of TLS key file used to secure communication with the fleet API or etcd
  -known-hosts-file="~/.fleetctl/known_hosts": File used to store remote machine fingerprints. Ignored if strict host key checking is disabled.
  -request-timeout=3: Amount of time in seconds to allow a single request before considering it failed.
  -ssh-timeout=10: Amount of time in seconds to allow for SSH connection initialization before failing.
  -ssh-username="core": Username to use when connecting to CoreOS instance.
  -strict-host-key-checking=true: Verify host keys presented by remote machines before initiating SSH connections.
  -tunnel="": Establish an SSH tunnel through the provided address for communication with fleet and etcd.
  -version=false: Print the version and exit
```
relates to #1445 https://github.com/coreos/etcd/issues/4595

kayrus:
It is also better to unify `fleetctl --help` and `fleetctl --foo` commands output.

kayrus:
@reneengelhard same issue with subhelp:

```sh
[...]

PR #1570, opened 2016-04-26
Move to cli

reneengelhard:
Hi,

while discussing #1453 and the implications of fixing it we thought it would make sense to move from our custom flag parsing to use 'cli' -  as etcd already does.

Due to some difficulties and big(ger) restructuring needs of (test) code it took longer than anticipated, but here it is.

Passes the unit and functional tests. I rebased it upto current master so it includes the new
--replace switch, too

Regards,

Rene


tixxdz:
@reneengelhard @antrik thanks for the work!

So if tests pass, we need some lgtm here, will review it later. If of course there are no backward incompatibilities this should go in.

Thanks again!

jonboulle:
nack, if you're going to move to a library can you please make it cobra
instead? https://github.com/coreos/docs/tree/master/golang#cli

We used to use codegangsta in fleet and moved away from it. If you look
into etcd a little deeper, it should only be using cli for some old
commands, and all the latest stuff should be using cobra.

tixxdz:
@jonboulle just in time, alright.

Issue #1394, opened 2015-11-27
New fleet-specific Option "Replaces" (feature)


onlyjob:
Using Fleet I'm trying to implement high-availability solution for a certain daemon which does  master-slave replication. There are two instances - "master" and "shadow master". When "master" is down, "shadow master" can be promoted to "master" which is trivial to implement with Fleet. However the problem that is difficult to solve is re-starting "shadow master" somewhere else after "shadow master" to "master" fail-over. 

A new fleet-specific option "Replaces" could be helpful in this scenario. Unlike stronger "Conflicts", I suppose "Replaces" could force re-scheduling of another unit to another machine. For example, "master.service" could contain

    [X-Fleet]
    Replaces=shadow.service

so when "master.service" is started on machine where "shadow.service" is running the latter is stopped and automatically re-scheduled to run on another machine.

I hope that such option can be considered useful.

scipetr:
+1

antrik:
I suggest calling it "Displaces": I think that would make it much clearer what it does...

Regarding the functionality, I guess it should work pretty much like "Conflicts", except that it is directional -- i.e. it's always the "displaced" unit that gets moved, never the "displacing" one?

kayrus:
What about taking RPM's dependency name? [Obsoletes](http://www.rpm.org/wiki/PackagerDocs/Dependencies#Obsoletes)

kayrus:
@onlyjob what if `shadow.service` will contain:

```ini
[X-Fleet]
Replaces=master.service
```

How would fleet handle that? Weights? But weights should be implemented by the clustered software itself like it's done in [keepalived](http://www.keepalived.org/LVS-NAT-Keepalived-HOWTO.html).

The workaround for the current fleet version is to combine systemd's [`ExecStartPre=`](https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStartPre=) and [`StartTimeout=0`](https://www.freedesktop.org/software/systemd/man/systemd.service.html#TimeoutStartSec=) inside global master unit and [sidekick](https://coreos.com/fleet/docs/latest/examples/service-discovery.html) for them.

You'll have to implement health check logic using script which will check etcd registry and set keys with the TTL.

ExecStartPre will not allow unit to start until etcd slot is empty. If slot is empty, systemd will start the unit with sidekick. Sidekick will update key's TTL every second.

antrik:
[...]

PR #1572, opened 2016-04-29
fleetd: introduce Replaces option in unit files

dongsupark:
Introduce a new option ``'Replaces'`` in X-Fleet the section, to get an existing unit rescheduled to another machine. This option basically works like an existing option ``'Conflicts'``, except that circular replaces in 2 units are not allowed. Also ``Conflicts`` cannot be used together with ``Replaces``, ``Global`` cannot be used together with ``Replaces``.

Fixes: https://github.com/coreos/fleet/issues/1394

/cc @onlyjob @kayrus @tixxdz


dongsupark:
Added more unit tests.

Note that I'm self-reviewing this PR more carefully than other minor things, as the original issue is marked as milestone 0.13. As it touches fleetd, we should be a little more careful. As I don't have an access to add github labels, I'm writing it as comments.

dongsupark:
Updated documents and functional tests.

onlyjob:
Thanks!

PR #1573, opened 2016-04-29
functional: retry to check states also after connectivity got alive

dongsupark:
``TestSingleNodeConnectivityLoss`` sometimes fails with the following error:

>  FAIL: TestSingleNodeConnectivityLoss (26.50s)
>  connectivity-loss_test.go:98: Failed listing unit files: exit status 1
>  connectivity-loss_test.go:210: Units not in expected state after restoring connectivity.

This is because the test runs ``checkExpectedStates()`` only once, without retrying to check the states repeatedly. While the test had already made use of ``util.WaitForState()`` earlier in the test by itself, it actually does not call ``util.WaitForState`` after being reattached to etcd. Fix it by using ``util.WaitForState()`` correctly to avoid such occasional errors, especially on semaphoreci.

/cc @tixxdz @kayrus

### DIFF
--- a/functional/connectivity-loss_test.go
+++ b/functional/connectivity-loss_test.go
@@ -206,7 +206,15 @@ func TestSingleNodeConnectivityLoss(t *testing.T) {
 	time.Sleep(ttl + agentReconcileInterval + slack)
 
 	// Check state after reconnect.
-	if isExpected, expected, actual := checkExpectedStates(); !isExpected {
+	var expected, actual map[string]string
+	var isExpected bool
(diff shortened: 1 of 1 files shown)


dongsupark:
ok, the 1st build failed, but not at ``TestSingleNodeConnectivityLoss``, but at ``TestReplaceSerialization``. That must be a different issue.

dongsupark:
All right, the 2nd build passed. I hope this fix could help.

kayrus:
ltgm

Issue #1432, opened 2016-02-15
fleetctl: stop command doesn't show correct message


kayrus:
Here is how you can reproduce that. Unit file:

```ini
[Service]
ExecStart=/usr/bin/true
Type=oneshot
```

```sh
$ fleetctl start app.service
Unit app.service inactive
Unit app.service launched on 0eb1b9b6.../coreos1
```

```sh
$ fleetctl status
● app.service
   Loaded: loaded (/run/fleet/units/app.service; linked-runtime; vendor preset: disabled)
   Active: inactive (dead)

Feb 15 15:24:33 coreos1 systemd[1]: Starting app.service...
Feb 15 15:24:33 coreos1 systemd[1]: Started app.service.
```

```sh
$ fleetctl stop app.service
Unit app.service loaded on 0eb1b9b6.../coreos1
```

fleetctl stops unit, but shows incorrect message `Unit app.service loaded on 0eb1b9b6.../coreos1`.

tixxdz:
When you stop a unit I guess it tries to return to the "JobStateLoaded" state which is defined here https://github.com/coreos/fleet/blob/master/job/job.go#L29 to "loaded" string. All the logic is printed here https://github.com/coreos/fleet/blob/master/fleetctl/fleetctl.go#L820

So in this branch https://github.com/endocode/fleet/tree/tixxdz/fleet-code-cleaning-v1 we can pass the correct string state down to these callers without adding or updating JobStates.

dongsupark:
Closed via https://github.com/coreos/fleet/issues/1577.

PR #1577, opened 2016-05-02
fleetctl: print out result message explicitly

dongsupark:
``Fleetctl stop`` or ``unload`` has been always printing out somewhat ambiguous messages like below:

```
  Unit app.service loaded on 0eb1b9b6.../coreos1
```

For better usability, print out a better result message, either about stopping (or unloading) successfully, or failing to do.

Partially resolves: https://github.com/coreos/fleet/issues/1432

/cc @kayrus @tixxdz

### DIFF
--- a/fleetctl/stop.go
+++ b/fleetctl/stop.go
@@ -85,6 +85,11 @@ func runStopUnit(args []string) (exit int) {
 	}
 
 	exit = tryWaitForUnitStates(stopping, "stop", job.JobStateLoaded, getBlockAttempts(), os.Stdout)
+	if exit == 0 {
+		stderr("Successfully stopped units %v.", stopping)
+	} else {
+		stderr("Failed to stop units %v. exit == %d.", stopping, exit)
+	}
 
 	return
 }
--- a/fleetctl/unload.go
+++ b/fleetctl/unload.go
@@ -66,6 +66,11 @@ func runUnloadUnit(args []string) (exit int) {
 	}
 
 	exit = tryWaitForUnitStates(wait, "unload", job.JobStateInactive, getBlockAttempts(), os.Stdout)
+	if exit == 0 {
+		stderr("Successfully unloaded units %v.", wait)
+	} else {
+		stderr("Failed to unload units %v. exit == %d.", wait, exit)
+	}
 
 	return
 }


dongsupark:
Updated, to print friendly messages for both stop and unload.

fleetctl/stop.go
```diff
@@ -85,6 +85,11 @@ func runStopUnit(args []string) (exit int) {
 	}
 
 	exit = tryWaitForUnitStates(stopping, "stop", job.JobStateLoaded, getBlockAttempts(), os.Stdout)
+	if exit == 0 {
+		stdout("Successfully stopped units %v.", stopping)
```
jonboulle: Why is this to stdout and not stderr? stdout is basically part of the contract of the output (i.e. this would become API)

dongsupark: Done, changed from stdout to stderr.

dongsupark:
Merged https://github.com/coreos/fleet/pull/1577. Thanks.

fleetctl/stop.go
```diff
@@ -85,6 +85,11 @@ func runStopUnit(args []string) (exit int) {
 	}
 
 	exit = tryWaitForUnitStates(stopping, "stop", job.JobStateLoaded, getBlockAttempts(), os.Stdout)
+	if exit == 0 {
+		stderr("Successfully stopped units %v.", stopping)
+	} else {
+		stderr("Failed to stop units %v. exit == %d.", stopping, exit)
```
jonboulle: My apologies for missing this previously. Why do we print the return value here? We already exit the process with that value so it can be easily inspected. Could you remove that from the string, please?

fleetctl/unload.go
```diff
@@ -66,6 +66,11 @@ func runUnloadUnit(args []string) (exit int) {
 	}
 
 	exit = tryWaitForUnitStates(wait, "unload", job.JobStateInactive, getBlockAttempts(), os.Stdout)
+	if exit == 0 {
+		stderr("Successfully unloaded units %v.", wait)
+	} else {
+		stderr("Failed to unload units %v. exit == %d.", wait, exit)
```
jonboulle: Here too.

PR #1578, opened 2016-05-03
functional: download CoreOS over HTTPS

marineam:


kayrus:
Did we give up with the GPG keys?

tixxdz:
lgtm, thanks!

marineam:
@kayrus no, gpg keys are still used. I'm just updating URLs everyone uses.

Issue #1397, opened 2015-12-17
Need way to set HeaderTimeoutPerRequest


themicster:
I noticed there is a new config option for the etcd client library. I think I need this to prevent fleet from reconciling and restarting my services when an etcd server goes down. I need to set HeaderTimeoutPerRequest to 1 second so that fleet will have time to try another etcd server before reconcile thinks the machine is down because it can't heartbeat in time.

themicster:
I found in the sample config file for fleet there is etcd_request_timeout=1.0, this may work. What is the default if I don't set this config parameter?

themicster:
Also can I set etcd_request_timeout from cloud config environment variables like this: FLEET_ETCD_REQUEST_TIMEOUT

themicster:
It looks like the default is 1.0 and this is in fleet 0.11.5 which is what is in CoreOS 835.9.0. So this isn't helping. I've done testing by dropping all traffic to and from one of my etcd machines to simulate it not being there, but fleet still reconciles and restarts services. I have 3 etcd machines and 5 etcd proxy compute nodes. My services run on the proxy machines. Bringing down one etcd server should not cause any failures or restarts of services except the ones on the server that is down, but it reschedules the services on the machines that are up because it seems like those machines can't heartbeat to etcd. Pinning helps make this happen less often but there is still a chance every 10 seconds I think that it marks the etcd server as available again until it finds out its down again, and it could be too late by this time.

jonboulle:
I think the behavioural issues here are captured in #708 and #715: really the agent needs to be more resilient to connectivity losses.

However you touch on another point which is now that this is available in the etcd client we can potentially [stop doing it in fleet itself](https://github.com/coreos/fleet/blob/ad01314ebd2e151c1202a2a247fcfdad5ef1798c/registry/etcd.go#L42) and instead just pass on the value to the etcd client library. I'm going to leave this open to track that clean up.

dongsupark:
Closed via https://github.com/coreos/fleet/issues/1580.
I suppose, the original issues are either resolved, or covered by other issues. If not, feel free to reopen this issue.

PR #1580, opened 2016-05-06
registry: use etcd.Config.HeaderTimeoutPerRequest instead of internal timeout

dongsupark:
As etcd's ``keysAPI`` already supports ``HeaderTimeoutPerRequest``, fleet should not handle request timeout on its own, but just pass on the timeout value to the etcd client. For that, set timeout only in ``etcd.Config.HeaderTimeoutPerRequest``, and get rid of internal timeout handling from both ``EtcdRegistry`` and ``LeaseManager``.

Suggested-by @jonboulle 
Partially resolves https://github.com/coreos/fleet/issues/1397

### DIFF
--- a/registry/etcd.go
+++ b/registry/etcd.go
@@ -16,32 +16,23 @@ package registry
 
 import (
 	"path"
-	"time"
 
 	etcd "github.com/coreos/etcd/client"
-	"golang.org/x/net/context"
 )
 
 const DefaultKeyPrefix = "/_coreos.com/fleet/"
 
-func NewEtcdRegistry(kAPI etcd.KeysAPI, keyPrefix string, reqTimeout time.Duration) *EtcdRegistry {
+func NewEtcdRegistry(kAPI etcd.KeysAPI, keyPrefix string) *EtcdRegistry {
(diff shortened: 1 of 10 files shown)


registry/etcd.go
```diff
@@ -16,32 +16,28 @@ package registry
...
 // EtcdRegistry fulfils the Registry interface and uses etcd as a backend
 type EtcdRegistry struct {
-	kAPI       etcd.KeysAPI
-	keyPrefix  string
-	reqTimeout time.Duration
+	kAPI      etcd.KeysAPI
+	keyPrefix string
 }
 
 func (r *EtcdRegistry) ctx() context.Context {
```
jonboulle: This function doesn't add much value anymore - how about just inlining context.Background where it's called instead?

jonboulle:
awesome! One minor comment, LGTM other than that

dongsupark:
@jonboulle Done. I removed ``EtcdRegistry.ctx()`` to just call ``context.Background()``
directly.

dongsupark:
Merged https://github.com/coreos/fleet/pull/1580. Thanks.

PR #1581, opened 2016-05-13
fleetctl: use cobra instead of cli

dongsupark:
Use Cobra (https://github.com/spf13/cobra) instead of cli (https://github.com/codegangsta/cli), for better cmdline user interface. Create a wrapper ``runWrapper()`` to be used for cobra, to wrap around a normal ``run*()`` function into a prototype for ``cobra.Command.Run()``. It also sets a global variable cAPI for running a normal command. Also remove unnecessary code for codegangsta/cli from fleetctl.go.

This PR consists of 6 patches.
* 1st ``"fleetctl: move to cli"`` was taken from https://github.com/coreos/fleet/pull/1570, though I had to modify that to get it working with the current vendor structure.
* 2-6 are my patches.
* The 2nd ``"vendor github.com/spf13/{cobra,pflag}"`` and 3rd ``"fleetctl: do not pass around cAPI"`` are just preparations. 
* The 4th [``"fleetctl: convert cli to cobra"``](https://github.com/coreos/fleet/pull/1581/commits/848d35616fc95d58bdd7771c60f1a7c90c747a33) actually covers most changes. As cobra affects nearly every fleetctl command, this is rather a huge change.
* 5th ``"fleetctl: use cobra also for unit tests"`` and 6th ``"functional: run correct commands for fleetctl help"`` are just minor updates for unit tests as well as functional tests.

Suggested-by: @jonboulle 
Fixes: https://github.com/coreos/fleet/issues/1453
Supersedes https://github.com/coreos/fleet/pull/1570

/cc @reneengelhard @kayrus @tixxdz


fleetctl/help.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func fleetFlagUsages(flagSet *pflag.FlagSet) string {
+	x := new(bytes.Buffer)
+
+	flagSet.VisitAll(func(flag *pflag.Flag) {
+		if len(flag.Deprecated) > 0 || flag.Hidden {
+			return
+		}
+		format := ""
```
jonboulle: really these should probably either write into the output buffer directly, or use another bytes buffer here to construct the formatting string

fleetctl/cat.go
```diff
@@ -17,24 +17,36 @@ package main
...
+	Run: runWrapper(runCatUnit),
+}
+
+func init() {
+	cmdFleet.AddCommand(cmdCat)
 }
 
-func runCatUnit(c *cli.Context, cAPI client.API) (exit int) {
-	args := c.Args()
+// GetCatCommand returns the cobra command for cat
```
jonboulle: why are these helper functions necessary? not clear to me why the tests can't just reference the run functions directly

fleetctl/journal.go
```diff
@@ -17,39 +17,57 @@ package main
...
+
+// GetJournalCommand returns the cobra command for journal
+func GetJournalCommand() *cobra.Command {
+	return cmdJournal
 }
 
-func runJournal(c *cli.Context, cAPI client.API) (exit int) {
-	args := c.Args()
+func runJournal(cCmd *cobra.Command, args []string, fakeAPI client.API) (exit int) {
+	if fakeAPI == nil {
```
jonboulle: this is an unfortunate duplication pasted everywhere - can it go within the wrapper as it used to? (e.g. formerly makeActionWrapper)

fleetctl/help.go
dongsupark: Hmm, I'm not sure why it's written like that, as I copied the help.go from rkt. I just thought this was a common pattern. Will try to fix it.

fleetctl/journal.go
dongsupark: The ``makeActionWrapper`` used for the original ``codegangsta/cli`` is now ``runWrapper`` used for cobra. Although they have different names, they are almost the same as each other, except the fakeAPI parameter.

Why does each runCommand take the fakeAPI as parameter? Because each command runner needs to distinguish a normal command context from a unit test context based on fakeAPI. I know, it looks somewhat ugly. But I couldn't come up with better way to distinguish them. I actually did go down through initializing cAPI in main() before executing commands, which didn't work, because it didn't quite well fit into the cobra's framework.

fleetctl/cat.go
dongsupark: Yeah right, it looks like ``Get*Command`` functions could be removed. Will do.

dongsupark:
Updated.
* removed ``fakeAPI`` parameters from ``runWrapper()`` and each corresponding ``run*()`` command, for simpler code. Instead, in each ``run*()`` command, determine the context by checking for a global variable cAPI. Unit tests will directly set the global cAPI to fakeAPI.
* removed ``fleetFlagUsages()``, and made it just directly call the native ``FlagUsages()`` method provided by spf13/pflag. This way we don't have to keep our own ``FlagUsages()``.
* removed ``Get*Command()`` functions, and made each call just get a cmd* variable.

dongsupark:
I managed to remove duplicated calls for setting cAPI in each command, by moving the call into runWrapper(), as @jonboulle suggested.

dongsupark:
Merged https://github.com/coreos/fleet/pull/1581.
Thanks for everyone who has been involved in this long-running issue, as well as doing extensive code reviews. If anything breaks after this, feel free to create an issue.

Issue #1532, opened 2016-04-04
tests: functional tests don't work on CoreOS 1000 release


kayrus:
looks like coreos-cloudinit is not triggered to configure the nspawn systemd machine

mischief:
might be caused by https://github.com/coreos/coreos-cloudinit/pull/419

dongsupark:
I have been also seeing this issue since CoreOS 1000. Functional test doesn't run at all. I'm still trying to understand what's going on. I suppose, in the near future we need to figure out how to get around it, before stable & beta got updated to versions higher than 1000.

So I tried to debug on coreos alpha (1032).

Fleet's functional test sets up a nspawn container under ``/tmp/smoke/_ID_/fs/``.
For example:
```
# ls /tmp/smoke/b6a70f9c5ed74d46b08a05ca648a4ba0/fs/
bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var

```

Inside the container, fleetd.conf was generated successfully as ``/tmp/smoke/_ID_/fs/media/configdrive/openstack/latest/user_data``.

```
# cat /tmp/smoke/b6a70f9c5ed74d46b08a05ca648a4ba0/fs/media/configdrive/openstack/latest/user_data
#cloud-config

write_files:
 - path: /opt/fleet/fleet.conf
   content: |
    verbosity=2
    etcd_servers=["http://172.18.0.1:4001"]
    etcd_key_prefix=/fleet_functional/smoke
    public_ip=172.18.1.1
    agent_ttl=3s

ssh_authorized_keys:
[...]

PR #1585, opened 2016-05-20
functional: fix path to user_data to /var/lib/coreos-install

dongsupark:
Since CoreOS alpha (1000), coreos-cloudinit doesn't seem to watch
on ``/media/configdrive/openstack/latest/user_data``. That's why
functional tests haven't been running, as systemd-nspawn container
runs without any configurations.
Fix it by changing into ``/var/lib/coreos-install/user_data``.

Fixes: https://github.com/coreos/fleet/issues/1532

/cc @kayrus

### DIFF
--- a/functional/platform/nspawn.go
+++ b/functional/platform/nspawn.go
@@ -344,7 +344,7 @@ func (nc *nspawnCluster) insertBin(src string, dst string) error {
 }
 
 func (nc *nspawnCluster) buildConfigDrive(dir, ip string) error {
-	latest := path.Join(dir, "media/configdrive/openstack/latest")
+	latest := path.Join(dir, "var/lib/coreos-install")
 	userPath := path.Join(latest, "user_data")
 	if err := os.MkdirAll(latest, 0755); err != nil {
 		return err


jonboulle:
I don't really understand how https://github.com/coreos/coreos-cloudinit/pull/419 is supposed to have affected this.. but if this change works, then lgtm

dongsupark:
I also don't know how fleet's functional test was originally written. Just for trying to understand the reason in the background:

* coreos-cloudinit had been watching for 2 different paths: ``/var/lib/coreos-install/user_data`` (a single file), and ``/media/configdrive`` (a non-empty directory). According to a comment, the latter seemed to have existed only for backward compatibility.
* On the other hand, for some reason, fleet has been relying on the mechanism of writing yaml config into ``/media/configdrive/openstack/latest/user_data``, which could be then processed by ``user-configdrive.path``. IMO this was a wrong approach to begin with, because inside the systemd-nspawn container, ``/media/configdrive`` is not mounted at all. Writing anything into a non-mounted directory seems to be a flaky mechanism after all.
* Then the ``user-configdrive.path`` was removed via https://github.com/coreos/coreos-cloudinit/pull/419 , so that the unit didn't get conflicted with the new behavior of systemd v229. Now coreos-cloudinit is only able to watch for ``/var/lib/coreos-install/user_data``. This change makes sense, although it suddenly broke the original expectation of fleet.

So the right fix is to change the path to user_data to the correct one, ``/var/lib/coreos-install/user_data``. I suppose, it should have been like that in the first place.

I'm going to merge it soon.

jonboulle:
SGTM

Issue #1586, opened 2016-05-23
functional: TestKnownHostsVerification fails with systemd v229 or higher


dongsupark:
Fleet's functional test ``TestKnownHostsVerification`` fails always with systemd v229 on CoreOS alpha (1032). This is a different issue from https://github.com/coreos/fleet/issues/1532.

Logs:

```
2016/05/23 12:12:26 systemctl -M smokea41e929489b64998b6b9ea0f1e8968e2 poweroff
2016/05/23 12:12:26 Destroying nspawn machine f1abee3a17ed4a0db0b9ac2686191687
2016/05/23 12:12:26 Failed tests, fetching logs from f1abee3a17ed4a0db0b9ac2686191687 machine
2016/05/23 12:12:26 journalctl --directory=/tmp/smoke/f1abee3a17ed4a0db0b9ac2686191687/fs/var/log/journal --root=/tmp/smoke/f1abee3a17ed4a0db0b9ac2686191687/fs --no-pager
2016/05/23 12:12:26 Wrote smokef1abee3a17ed4a0db0b9ac2686191687 logs to /home/core/fleet/gopath/src/github.com/coreos/fleet/functional/TestKnownHostsVerification_smokef1abee3a17ed4a0db0b9ac2686191687.log
2016/05/23 12:12:26 machinectl terminate smokef1abee3a17ed4a0db0b9ac2686191687
2016/05/23 12:12:26 rm -f /run/systemd/system/machine-smokef1abee3a17ed4a0db0b9ac2686191687.scope
2016/05/23 12:12:26 rm -f /run/systemd/system/smokef1abee3a17ed4a0db0b9ac2686191687.service
2016/05/23 12:12:26 rm -fr /run/systemd/system/smokef1abee3a17ed4a0db0b9ac2686191687.service.d
2016/05/23 12:12:26 rm -r /tmp/smoke/f1abee3a17ed4a0db0b9ac2686191687
2016/05/23 12:12:26 ip link del vb-smokef1abee3a17ed4a0db0b9ac2686191687
2016/05/23 12:12:26 Destroying nspawn machine a41e929489b64998b6b9ea0f1e8968e2
2016/05/23 12:12:26 Failed tests, fetching logs from a41e929489b64998b6b9ea0f1e8968e2 machine
[...]

PR #1587, opened 2016-05-23
functional: poweroff machine using machinectl instead of systemctl

dongsupark:
So far ``ReplaceMember`` has been checking for error, by looking up an explicit string ``"Success"`` in stderr from ``"systemctl -M <ID> poweroff"``. However, with systemd v229 or newer, systemctl poweroff does not print out ``"Success"`` even if poweroff succeeded. This causes ``TestKnownHostsVerification`` to always return error like:

```
  --- FAIL: TestKnownHostsVerification (2.66s)
      client_test.go:60: Failed replacing machine: poweroff failed:
```

Fix it by changing the systemctl call to ``"machinectl poweroff <ID>"``, which should return error correctly. This would be better in the long run, than checking for messages in stderr.

Fixes: https://github.com/coreos/fleet/issues/1586

### DIFF
--- a/functional/platform/nspawn.go
+++ b/functional/platform/nspawn.go
@@ -591,15 +591,9 @@ func (nc *nspawnCluster) ReplaceMember(m Member) (Member, error) {
 	count := len(nc.members)
 	label := fmt.Sprintf("%s%s", nc.name, m.ID())
 
-	// The `machinectl poweroff` command does not cleanly shut down
-	// the nspawn container, so we must use systemctl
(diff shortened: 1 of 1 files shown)


jonboulle:
Can't remember why we couldn't use this in the first place - but this LGTM if the tests pass.

dongsupark:
Yesterday I did a minor update by changing error message. Functional tests also passed. I'll merge it soon.

dongsupark:
Merged https://github.com/coreos/fleet/pull/1587.

Issue #1588, opened 2016-05-24
functional: TestSingleNodeConnectivityLoss fails with unexpected systemdFileState


dongsupark:
``TestSingleNodeConnectivityLoss`` fails on CoreOS alpha (1032) with the following logs. This time I'm not sure what's going on. On CoreOS beta, the test is working well. Will investigate.

```
--- FAIL: TestSingleNodeConnectivityLoss (15.69s)
    connectivity-loss_test.go:174: Units files not in expected state after losing connectivity.
        Expected: map[single@launched.service:enabled single@loaded.service:enabled global@loaded.service:enabled global@launched.service:enabled]
        Actual: map[global@launched.service:linked-runtime global@loaded.service:linked-runtime single@launched.service:linked-runtime single@loaded.service:linked-runtime]
```

dongsupark:
Pushed a PR https://github.com/coreos/fleet/pull/1589

PR #1589, opened 2016-05-24
functional: call systemctl is-enabled instead of list-unit-files

dongsupark:
``TestSingleNodeConnectivityLoss`` has been so far retrieving is-enabled status by calling ``"systemctl list-unit-files UNITS"``. It was expected to get ``"enabled"`` for each unit. This has worked until systemd v227. However, that's not the case any more since systemd v228, precisely since https://github.com/systemd/systemd/commit/0ec0deaa30d0e68430f03fa6f32affa576481d18 ("install: follow unit file symlinks in /usr, but not /etc when looking for [Install] data"). Since then, systemctl list-unit-files returns ``"linked-runtime"`` status for each fleet unit, which is actually reasonable behavior for units that are not installed persistently under ``/etc/systemd``.

With systemd v228 or newer, however, ``TestSingleNodeConnectivityLoss`` in fleet functional tests cannot make use of ``"systemctl list-unit-files"`` any more. And it would not be optimal to simply change ``"enabled"`` to ``"linked-runtime"``, because then the test couldn't work with systemd < v228.

So let's work around it, by calling ``"systemctl is-enabled UNIT"`` instead. That command has returned correct status ``"linked-runtime"``, no matter whether systemd >= 228 or not.

Fixes: https://github.com/coreos/fleet/issues/1588

### DIFF
--- a/functional/connectivity-loss_test.go
+++ b/functional/connectivity-loss_test.go
@@ -54,8 +54,8 @@ func TestSingleNodeConnectivityLoss(t *testing.T) {
 		systemdState     []string
 	}{
 		"inactive": {[]string{"submit"}, "", "", nil},
-		"loaded":   {[]string{"load", "--no-block"}, "inactive", "enabled", nil},
-		"launched": {[]string{"start", "--no-block"}, "active", "enabled", []string{"loaded", "active", "running"}},
+		"loaded":   {[]string{"load", "--no-block"}, "inactive", "linked-runtime", nil},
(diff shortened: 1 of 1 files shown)


dongsupark:
I just updated minor things: ``s/break/continue``, and fixing a comment.

dongsupark:
If no one objects, I'd like to merge it tomorrow. Actually this is a critical one, as systemd v228 has been already around since many months.

dongsupark:
Merged. Thanks.

Issue #1591, opened 2016-05-25
functional: sudo command returns error on CoreOS 1053.2.0


dongsupark:
``"sudo systemctl stop fleet"`` seems to fail on CoreOS 1053.2.0, the most recent developer version. Multiple tests like ``TestDynamicClusterNewMemberUnitMigration`` or ``TestShutdown`` fail like below:

```
--- FAIL: TestShutdown (1.99s)
    shutdown_test.go:43: exit status 1
```

But those tests never failed on CoreOS 1032.1.0. It's also interesting, that both CoreOS versions, 1032 and 1053, have systemd v229.
Will investigate.

dongsupark:
It turns out, every sudo command fails like above. Hence this has nothing to do with systemd, but it rather looks like a PAM issue.
Digging into this a little more, I got such a message from stderr:
```
--- FAIL: TestShutdown (1.75s)
    shutdown_test.go:49: 
        stderr: Warning: Permanently added '172.18.1.1' (RSA) to the list of known hosts.^M
        sudo: unable to initialize PAM: No such file or directory
```

PR #1592, opened 2016-05-25
functional: copy PAM configurations into nspawn containers

dongsupark:
As CoreOS requires PAM configurations starting from v104x, PAM configurations on the host need to copied into the nspawn container. Otherwise, every sudo command fails with messages like:

```
  --- FAIL: TestShutdown (1.99s)
      shutdown_test.go:43: exit status 1
```

Fixes: https://github.com/coreos/fleet/issues/1591
See also https://coreos.com/blog/security-brief-coreos-linux-alpha-remote-ssh-issue.html

### DIFF
--- a/functional/platform/nspawn.go
+++ b/functional/platform/nspawn.go
@@ -400,9 +400,15 @@ func (nc *nspawnCluster) createMember(id string) (m Member, err error) {
 		// set up directory for fleet service
 		fmt.Sprintf("mkdir -p %s/etc/systemd/system", fsdir),
 
-		// minimum requirements for running systemd/coreos in a container
+		// minimum requirements for running systemd/coreos in a container.
+		// NOTE: copying /etc/pam.d is necessary only for such setups with
+		// sys-auth/pambase installed, for example developer image of CoreOS
+		// 1053.2.0. It should be fine also for systems where /etc/pam.d is
+		// empty, because then it should automatically fall back to
+		// /usr/lib64/pam.d, which belongs to sys-libs/pam.
 		fmt.Sprintf("mkdir -p %s/usr", fsdir),
 		fmt.Sprintf("cp /etc/os-release %s/etc", fsdir),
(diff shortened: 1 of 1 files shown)


kayrus:
@dongsupark Does it sufficient just to create a symlink?

jonboulle:
Better to copy as it's not read-only

kayrus:
Ok. @dongsupark `-a` already includes `-r`. http://linux.die.net/man/1/cp

dongsupark:
@kayrus Done. I removed ``"-r"``.

dongsupark:
If no one objects, I'd like to merge it tomorrow.

jonboulle:
I'm a bit confused about this actually - on my CoreOS 1053 machine, /etc/pam.d is empty just fine:
```
core@coreos ~ $ find /etc/pam.d/
/etc/pam.d/

```
On the other hand `/usr/lib64/pam.d/` contains a bunch of files. Is that somehow not being propagated correctly into the container?

dongsupark:
@jonboulle Hmm, I had a short look. On CoreOS 1053.2.0, /etc/pam.d is not empty, while on CoreOS 1032.1.0, /etc/pam.d is empty. Is there any difference between the 2 versions regarding directory structure? I'm not sure. I'll have a deeper look tomorrow.

dongsupark:
According to OOB discussion: It turns out, a portage ``sys-auth/pambase`` that includes ``/etc/pam.d``, is not included in a production image for CoreOS 1053.2.0, but only in a developer image built by CoreOS SDK. That made such a difference between our setups. 
As far as I just tested with both a production image and a developer image, it works already well. I suppose that's because when ``/etc/pam.d`` is empty, it falls back automatically to ``/usr/lib64/pam.d``. To reflect those behind stories, I just added some comments in the source code.

dongsupark:
Merged https://github.com/coreos/fleet/pull/1592. Thanks.

Issue #1594, opened 2016-05-25
functional: nspawn container doesn't start with systemd v230


dongsupark:
Functional tests don't work at all with systemd v230 (CoreOS 1053.2.0), because systemd-nspawn container fails to start with messages like:

```
--- FAIL: TestKnownHostsVerification (10.80s)
	client_test.go:36: unable to detect machine PID
```

On the other hand, the tests with systemd v229 + CoreOS 1053.2.0 are working well. From a quick investigation, this issue relates to https://github.com/systemd/systemd/pull/3014, commit https://github.com/systemd/systemd/commit/e01ff70a77e781734e1e73a2238af2e9bf7967a8.
See also https://github.com/coreos/rkt/pull/2440.

Full log:

```
2016/05/25 14:12:30 Creating nspawn container: /usr/bin/systemd-nspawn --bind-ro=/usr -b --uuid=79cb2498fef74c8b9b5aed641a957aef -M smoke79cb2498fef74c8b9b5aed641a957aef --capability=CAP_NET_BIND_SERVICE,CAP_SYS_TIME --network-bridge fleet0 -D /tmp/smoke/79cb2498fef74c8b9b5aed641a957aef/fs
2016/05/25 14:12:30 Creating transient systemd unit smoke79cb2498fef74c8b9b5aed641a957aef.service
2016/05/25 14:12:30 machinectl show -p Leader smoke79cb2498fef74c8b9b5aed641a957aef
...
2016/05/25 14:12:41 machinectl show -p Leader smoke79cb2498fef74c8b9b5aed641a957aef
2016/05/25 14:12:41 Failed detecting machine smoke79cb2498fef74c8b9b5aed641a957aef PID: unable to detect machine PID
2016/05/25 14:12:41 Destroying nspawn machine 79cb2498fef74c8b9b5aed641a957aef
2016/05/25 14:12:41 Failed tests, fetching logs from 79cb2498fef74c8b9b5aed641a957aef machine
2016/05/25 14:12:41 journalctl --directory=/tmp/smoke/79cb2498fef74c8b9b5aed641a957aef/fs/var/log/journal --root=/tmp/smoke/79cb2498fef74c8b9b5aed641a957aef/fs --no-pager
[...]

PR #1596, opened 2016-05-26
functional: write machine ID into nspawn container

dongsupark:
As systemd v230 or newer requires ``/etc/machine-id`` on a nspawn container, we need to write the container's own machine ID into ``/etc/machine-id``. Otherwise, functional tests don't start at all like:

```
  --- FAIL: TestKnownHostsVerification (10.80s)
    client_test.go:36: unable to detect machine PID
```

Fixes: https://github.com/coreos/fleet/issues/1594
See also https://github.com/systemd/systemd/pull/3014
See also https://github.com/coreos/rkt/pull/2440

### DIFF
--- a/functional/platform/nspawn.go
+++ b/functional/platform/nspawn.go
@@ -466,6 +466,11 @@ func (nc *nspawnCluster) createMember(id string) (m Member, err error) {
 			0644,
 		},
 		{
+			"/etc/machine-id",
+			nm.ID(),
+			0644,
+		},
+		{
 			"/home/core/.bash_profile",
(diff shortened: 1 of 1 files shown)


jonboulle:
this is backwards compatible right?

dongsupark:
@jonboulle Yes I think so.

jonboulle:
SGTM

kayrus:
@dongsupark LGTM

dongsupark:
Merged https://github.com/coreos/fleet/pull/1596. Thanks.

PR #1600, opened 2016-06-03
build: move definition of $VERSION to build-env

dongsupark:
VERSION should be defined in ``build-env``, not in individual scripts like ``build`` or ``functional/test``. Otherwise, a simple change in one of the scripts could cause unexpected version mismatch failures during functional tests.

I stumbled across this issue while rebasing giantswarm's gRPC branch on top of master.

/cc @hectorj2f


hectorj2f:
@dongsupark All good! 👍

PR #1601, opened 2016-06-10
scripts: new script genproto.sh for generating protobuf

dongsupark:
A new script ``genproto.sh`` generates all fleet protobuf bindings under ``./protobuf/``. This requires a native protoc v3.0 or higher, gogo/protobuf with commit c3995ae4. Protoc also needs to run with explicit options, such as --proto_path and --gogofast_out.

This script is originally taken from etcd's scripts/genproto.sh.

/cc @hectorj2f

### DIFF
--- a/scripts/genproto.sh
+++ b/scripts/genproto.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+#
+# Generate all fleet protobuf bindings.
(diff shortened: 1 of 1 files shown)


hectorj2f:
Thanks @dongsupark, lgtm

PR #1605, opened 2016-06-13
CONTRIBUTING: add description about adding tests

dongsupark:
Add description about adding tests to CONTRIBUTING.md.

> Ideally most functionalities should be covered by both unit tests and functional tests. This means, every time when you add a new functionality, you should also add unit tests, which is stored in *SRC_test.go*, as well as a [functional test](./functional/README.md), located under directory *./functional/*.
> Note that when adding a unit test under a new directory, you need to make sure that the new directory must be added to *TESTABLE* in [./test](./test). Without doing that, CI cannot run the new unit tests at all. Even build errors cannot be automatically detected.


jonboulle:
LGTM

dongsupark:
Merged https://github.com/coreos/fleet/pull/1605. Thanks.

Issue #1607, opened 2016-06-16
Documentation shows AWS_ACCESS_KEY & AWS_SECRET_KEY


ewoutp:
See github.com/coreos/fleet/blob/master/Documentation/examples/example-deployment.md

dongsupark:
@ewoutp ok, that's definitely not good. I wasn't aware of that, as it was written a long time ago. Can you probably create a PR to fix it?

ewoutp:
@dongsupark See #1608

PR #1608, opened 2016-06-16
Removed visible credentials

ewoutp:
Fixes https://github.com/coreos/fleet/issues/1607


dongsupark:
In general it looks good.
Ideally the subject line should be prefixed by a subsystem name, e.g. "docs: ", but in the case of documents, I'd suppose, that's not a strict rule.
BTW Reading this document, I start to think that this is not the only place where AWS keys are exposed. But sometimes I have to admit, I can't fix everything. ;-)
So if there's no objection, I'll merge it tomorrow.

dongsupark:
Merged https://github.com/coreos/fleet/pull/1608. Thanks.

PR #1609, opened 2016-06-16
fleetctl: fix races in unit tests

dongsupark:
Introduce 2 mutexes, ``cmu`` for accessing cAPI and ``smu`` for accessing sharedFlags, to avoid races during unit tests.

For example, output from ``"go test --race"``:

```
    WARNING: DATA RACE
    Write by goroutine 26:
      github.com/coreos/fleet/fleetctl.doUnloadUnits()
          /home/dpark/go/src/github.com/coreos/fleet/fleetctl/unload_test.go:26 +0x60
      github.com/coreos/fleet/fleetctl.TestRunUnloadUnits.func3()
          /home/dpark/go/src/github.com/coreos/fleet/fleetctl/unload_test.go:92 +0x89
    
    Previous write by goroutine 25:
      github.com/coreos/fleet/fleetctl.doUnloadUnits()
          /home/dpark/go/src/github.com/coreos/fleet/fleetctl/unload_test.go:26 +0x60
      github.com/coreos/fleet/fleetctl.TestRunUnloadUnits.func2()
          /home/dpark/go/src/github.com/coreos/fleet/fleetctl/unload_test.go:88 +0x89
```

### DIFF
--- a/fleetctl/destroy_test.go
+++ b/fleetctl/destroy_test.go
@@ -68,7 +68,9 @@ func TestRunDestroyUnits(t *testing.T) {
 		var wg sync.WaitGroup
 		errchan := make(chan error)
 
+		cmu.Lock()
 		cAPI = newFakeRegistryForCommands(unitPrefix, len(r.units), false)
+		cmu.Unlock()
 
 		wg.Add(2)
 		go func() {
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -112,6 +112,8 @@ var (
 		SSHPort       int
 	}{}
 
+	smu sync.Mutex
+
 	// current command being executed
 	currentCommand string
 
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -74,6 +74,8 @@ var (
 	// global API client used by commands
 	cAPI client.API
 
+	cmu sync.Mutex
+
 	// flags used by all commands
 	globalFlags = struct {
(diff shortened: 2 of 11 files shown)


jonboulle:
I'm not really thrilled about this - it's adding a bunch of locking just for the purposes of fixing the tests right?

dongsupark:
@jonboulle yeah I agree. I'm thinking myself too, these locks are pretty ugly. Today I have been trying to make the locking mechanism as generic as possible, though I don't like it either.
I'll try to rewrite the PR based on test-local locks. Thanks.

dongsupark:
Hmm, I just tried to add only test-local locks. But it looks like, test-local locks are not sufficient for race-free unit tests, because the shared states like cAPI or sharedFlags are spread both in core parts (like fleetctl.go) and in unit tests (like fleetctl_test.go). If I would really want to prevent races, I would have to create mutex that is available in both parts. As for cAPI for example, hold a mutex when assigning fakeAPI in unit tests, and also hold a mutex when reading from cAPI in core parts.

I think the underlying issue is that those shared states in fleet have never been protected from races, starting from the beginning. Of course it might not be a critical issue. Anyway it seems to need a major surgery.

dongsupark:
Recently I have been trying again to reimplement this PR, to achieve both race-free unit tests and clean API. To do that, however, I need to touch too many parts in fleetctl.
So for now I'm closing it, until I could have a better idea. Thanks.

PR #1612, opened 2016-06-20
functional: split part of TestScheduleReplace into ScheduleCircularReplace

dongsupark:
The last part of ``TestScheduleReplace`` has been occasionally failing with 2 units being active, instead of 1 active unit. Its reason is probably because the circular replace test has been running immediately after a simple replace test. Even if one of the units was destroyed before testing circular replace, the other unit remained scheduled on the other
machine, which prevented a new unit from detecting the existing unit as circular replaces. That resulted in a pretty subtle undefined state.

To fix this issue, split the last part of ``TestScheduleReplace`` into a new test ``TestScheduleCircularReplace``, so that the circular replace part could start in a fresh status. This will improve code readability too.

### DIFF
--- a/functional/scheduling_test.go
+++ b/functional/scheduling_test.go
@@ -341,7 +341,6 @@ func TestScheduleReplace(t *testing.T) {
 	}
 	defer cluster.Destroy(t)
 
-	// Start with a simple three-node cluster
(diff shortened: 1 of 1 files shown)


dongsupark:
Merged it to get rid of this annoying problem.

PR #1613, opened 2016-06-21
build-env: determine GO15VENDOREXPERIMENT correctly

dongsupark:
Build environment has been buggy as it always set ``GO15VENDOREXPERIMENT`` to 1, even when a local go 1.6 compiler is used on a host where the default go compiler is 1.5. A typical example is functional test, which automatically installs go 1.6 runtime under ``/home/core/go`` for every test, while ``/usr/bin/go`` still remains as 1.5. Then ``GO15VENDOREXPERIMENT`` is set to 1 before running the functional test, but it's not set back to 0, even when a 1.6 compiler is supposed to be used. For instance, this bug breaks a build of ``github.com/ugorji/go``, which is going to be necessary for building dependencies in the future.

So the fix is to set ``GO15VENDOREXPERIMENT`` to 0, when go compiler is not 1.5. Also such action must run after ``"eval $(go env)"``, otherwise that value could become back to 1.

### DIFF
--- a/build-env
+++ b/build-env
@@ -1,19 +1,26 @@
-version=$(go version)
-regex="go([0-9]+).([0-9]+)."
-if [[ $version =~ $regex ]]; then
-  if [ ${BASH_REMATCH[1]} -eq "1" -a ${BASH_REMATCH[2]} -eq "5" ]; then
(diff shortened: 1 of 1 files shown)


dongsupark:
Please ignore this PR, as it's not a correct fix. I'll create another one.

PR #1614, opened 2016-06-21
functional: increase RAM to 1024MB to fix build errors

dongsupark:
Increase RAM of test VM from 512MB to 1024MB, to fix build errors when compiling packages like ``github.com/ugorji/go``, which will be needed in the future for dependencies. Without this change, build with go 1.6 or higher fails like that:

```
  github.com/ugorji/go/codec: /home/core/go/pkg/tool/linux_amd64/compile: signal: killed
```

### DIFF
--- a/functional/config.rb
+++ b/functional/config.rb
@@ -42,7 +42,7 @@ $update_channel='stable'
 
 # Customize VMs
 #$vm_gui = false
-$vm_memory = 512
+$vm_memory = 1024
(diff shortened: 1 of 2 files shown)


dongsupark:
I merged it quickly, as it's a blocker for further testing of glide conversion. Nowadays it's almost always sane to have 1G RAM on any VM, I suppose.

jonboulle:
LGTM

PR #1617, opened 2016-06-22
travis: add tip to the compiler and update build-env.

dongsupark:
Allow travis to build fleet also with the most recent go compiler, "tip".
For example, as of 2016-06-22, go "tip" compiler's version is ``"devel +1f44643 Wed Jun 22 00:12:55 2016 +0000"``.
To make it work, build-env needs to be also updated, not to exit with error in case of the devel version.


dongsupark:
I just changed ``.travis.yml`` to allow ``"tip"`` compilation to fail, which is reasonable for experimental go versions.

PR #1618, opened 2016-06-23
scripts: use latest gogo/protobuf for proto files

dongsupark:
As ``gogo/protobuf`` was updated to generate grpc.SupportPackageIsVersion3 by default, ``genproto.sh`` needs to be also updated.

See also https://github.com/coreos/etcd/pull/5703


jonboulle:
lgtm

dongsupark:
Merged. Thanks.

PR #1620, opened 2016-06-24
[WIP] ssh: workaround to not block Stdin from returning corrrectly

dongsupark:
Assign ``session.stdin`` using ``StdinPipe()`` to avoid cases of ssh connetions being blocked on user input.

WIP, do not merge. Not tested.

Reported-by @kayrus 
maybe partly resolves https://github.com/coreos/fleet/issues/1499

### DIFF
--- a/ssh/ssh.go
+++ b/ssh/ssh.go
@@ -16,6 +16,7 @@ package ssh
 
 import (
 	"errors"
+	"io"
 	"net"
 	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


dongsupark:
After traversing a long rabbit hole, I realized that it would not be easy at all to fix this issue. Even with goroutines & wrappers around stdin, io.Copy still remained being blocked.
I'll reopen it when I could have a better idea.

Issue #1622, opened 2016-06-30
ETCD key already exists


ewoutp:
When testing the current master version on a cluster where all fleet instances are v0.11.7, I get these errors when starting a (master) version of fleet on one of the machines.

```
fleetd[4900]: INFO fleetd.go:75: Starting fleetd version v0.13.0-31-geb95cd9
fleetd[4900]: INFO fleetd.go:203: No provided or default config file found - proceeding without
fleetd[4900]: INFO server.go:175: Establishing etcd connectivity
fleetd[4900]: ERROR server.go:192: Server register machine failed: 105: Key already exists (/_coreos.com/fleet/machines/d5ca2e9d4039e480807464280c9ab4a2/object) [203152748]
```

A little while later I get this panic.

```
fleetd[4900]: ERROR server.go:192: Server register machine failed: 105: Key already exists (/_coreos.com/fleet/machines/d5ca2e9d4039e480807464280c9ab4a2/object) [203153266]
fleetd[4900]: INFO server.go:188: hrt.Register() success
fleetd[4900]: INFO server.go:198: Starting server components
fleetd[4900]: panic: runtime error: invalid memory address or nil pointer dereference
fleetd[4900]: [signal 0xb code=0x1 addr=0x20 pc=0x4bc024]
fleetd[4900]: goroutine 49 [running]:
fleetd[4900]: panic(0xc06080, 0xc82000a0a0)
fleetd[4900]:         /usr/local/go/src/runtime/panic.go:481 +0x3e6
fleetd[4900]: github.com/coreos/fleet/pkg.(*reconciler).Run.func1(0xc8201cef00, 0xc8201b9560, 0xc8201cf2c0)
fleetd[4900]:         /opt/fleet/gopath/src/github.com/coreos/fleet/pkg/reconcile.go:65 +0x74
fleetd[4900]: created by github.com/coreos/fleet/pkg.(*reconciler).Run
fleetd[4900]:         /opt/fleet/gopath/src/github.com/coreos/fleet/pkg/reconcile.go:69 +0x89
systemd[1]: fleet.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
systemd[1]: fleet.service: Unit entered failed state.
systemd[1]: fleet.service: Failed with result 'exit-code'.
```

dongsupark:
Thanks for the bug report.

[...]

PR #1623, opened 2016-07-01
fleetd: check that eStream is available before accessing to that

dongsupark:
To avoid potential nil-dereferences in agent reconciliation, check that eStream is available before accessing to ``eStream``. Otherwise, just return.
    
Example error logs:
```
ERROR server.go:192: Server register machine failed: 105: Key already exists
(/_coreos.com/fleet/machines/d5ca2e9d4039e480807464280c9ab4a2/object) [203153266]
INFO server.go:188: hrt.Register() success
INFO server.go:198: Starting server components
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x20 pc=0x4bc024]
goroutine 49 [running]:
panic(0xc06080, 0xc82000a0a0)
        /usr/local/go/src/runtime/panic.go:481 +0x3e6
github.com/coreos/fleet/pkg.(*reconciler).Run.func1(0xc8201cef00, 0xc8201b9560, 0xc8201cf2c0)
/opt/fleet/gopath/src/github.com/coreos/fleet/pkg/reconcile.go:65 +0x74
created by github.com/coreos/fleet/pkg.(*reconciler).Run
/opt/fleet/gopath/src/github.com/coreos/fleet/pkg/reconcile.go:69 +0x89
systemd[1]: fleet.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
systemd[1]: fleet.service: Unit entered failed state.
systemd[1]: fleet.service: Failed with result 'exit-code'.
```
    
Fixes: https://github.com/coreos/fleet/issues/1622

### DIFF
--- a/pkg/reconcile.go
+++ b/pkg/reconcile.go
@@ -58,6 +58,9 @@ func (r *reconciler) Run(stop <-chan struct{}) {
 	go func() {
 		abort := make(chan struct{})
 		for {
+			if r.eStream == nil {
+				return
+			}
 			select {
 			case <-stop:
 				close(abort)
--- a/server/server.go
+++ b/server/server.go
@@ -189,7 +189,7 @@ func (s *Server) Run() {
 				break
 			}
 		}
-		log.Errorf("Server register machine failed: %v", err)
+		log.Warningf("Server register machine failed: %v, retrying in %d sec.", err, sleep)
 		time.Sleep(sleep)
 	}
 


jonboulle:
when would this ever be nil? Isn't that a bug in other code?

dongsupark:
@jonboulle 
To be honest, I don't know.
AFAICT the only case of eStream being nil is when ``disable_watches`` is turned on. (See https://github.com/coreos/fleet/blob/master/server/server.go#L112) ``disable_watches`` defaults to false. If disable_watches is set to true by user, rStream (eStream) could be nil.
The thing is, I cannot reproduce the crash. Even if the commit could fix the case, I'm not sure if that's the same one as what @ewoutp has seen. So if anyone has a better idea, please let me know.

ewoutp:
I do have `disabled_watches` turned on.

ewoutp:
This is the config of one of my machines (only METADATA differs per machines)
```
[Service]
Environment=FLEET_METADATA=region=fr-1,odd=true,core=true,lb=true,gogs-sync=true,registry-sync=true
Environment=FLEET_PUBLIC_IP=192.168.35.3
Environment=FLEET_AGENT_TTL=30s
Environment=FLEET_DISABLE_ENGINE=false
Environment=FLEET_DISABLE_WATCHES=true
Environment=FLEET_ENGINE_RECONCILE_INTERVAL=10
Environment=FLEET_TOKEN_LIMIT=50
```

dongsupark:
> I do have disabled_watches turned on.

@ewoutp Aha. Then it makes sense. Hopefully this PR could fix the bug in your case. Thanks!

dongsupark:
Merged. Thanks.

Issue #1559, opened 2016-04-19
fleetctl: get statuses (and restart) of the global unit


kayrus:
At the moment it is not possible to get status of the global unit. fleetctl returns `Unable to determine status of global unit global.service.` error message. The workaround it to use `fleetctl ssh 6abfab3268cf6bb745b0aaf5da9b0cad systemctl status global.service`, which is too tricky.

I suggest to get all machines where global unit is executed and iterate through them. This will potentially create a long running process when you have thousands of machines in the cluster, and that is why we have to implement `machineid/hostname/machineIP` filter.

This solution will allow us to restart global units as well (we will use same code as in `fleetctl/status.go`). And we can resolve #975. `fleetctl ssh <whatever> sudo systemctl restart <whatever>` looks too tricky. When you need to restart several units, you'll have to write something like:

``` sh
for u in "unit1 unit2 unit3"; do fleetctl ssh $u sudo systemctl restart $u; done
```

I don't like the solution provided in #961, because it can create several simultaneous ssh connections when you will trigger batch of units and you can not limit amount of connections. The most bug-free and simple solution is to use `status.go` code.

/cc @antrik @jonboulle @tixxdz does it sound reasonable for you?

dongsupark:
+1 for the thought that ``fleetctl restart`` would be a nice feature. There must be a better way than ``"fleetctl ssh <whatever> sudo systemctl restart"``, which is indeed tricky. Maybe after the milestone v0.13, we could work on it.

antrik:
@kayrus I still think the approach I proposed for `restart` is more appropriate. (And I'd love to implement it if I'm given permission.)

I don't have much of an opinion on using `systemctl status`... In what situations would you use it?

antrik:
Eh, the last bit was bogus. I don't know though how `fleetctl status` works right now, and what the problem is with showing the status for global units... Would have to investigate first before I can give useful answer.

tixxdz:
> @kayrus I still think the approach I proposed for restart is more appropriate. (And I'd love to implement it if I'm given permission.)

Please go ahead there are no such permissions, just code the feature if you think that your solution is the best, or show something.

dongsupark:
I just added this PR to milestone v1.0, as it has been requested since quite a long time. We need to somehow add this stuff.

dongsupark:
Closed via https://github.com/coreos/fleet/pull/1624

PR #1624, opened 2016-07-06
fleetctl: status and restart commands with global units

dongsupark:
This PR is _yet another_ revision of a long-running issue, `fleetctl restart` and `status` with global units.
- First allow `"fleetctl status"` to run with global units. For the purpose, implement a helper `cmdGlobalMachineState()` that filters a list of machines only with unique entries.
- Introduce a new command `"fleetctl restart"`. This part is originally taken from https://github.com/coreos/fleet/pull/1238 written by "Kyle Boorky kboorky@turbine.com", committed by @hhoover . 
- Introduce unit test and functional test for `fleetctl restart`.
- Add new tests for global tests in `TestScheduleGlobalUnits()`.

This PR is not about changing API or metadata format in etcd registry. It's not introducing any new mechanism, because I don't think I can afford changing such design at the moment. It's rather a compromise between practical requirements and the current design of fleet. Though we still need to think about how to decrease number of ssh connections when restarting global units, as suggested by @kayrus. See also https://github.com/coreos/fleet/issues/1559.

Fixes https://github.com/coreos/fleet/issues/975

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -1246,3 +1246,35 @@ func waitForState(stateCheckFunc func() error) (time.Duration, error) {
 		}
 	}
 }
+
+// cmdGlobalMachineState runs a specific fleetctl command on each target machine
+// where global units are started. To avoid unnecessary ssh connections being
+// alive, it filters out the list of machines as much as possible.
+func cmdGlobalMachineState(cCmd *cobra.Command, globalUnits []schema.Unit) (err error) {
+	cmd := cCmd.Name()
+	mapUNs := map[string]string{}
+	for _, unit := range globalUnits {
+		m := cachedMachineState(unit.MachineID)
+		if m == nil || m.ID == "" || m.PublicIP == "" {
+			continue
+		}
+		mapUNs[m.ID] = unit.Name
+	}
+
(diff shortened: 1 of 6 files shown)


simonvanderveldt:
@dongsupark have you been using this since you created this PR? How's it holding up?
This will only allow for restarting existing units and doesn't enable updating an existing global unit one by one, right?

We've been doing something similar outside of fleet using SSH connections and calling `systemctl`, but had some issues with fleet getting in a broken state when updating global units, which seemed to be caused by the fact that we weren't updating the desired state of the units in fleet (ie we submitted a new one using fleet and then started it using `systemctl` without doing a `fleetctl start`).

Is anyone planning to review and merge this in the near future?

dongsupark:
@simonvanderveldt 
I'm just waiting for anyone to review this PR. And I thought it wouldn't be that urgent issue.
I'm also aware that in the past not everyone was happy with supporting the restart command. So I wouldn't just merge this PR, unless anyone could review it.

dongsupark:
Fixed minor things, and force-pushed.
If no one objects, I'd like to merge it before the release v1.0, to close https://github.com/coreos/fleet/issues/1559.

PR #1625, opened 2016-07-07
test: Updated GPG pub key ID

kayrus:
CoreOS has changed GPG keys for new releases. If we don't use `48F9B96A2E16137F` here and there is already imported old GPG key, it will be not updated and GPG check will fail.


dongsupark:
LGTM.

Per OOB discussion with @kayrus , I verified like that:

```bash
$ curl -s https://coreos.com/security/image-signing-key/CoreOS_Image_Signing_Key.asc \
  | gpg --import --keyid-format LONG
$ gpg --list-keys --keyid-format LONG "CoreOS Buildbot (Offical Builds)"
pub   4096R/50E0885593D2DCB4 2013-09-06
uid                          CoreOS Buildbot (Offical Builds) <buildbot@coreos.com>
sub   4096R/07FA9ED31CB5FA26 2015-08-31 [expires: 2017-08-30]
sub   4096R/48F9B96A2E16137F 2016-05-16 [expires: 2017-05-16]
```

So here the new GPG key is the last subkey, ``48F9B96A2E16137F``. Checking only the primary key ``50E0885593D2DCB4`` is not sufficient. That's why we need to change key in the script to the new subkey.

dongsupark:
Merged. Thanks.

PR #1626, opened 2016-07-07
functional: retrieve stdout/stderr after running command

dongsupark:
Make every call ``cluster.Fleetctl()`` (or ``run()``) return both ``stdout`` and ``stderr``, to get them included in the result error buffer. We need to do this to make sure that every potential error message is printed out. This change will help us investigate occasional failures in functional tests.


dongsupark:
Expanded a little more to several call sites. It's still trivial bulk changes.
I'll merge it soon, to get more hints when functional tests would fail again on semaphoreci.

Issue #1631, opened 2016-07-13
Ignored environment variables.


odolbeau:
Hi.

In v0.13.0 it seems environment variables are not taken into account.
I tried with `FLEETCTL_ENDPOINT` for example.

dongsupark:
Yeah, true.
From a quick research, I think the env variable ``FLEETCTL_ENDPOINT`` is never taken into account, neither in v0.13 nor in v0.11.7. I'm not sure since when it has been broken, but that's clearly a bug.
As a workaround, please use the cmdline option ``--endpoint``.

odolbeau:
It works correctly in v0.11.7. :)

odolbeau:
It works correctly in v0.11.7. :)

dongsupark:
@odolbeau Fixed a bug. See https://github.com/coreos/fleet/pull/1633. Thanks for the bug report.

berglh:
Confirmed latest master ```fleetctl version v0.12.0-281-ge98df4c``` resolved this issue - thanks.

dongsupark:
@berglh Glad to hear that. Thanks!

PR #1633, opened 2016-07-13
fleetctl: reenable support for environment variables

dongsupark:
``Fleetctl`` has not been able to make use of environment variables such as ``FLEETCTL_ENDPOINT``, because of a missing call to ``getFlagsFromEnv()``. That was a regression from commit 848d3561 ``("fleetctl: convert cli to cobra")``.

Fix it by calling ``getFlagsFromEnv()`` again based on ``spf13/pflag`` instead of the native flag.

Reported by @odolbeau.
Fixes https://github.com/coreos/fleet/issues/1631

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -31,6 +31,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


jonboulle:
awkward - nice catch!

PR #1635, opened 2016-07-14
systemd: get dbus connection directly in NewSystemdUnitManager

dongsupark:
Let's get dbus connection directly inside ``NewSystemdUnitManager``, instead of calling a separate function ``createDbusConnection()``.
    
Suggested by @jonboulle.
See https://github.com/coreos/fleet/pull/1393#discussion_r70782867
/cc @hmalphettes


jonboulle:
lgtm

PR #1637, opened 2016-07-15
*: update license header

dongsupark:
In each *.go file, Replace ``"Copyright 201? CoreOS, Inc"`` with ``"Copyright 201? The fleet Authors"``, to be correspondent with other CoreOS projects.

Also update ``./test`` to check if each go file contains license headers.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,5 +1,5 @@
 CoreOS Project
-Copyright 2014 CoreOS, Inc
+Copyright 2014 The fleet Authors
 
 This product includes software developed at CoreOS, Inc.
 (http://www.coreos.com/).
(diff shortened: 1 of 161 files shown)


jonboulle:
Seems good at a  high level, although I don't think you need to change the year, and travis is failing

dongsupark:
@jonboulle As for travis failures, I think something between etcd and grpc-gateway has changed. It always fails with the message below:
 
```
package github.com/coreos/fleet/vendor/github.com/coreos/etcd
	imports github.com/coreos/etcd/etcdmain
	imports github.com/coreos/etcd/clientv3
	imports github.com/coreos/etcd/etcdserver/etcdserverpb
	imports github.com/gengo/grpc-gateway/runtime
	imports github.com/grpc-ecosystem/grpc-gateway/runtime/internal: use of internal package not allowed
```

I have no idea. Let's see.

dongsupark:
I updated the tree to keep each year string in the license header just as before.
The travis error is magically gone, too. Now I suppose this PR can be merged.

PR #1638, opened 2016-07-15
functional: introduce a test TestFleetctlWithEnv

dongsupark:
``TestFleetctlWithEnv`` runs simply fleetctl list-machines, but by setting an environment variable ``FLEETCTL_ENDPOINT``, instead of the cmdline option ``'--endpoint'``. This way functional test is able to detect regressions like https://github.com/coreos/fleet/issues/1631.


jonboulle:
LGTM on green

Issue #555, opened 2014-06-20
Update machine metadata through etcd


boombatower:
Being able to update metadata would allow for much more powerful scheduling in lieu of resource based scheduling. Perhaps this is already possible and just needs proper documentation, but I was unable to find anything.

From what I understand etcd is used to store the fleet registry and presumably the metadata. So perhaps one can manually tweak the etcd entries? I was unable to find where they were stored as all I see are update related files in etcd. And the etcd reference in cloudinit only lists basic machine info and no fleet information.

I would imagine restarting the fleet service with different metadata environment variable would do the trick, but would be rather brute force and would interrupt things. I'll try that next and see if state of services and such remains. EDIT: Seems to do the trick. Various status checks and list-units/list-machines do not show it while restarting obviously, but running services are not interrupted and show up again once up. Given my workaround was creating fake services (which unfortunately have to be on fleet and mapped to specific machine) to make up for lack of updating metadata...this approach of restarting seems better.

My usecase:
I am attempting to simulate resource based scheduling in a manner specific to my services. I have a few services of varying resource requirements and I can calculate how many I can fit (leaving margin for error). My thinking then was to add metadata indicating the types of services that could still fit on the machine (via a script on each node). The services would be define to look for their related metadata tag. Once the machine was too full for certain services the metadata could be removed.

This allows for hosting multiple of the same service on one machine (perhaps of varying sizes) instead of primitive...conflicts checks and what not.

Looking forward to your thoughts.

chr0n1x:
+1

dongsupark:
As this issue belongs to milestone 1.0, I have been reading the discussions in the issue.
I would assume that the basic functionality of updating machine metadata through etcd was already done via https://github.com/coreos/fleet/issues/1642.

I can also see that there were discussions about adding schedulers like https://github.com/coreos/fleet/issues/922. Though I think that at the moment it's not a smart choice to add such schedulers, as it would require a lot of work while taking a risk of breaking changes.

So I'm closing it. If I'm missing anything, please let me know.

PR #1632, opened 2016-07-13
fleetd: support operators in metadata

dongsupark:
This PR allows users to define metadata in units, such as ``"ram=1024"``. Precisely we can define the operator in the ``[X-Fleet]`` section in a unit as below:

```ini
[X-Fleet]
MachineMetadata=ram>=1024
```

Supported operators are ``"<="``, ``">="``, ``"!="``, ``"=="``, ``"<"``, ``">"``.

This PR also contains more checks in fleetctl, more test cases in unit tests, and a new functional test.

Fixes https://github.com/coreos/fleet/issues/1143
Supersedes https://github.com/coreos/fleet/pull/1294

PR #1642, opened 2016-07-16
fleetd: support dynamic metadata

dalbani:
Source: https://github.com/coreos/fleet/pull/1077

### DIFF
--- a/api/machines.go
+++ b/api/machines.go
@@ -15,14 +15,20 @@
 package api
 
 import (
-	"fmt"
+	"encoding/json"
+	"errors"
 	"net/http"
 	"path"
+	"regexp"
 
 	"github.com/coreos/fleet/client"
 	"github.com/coreos/fleet/log"
 	"github.com/coreos/fleet/machine"
 	"github.com/coreos/fleet/schema"
+)
+
+var (
+	metadataPathRegex = regexp.MustCompile("^/([^/]+)/metadata/([A-Za-z0-9_.-]+$)")
 )
 
 func wireUpMachinesResource(mux *http.ServeMux, prefix string, tokenLimit int, cAPI client.API) {
@@ -36,12 +42,24 @@ type machinesResource struct {
 	tokenLimit uint16
 }
 
-func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	if req.Method != "GET" {
-		sendError(rw, http.StatusBadRequest, fmt.Errorf("only HTTP GET supported against this resource"))
-		return
-	}
+type machineMetadataOp struct {
+	Operation string `json:"op"`
+	Path      string
+	Value     string
+}
 
+func (mr *machinesResource) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	switch req.Method {
+	case "GET":
+		mr.list(rw, req)
+	case "PATCH":
+		mr.patch(rw, req)
+	default:
+		sendError(rw, http.StatusMethodNotAllowed, errors.New("only GET and PATCH supported against this resource"))
+	}
+}
+
+func (mr *machinesResource) list(rw http.ResponseWriter, req *http.Request) {
 	token, err := findNextPageToken(req.URL, mr.tokenLimit)
 	if err != nil {
 		sendError(rw, http.StatusBadRequest, err)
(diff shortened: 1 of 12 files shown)


api/machines.go
```diff
@@ -36,12 +42,24 @@ type machinesResource struct {
 	tokenLimit uint16
 }
 
+type machineMetadataOp struct {
+	Operation string `json:"op"`
+	Path      string
+	Value     string
```
dalbani: In relation with the question I asked about a possible TTL feature, I think this structure should be made more future-proof.
In practice, `Value` should be an object instead of a `string`.
That gives the room to later implement such improvement for example: https://github.com/dalbani/fleet/commit/c0a305037a214351e62442f7d234865699e9520f#diff-82e778d9c582a0fad1db38076b55deaaR47

api/machines_test.go
```diff
@@ -136,3 +137,126 @@ func TestExtractMachinePage(t *testing.T) {
...
 	}
 }
+
+func TestMachinesPatchAddModify(t *testing.T) {
+	reqBody := `
+	[{"op": "add", "path": "/XXX/metadata/foo", "value": "bar"},
+	 {"op": "replace", "path": "/YYY/metadata/ping", "value": "splat"}]
+	`
+
+	resource, rw := fakeMachinesSetup()
```
dongsupark: Build of unit test fails at this line as well as other 4 other call sites, because ``fakeMachinesSetup()`` is undefined. In the previous PR #1077 , the function actually did exist.

dalbani: Oops, will fix that.

dongsupark:
First of all, thank a lot for the rebase. This helps us a lot go on with the feature.
Is there any chance to write functional tests to cover this feature?

dalbani:
I have very little experience with Go and fleet, so if you could give me
some directions, that would be very useful.

dongsupark:
@dalbani It's basically about trying to reproduce an exact test scenario inside the fleet's functional test framework, where systemd-nspawn instances run. Normally you could just make use of existing helpers that are already available under the source directory ``functional``. It might be also a good idea to create a sample unit file under ``functional/fixtures/units``.

One of the examples would be a new functional test ``TestMetadataOperator``, which I wrote last week: https://github.com/coreos/fleet/pull/1632/commits/68028c7e5dcae80013124ff34a3794452770ec58 in https://github.com/coreos/fleet/pull/1632. As the original PR didn't include such a test, I needed to write the test on my own. See also https://github.com/coreos/fleet/blob/master/functional/README.md.

api/machines_test.go
dalbani: I've modified `api/machines_test.go` according to your remark.
Running the test suite locally is now successful.
What is the reason why the CI test triggered by GitHub didn't show up any error by the way?

dongsupark: > I've modified api/machines_test.go according to your remark.
> Running the test suite locally is now successful.
> What is the reason why the CI test triggered by GitHub didn't show up any error by the way?

Great. Unit test builds now without error. Thanks!
I'm not sure why the travis job didn't get triggered in the first place. Usually it should run no matter whether there's a build error or not. It was maybe a temporary error from travisci.

dongsupark:
Gentle ping. Any updates on functional test?
BTW I added this PR to milestone v1.0.0, as it could resolve the issue https://github.com/coreos/fleet/issues/555.

dalbani:
Oh, dear, I had forgotten with the summer and everything :-)
In order to first make my PR in sync with the `master` branch, I've just committed some RPC-related code.
How important is it to implement the related methods in [registry/rpc/rpcregistry.go](https://github.com/dalbani/fleet/blob/21c84d6079a034deb94c6645405e724c4f1183aa/registry/rpc/rpcregistry.go#L247) by the way?
And I've just noticed that the SemaphoreCI build has failed. Do you know why?

dongsupark:
@dalbani Thanks for pushing the RPC related code. Good point. I think we should also add the MachineMetadata methods to rpc/registry.
As for the semaphoreci failures, every grpc-enabled test is failing. That's interesting. I'd need to have a closer look.

dongsupark:
@dalbani How about adding into RegistryMux like below:
```
func (r *RegistryMux) MachineState(machID string) (machine.MachineState, error) {
        return r.getRegistry().MachineState(machID)
}
```
Wouldn't that resolve the testing error of machine being unable to be found, when enable_grpc=on?

dongsupark:
I had another look. Unfortunately it turns out, there are still things to be done.
For the non-gRPC path, it looks fine. Functional tests run without errors.
For the gRPC path, necessary methods in RPCRegistry, such as SetMachineMetadata or MachineState, are not implemented. To support the methods, we would need to add also necessary methods into `fleet.proto`. Then we could support them also in RPCRegistry.
That sounds like a challenging weekend project. ;-)

dongsupark:
Ah, sorry for noise. On Tuesday I misunderstood something.
We don't have to do extra work in protobuf. Actually we just need to tweak registryMux to make it call methods from etcdRegistry.
So I think we can simply merge this PR, together with an additional functional test. Tomorrow I'll review this PR again, merge it, and I'll create another PR for a simple registryMux fix as well as a functional test for dynamic metadata.

dalbani:
I am hardly in position to complain, given that I haven't been to the task of writing a functional test :-(
(I haven't busy lately with another project even though that's not really an excuse.)
But I have a remark regarding the HTTP API -- well, in fact, I'd like to reiterate the remark that I made above: https://github.com/coreos/fleet/pull/1642#r71066179
In `type machineMetadataOp struct`, I would turn:
```go
	Value     string
```
into:
```go
	Value     struct {
		Value	string	`json:"value"`
	}
```
That would keep the ability to enhance the API *in the future*, for example for a TTL feature mentioned (and proof-of-concept implemented).
Otherwise, with simply `Value     string`, API-breaking changes would be required.

By the way, I have "allowed edits from maintainers" on this PR, should you want to edit it directly.

dongsupark:
@dalbani All right, I'll do it for avoiding API breakage. Thanks!

PR #1643, opened 2016-07-17
build: secured CDIR variable, now should work with spaces in path

kayrus:


dongsupark:
Generally I think it's a good idea.
But functional test's build is always failing with:
```
can't load package: package github.com/coreos/fleet/functional: cannot find package "github.com/coreos/fleet/functional" in any of:
	/home/core/go/src/github.com/coreos/fleet/functional (from $GOROOT)
	/home/core/fleet/functional/gopath/src/github.com/coreos/fleet/functional (from $GOPATH)
```

Interesting.

dongsupark:
Ok, the build error happens because both ``functional/test`` and ``build-env`` try to define a variable ``${CDIR}``, although ``functional/test`` sources ``build-env``. You need to rename ``${CDIR}`` into another name, either in ``functional/test`` or ``build-env``. I think a correct fix would be to update ``build-env``, as that's also imported by other files.

kayrus:
fixed in other way. just use `CDIR=$(cd `dirname "$0/.."` && pwd)`

dongsupark:
LGTM

PR #1644, opened 2016-07-19
vendor: update glide and vendors

dongsupark:
Update vendors to the most recent versions as of 2016-07-19.
Also fix a build error in fleetctl due to an API change in ``spf13/cobra``.


dongsupark:
Also updated ``scripts/glide-update.sh`` to ensure that it's running inside ``$GOPATH``.

PR #1645, opened 2016-07-19
travis,functional: bump golang to 1.6.3

dongsupark:
A new golang 1.6.3 is available especially for a security fix.


dongsupark:
As it's trivial, I'll merge it right away.

Issue #1216, opened 2015-05-07
Race condition in unloading units


patrickbcullen:
I have noticed a race condition in production where job is supposed to be unloaded, but never stops.

I think I found the bug here https://github.com/coreos/fleet/blob/7a648771ca405019486f52d83ec8ee3f6810a6ce/agent/agent.go#L98. 

I have an example where the systemd file is gone, but the unit is still running. This happened because the kill command in the systemd file did not succeed. Since this code does not check for error it happily removes all the systemd files. Now that the systemd files are gone I cannot stop it through fleet since fleet cannot call the systemd stop because it already deleted the systemd file.

bcwaldon:
@patrickbcullen Can you provide an explicit set of repro steps for this?

PR #1646, opened 2016-07-20
systemd,agent: unload unit only when TriggerStop() runs successfully

dongsupark:
In ``Agent.unloadUnit()``, if ``systemdUnitManager.TriggerStop()`` returns any error, do not unload systemd units. Otherwise the unit could get into a state where the unit cannot be stopped via fleet, because the unit file was already removed.

Fixes https://github.com/coreos/fleet/issues/1216
/cc @patrickbcullen

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -98,11 +98,22 @@ func (a *Agent) unloadUnit(unitName string) {
 	a.registry.ClearUnitHeartbeat(unitName)
 	a.cache.dropTargetState(unitName)
 
-	a.um.TriggerStop(unitName)
+	errStop := a.um.TriggerStop(unitName)
+	if errStop != nil {
+		log.Warningf("Failed stopping unit(%s): %v", unitName, errStop)
+	} else {
+		log.Infof("Stopped unit(%s)", unitName)
+	}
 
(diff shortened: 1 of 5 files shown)


dongsupark:
BTW the travis error is again caused by an API change in go-systemd needed by etcd. It's going to be fixed in days from the etcd side.

dongsupark:
I'll merge it soon, as it's clearly a potential bug.

agent/agent.go
```diff
@@ -98,11 +98,22 @@ func (a *Agent) unloadUnit(unitName string) {
 	a.registry.ClearUnitHeartbeat(unitName)
 	a.cache.dropTargetState(unitName)
 
-	a.um.TriggerStop(unitName)
+	errStop := a.um.TriggerStop(unitName)
+	if errStop != nil {
+		log.Warningf("Failed stopping unit(%s): %v", unitName, errStop)
```
jonboulle: nit - these messages are a bit misleading - because it's just the trigger, not the actual stopping, right?

dongsupark: @jonboulle Right. I'll change it tomorrow. Thanks.

Issue #998, opened 2014-10-22
fleet apparently ignores failed attempt to start unit


bcwaldon:
Discovered in https://github.com/coreos/fleet/issues/974:

```
Oct 17 04:12:35 infra01.c.secondbit-infra.internal fleetd[562]: ERROR manager.go:80: Failed to trigger systemd unit hgweb.1.service start: Unit code-ssh.1.service failed to load: No such file or directory.
Oct 17 04:12:35 infra01.c.secondbit-infra.internal fleetd[562]: INFO reconcile.go:274: AgentReconciler completed task: type=StartUnit job=hgweb.1.service reason="unit currently loaded but desired state is launched"
```

PR #1647, opened 2016-07-22
agent,unit: check errors from start/stop/unload-ing unit

dongsupark:
Now that UnitManager.{TriggerStart,TriggerStop,Unload}() could return an error, the error should be delivered up to ``mapTaskToFunc()``, not to be ignored by the task manager. Doing that, we can fix a bug in fleet ignoring failed attempt to start/stop a unit.

Fixes https://github.com/coreos/fleet/issues/998

### DIFF
--- a/systemd/manager.go
+++ b/systemd/manager.go
@@ -114,11 +114,11 @@ func (m *systemdUnitManager) Load(name string, u unit.UnitFile) error {
 
 // Unload removes the indicated unit from the filesystem, deletes its
 // associated Hash from the cache and clears its unit status in systemd
-func (m *systemdUnitManager) Unload(name string) {
+func (m *systemdUnitManager) Unload(name string) error {
 	m.mutex.Lock()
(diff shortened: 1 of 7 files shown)


systemd/manager.go
```diff
@@ -269,14 +269,31 @@ func (m *systemdUnitManager) writeUnit(name string, contents string) error {
...
+func (m *systemdUnitManager) removeUnit(name string) (err error) {
 	log.Infof("Removing systemd unit %s", name)
 
-	m.systemd.DisableUnitFiles([]string{name}, true)
-	m.systemd.ResetFailedUnit(name)
+	// both DisableUnitFiles() and ResetFailedUnit() must be followed by
+	// removing the unit file. Otherwise "systemctl stop fleet" could end up
+	// hanging forever.
+	var errf error
+	func(name string) {
```
jonboulle: What's the purpose of these closures?

Issue #995, opened 2014-10-22
Clarify failure to SSH to remote hosts


bcwaldon:
If I don't have an SSH agent running and I call `fleetctl journal`, I see the following output:

```
cluster-1413996494-01 ~ # fleetctl journal -f app-populate.service
Error running remote command: SSH_AUTH_SOCK environment variable is not set. Verify ssh-agent is running. See https://github.com/coreos/fleet/blob/master/Documentation/using-the-client.md for help.
```

The first thing this error message should say is "unable to SSH to remote host" to make it easier to understand. The same thing goes for any command that attempts to SSH around the cluster.

ianbytchek:
:+1:

sheerun:
Yes, please

PR #1648, opened 2016-07-22
fleetctl: fix error message when fleetctl ssh fails

dongsupark:
Error message of fleetctl ssh should be ``"Unable to SSH to remote host"``.

Fixes https://github.com/coreos/fleet/issues/995


dongsupark:
Trivial. So I'll just merge it.

PR #1649, opened 2016-07-22
fleetctl: print out friendly error message from fleetctl list-machines

dongsupark:
Print out friendly error message when ``fleetctl list-machines`` fails. Indicate that the API isn't working correctly instead of being unavailable. Also mention the most likely culprit, which is something to do with etcd.

Note that I didn't remove the prefix ``"googleapi:"`` as suggested by the original issue, because the prefix doesn't come from fleet, but from the google API itself.

Suggested by @robszumski
Fixes https://github.com/coreos/fleet/issues/1203


dongsupark:
Trivial. Merging.

Issue #1390, opened 2015-11-10
Support templates with some of systemd units


kayrus:
Following unit types don't support templates

* automount
* busname
* device
* mount
* scope
* slice
* snapshot
* swap

We can fix that in systemd code. I suppose changing [this string](https://github.com/systemd/systemd/blob/c386f5886466de6022b3b4b1c8ac8df72871fbc7/src/core/mount.c#L1832) should solve at least `*.mount` units. I like this idea as it will add additional flexibility for fleet. But potentially it could case security risks.

Other solution - is to reject such unit files at fleet side. This is most simple solution but it will not increase fleet flexibility.

/cc @antrik @philips @crawford 

relates to https://github.com/coreos/fleet/issues/303 and https://github.com/coreos/fleet/issues/1194

mrl045:
I would appreciate it when these templates will be available in fleet.

dongsupark:
Apparently it's not that simple to support templates for mount units on the systemd side, according to https://github.com/systemd/systemd/issues/2556#issuecomment-211484263.
I'll create a PR for fleet, to reject template units for the particular unit types, until systemd could somehow manage to support templates for them.

PR #1653, opened 2016-07-28
api: reject templates for particular unit types

dongsupark:
Following unit types don't support templates:

* automount
* busname
* device
* mount
* scope
* slice
* snapshot
* swap

On the other hand, only the following types are allowed for templates:

* path
* service
* socket
* target
* timer

As it's not that simple to support templates for particular unit types on systemd, reject templates for such units from the fleet side.

Fixes https://github.com/coreos/fleet/issues/1390

### DIFF
--- a/api/units.go
+++ b/api/units.go
@@ -160,19 +160,27 @@ const (
 )
 
 var validUnitTypes = pkg.NewUnsafeSet(
-	"service",
-	"socket",
+	"automount",
 	"busname",
-	"target",
-	"snapshot",
 	"device",
 	"mount",
-	"automount",
-	"swap",
-	"timer",
 	"path",
-	"slice",
 	"scope",
+	"service",
+	"slice",
+	"snapshot",
+	"socket",
+	"swap",
+	"target",
+	"timer",
+)
+
+var validTemplateTypes = pkg.NewUnsafeSet(
+	"path",
+	"service",
+	"socket",
+	"target",
+	"timer",
 )
 
 // ValidateName ensures that a given unit name is valid; if not, an error is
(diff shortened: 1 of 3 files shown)


api/units.go
```diff
@@ -193,9 +201,13 @@ func ValidateName(name string) error {
 	if dot == length-1 {
 		return errors.New(`unit name cannot end in "."`)
 	}
-	if suffix := name[dot+1:]; !validUnitTypes.Contains(suffix) {
+	suffix := name[dot+1:]
+	if !validUnitTypes.Contains(suffix) {
 		return fmt.Errorf("invalid unit type: %q", suffix)
 	}
+	if strings.Index(name, "@") != -1 && !templateTypes.Contains(suffix) {
```
jonboulle: strings.Contains

api/units.go
```diff
@@ -160,19 +160,27 @@ const (
...
+	"service",
+	"slice",
+	"snapshot",
+	"socket",
 	"swap",
+	"target",
 	"timer",
+)
+
+var templateTypes = pkg.NewUnsafeSet(
```
jonboulle: validTemplateTypes

jonboulle:
LGTM after a couple of minor ones

dongsupark:
@jonboulle Done.
Replaced `Index() != 1` with `Contains()`, renamed `templateTypes` to `validTemplateTypes`.
I'll merge it tomorrow. Thanks.

stevehorsfield:
This is quite frustrating. As an outsider, it's hard to understand why templates are only supported in some cases.

dongsupark:
@stevehorsfield Yeah, I feel your pain.
As I once mentioned, there's a systemd issue https://github.com/systemd/systemd/issues/2556, which clarifies the underlying issue. It doesn't look trivial to achieve both template support and backward compatibility. TBH I don't have a brilliant idea.

stevehorsfield:
@dongsupark - Thanks for the pointer. I've commented on that ticket. I have an ugly workaround for my current needs using a `service` unit to perform the mount. It's far from ideal.

PR #1654, opened 2016-07-29
fleetctl: use time ticker in waitForSystemdActiveState

dongsupark:
Use time ticker in `waitForSystemdActiveState()` just like other parts of fleet, instead of going into a loop with an arbitrary max number.


dongsupark:
As it's a relatively simple cleanup, I'll merge it today later, unless there's any objection.

PR #1659, opened 2016-08-12
*: fix travis errors from gRPC patches

dongsupark:
This is a follow-up to 2 PRs for gRPC, https://github.com/coreos/fleet/pull/1426 and https://github.com/coreos/fleet/pull/1657.

We need to remove `vendor/google.golang.org/grpc/test`, and also add license headers to gRPC source files, to get rid of errors from travis CI.


dongsupark:
I'm merging it right away, as it's about hotfixes. Without these fixes, travis CI won't pass at all.

PR #1660, opened 2016-08-12
test: add registry/rpc to the list of unit tests

dongsupark:
A new directory `registry/rpc` should be added to the list of unit tests, so that `./test` can test also unit tests under RPC registry. Without this change, even compile errors in unit tests cannot be easily detected, because the `./test` ignores the unit tests.

Actually this fix was originally included in https://github.com/giantswarm/fleet/pull/32, and as far as I can remember, it was merged correctly to https://github.com/coreos/fleet/pull/1426. But somehow it disappeared during rebases, and I just realized that. Though it's not that critical, because OTOH all the relevant fixes included in https://github.com/giantswarm/fleet/pull/32 were correctly merged. So it's just the matter of completeness.


jonboulle:
lgtm

PR #1662, opened 2016-08-16
vendor: clean up unnecessary files to include only code

dongsupark:
We should clean up unnecessary files from `vendor/*`, for example, tests, docs, or etc. Only source code should remain in vendor.

What I did:

```
$ glide vc --only-code --no-tests --no-legal-files
$ git checkout vendor/golang.org/x/net/internal/timeseries
```

Note that we need to keep `golang.org/x/net/internal/timeseries` though, to avoid build errors that internal packages cannot be imported.


jonboulle:
LGTM if it works :-)

PR #1663, opened 2016-08-16
travis,functional: add go 1.7 and exclude ./gopath

dongsupark:
First, for travis, add go 1.7 to the list of required go compiler versions. Now that go 1.7 was released, the list of required go compiler versions should also include go 1.7.
Also in functional tests, install go 1.7 instead of 1.6.3.

And also exclude `./gopath` also from the list for checking license headers.


jonboulle:
lgtm

Issue #1593, opened 2016-05-25
fleetctl: started units output


kayrus:
Played with a batch of units and noticed that output. Not quite sure whether we have to print all units in one line

```
[...]

PR #1664, opened 2016-08-18
fleetctl: introduce an option --max-print-units

dongsupark:
For users who want to limit number of units to be printed, introduce an option `"--max-print-units"` for these commands: destroy, unload, stop, list-unit-files, and list-units.

Note that the default value of `--max-print-units` should be 0, which means `"all units are printed"`, so that the default behavior should remain as before.

Fixes https://github.com/coreos/fleet/issues/1593

/cc @kayrus


jonboulle:
NACK, why bake this into fleet? This doesn't affect performance or anything, so can't people just use `head`?

PR #1666, opened 2016-08-19
fleetctl: support options --no-block, --block-attempts for destroy command

dongsupark:
Apparently fleetctl destroy command has never supported both options `--no-block` and `--block-attempts`, although `runDestroyUnit()` tries to use the flags. Fix it by defining the options for `cmdDestroy`.

Note that it's not a regression caused by conversion into cobra. It has been just like that since the beginning. I just happened to find it out yesterday when I was working on #1664 .


jonboulle:
Huh. Great catch! LGTM

dongsupark:
Fixed minor typos: `s/stop/destroy/`. Will merge it soon.

PR #1667, opened 2016-08-19
fleetctl: check systemd active state via client API

dongsupark:
It's not appropriate to check systemd active states with DBUS connection, as that will return correct states only on the same machine. So we should also introduce a check using `cAPI.UnitStates()`, to check states on other machines.

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -1067,22 +1067,25 @@ func assertUnitState(name string, js job.JobState, out io.Writer) (ret bool) {
 // tryWaitForSystemdActiveState tries to wait for systemd units to reach an
 // active state, making use of the systemd's DBUS interface. First it takes one
 // or more units as input, and ensures that every unit in the []units must be
-// in the active state. If yes, return 0. Otherwise return 1.
+// in the active state. If yes, return nil. Otherwise return error.
 //
 // NOTE: To avoid unnecessary conflicts with unit tests in start_test.go, we
-// should return 0 for the case of sd_dbus.New() returning error. In that case,
+// should return nil for the case of sd_dbus.New() returning error. In that case,
 // it just means that the unit test environment based on Travis is not capable
 // of setting up a DBUS connection of systemd, which would actually work fine
 // for production systems or functional tests.
-func tryWaitForSystemdActiveState(units []string) (ret int) {
+func tryWaitForSystemdActiveState(units []string, maxAttempts int) (err error) {
+	if maxAttempts <= -1 {
+		for _, name := range units {
+			stdout("Triggered unit %s start", name)
+		}
+		return nil
+	}
+
 	conn, err := sd_dbus.New()
 	if err != nil {
 		stderr("Failed to get systemd-dbus conn: %v", err)
-		return 0
-	}
-	if conn == nil {
-		stderr("Got a nil connection")
-		return 0
+		return nil
 	}
 	defer conn.Close()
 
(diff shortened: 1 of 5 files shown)


jonboulle:
I don't think it's ever really sane to fall back to D-Bus, is it?

dongsupark:
@jonboulle Good question.
In the beginning I also tried to get it working only via cAPI. But it didn't work as expected. For some units it's reported to be not active, which made functional tests fail. So I had to fall back to checking via DBUS too, to get it working. I'd also like to know the reason.

jonboulle:
@dongsupark could you link to an example of test failures in that case?

dongsupark:
@jonboulle 
Reverted fallback, printed out more debug messages:
https://gist.github.com/dongsupark/c444e10ad843090cc45750b85db51d7e

dongsupark:
Finally I fixed a bug, and force-pushed.

It turns out, checking states via client API has been already working correctly, and "fleetctl list-units" has actually shown correct output. The only broken thing was the way of using loop around `waitForSystemdActiveState()`, where it didn't actually repeatedly try to retrieve states. So I fixed the logic to make it try to fetch the states repeatedly up to 15 seconds. The fallback routine to checking DBUS connection was removed.

Please have a look. Thanks.

fleetctl/fleetctl.go
```diff
@@ -1067,78 +1072,87 @@ func assertUnitState(name string, js job.JobState, out io.Writer) (ret bool) {
 // tryWaitForSystemdActiveState tries to wait for systemd units to reach an
 // active state, making use of the systemd's DBUS interface. First it takes one
```
jonboulle: This docstring needs updating now

dongsupark: @jonboulle Done. Thanks for spotting it.

fleetctl/fleetctl.go
```diff
@@ -1067,78 +1072,87 @@ func assertUnitState(name string, js job.JobState, out io.Writer) (ret bool) {
...
+	if uState.ActiveState != "active" || uState.LoadState != "loaded" {
+		return fmt.Errorf("Failed to find an active unit %s", unitName)
+	}
+
+	return nil
+}
+
+// getSingleUnitState returns a single uState of type suState, which consists
+// of necessary systemd states, only for a given unit name.
+func getSingleUnitState(unitName string) (suState, error) {
```
jonboulle: I'm a bit concerned about the efficiency of this - `UnitStates()` is a relatively expensive call, and we're only concerned about a single item. It's actually a little worse because we're potentially doing this in multiple goroutines simultaneously, per [`waitForUnitStates`](https://github.com/endocode/fleet/blob/88a0a30207caf0dfaa08ec18a7c2242a7f066278/fleetctl/fleetctl.go#L994).

Two suggestions:
1) introduce a `UnitState.Get` call (similar to how, for units themselves, we have both `Unit.Get` and `Unit.List`)
or,
2) use a single `UnitState.List` call and share the results across all goroutines (in waitForUnitStates)

thoughts?

dongsupark: @jonboulle 
First of all, `getSingleUnitState()` is being called only by [waitForSystemdActiveState()](https://github.com/endocode/fleet/blob/88a0a30207caf0dfaa08ec18a7c2242a7f066278/fleetctl/fleetctl.go#L1104), not by [waitForUnitStates()](https://github.com/endocode/fleet/blob/88a0a30207caf0dfaa08ec18a7c2242a7f066278/fleetctl/fleetctl.go#L994). So it's not true that `getSingleUnitState()` is called from multiple goroutines, because `waitForSystemdActiveState()` simply iterates over units in a sequential way.

On the other hand, your concern is still valid, because it's basically not efficient to call `cAPI.UnitStates()` for each unit. I think I'm going to choose the 2nd option, which is calling only a single `UnitState.List` to share its results for all units. Let me see.

Thanks for pointing it out! :+1:

jonboulle: @dongsupark ah! good point, sorry about that. But now you've pointed that out - wouldn't it make more sense to unify the behaviour of the two so that they're both parallelised? :-)

dongsupark: @jonboulle 
I pushed a new commit to call `cAPI.UnitStates()` only once. Please have a look.

I actually tried to also introduce `cAPI.UnitState()`, to allow this code to fetch a single unit status. However, that's definitely a lot of work all over the place in fleet. (schema, etcdRegistry, rpcRegistry, etc) So I'll do that in the future, creating another PR.

As for using goroutines also for waitForSystemdActiveState(), I'll try to do that.

dongsupark: @jonboulle Force-pushed the commit, after changing the logic from sequential iteration to goroutines.

fleetctl/fleetctl.go
```diff
@@ -1065,80 +1070,128 @@ func assertUnitState(name string, js job.JobState, out io.Writer) (ret bool) {
...
+	}()
+
+	return errchan
+}
+
+func checkSystemdActiveState(apiStates []*schema.UnitState, name string, maxAttempts int, wg *sync.WaitGroup, errchan chan error) {
+	defer wg.Done()
+
+	if maxAttempts < 1 {
 		for {
```
jonboulle: Not particularly fond of the duplication of these loop bodies - can we instead do something more clever with the loop conditional to collapse them into a single `for`?

jonboulle:
Looks pretty good overall

fleetctl/fleetctl.go
dongsupark: @jonboulle Good point. I just split out the common part into `assertFetchSystemdActiveState()`, and let `checkSystemdActiveState()` go into a single loop. Such duplication was just a result of copy-and-paste.

dongsupark:
@jonboulle Now I suppose everything was addressed. I'll merge it soon.

PR #1668, opened 2016-08-22
[WIP] use also systemd's direct connection

dongsupark:
WIP. do not merge.
Just for testing travis.


dongsupark:
Heh, it still fails, even with falling back to NewSystemdConnection(). I'll close it.

PR #1670, opened 2016-08-24
*: avoid using err.Error()

dongsupark:
In principle we should avoid returning `err.Error()`. Instead we should just return err, to be able to inspect the error in a flexible way.

Also in `rpcRenewLeadership()` check for an exact error code from etcd, `EcodeKeyNotFound`, instead of comparing error string with a particular string pattern. Doing that, we can also avoid using `err.Error()`.


dongsupark:
Will merge this PR tomorrow, in case of no objection.

PR #1671, opened 2016-08-25
vendor: update vendors as of 2016-08-25

dongsupark:
Update vendor tree to be up-to-date as of 2016-08-25 to fix commit IDs in `glide.lock`. This part has been probably broken since a month.

Note that this PR does not update `google.golang.org/grpc`, which should stay with [178b68e2](https://github.com/grpc/grpc-go/commit/178b68e2819d3ed711040301c12f3f97b6c270ce) ("Merge pull request 556 from bradfitz/testcleanup", 2016-02-17). That's because the current gRPC logic in fleet requires `ClientConn.State()`, which is currently not available in gRPC 1.0. It's not trivial to change the logic.

Also update `glide-update.sh` to automatically clean up unnecessary directories under etcd and grpc.


dongsupark:
... finally solved all build errors both locally and on travis-ci. Whew!
Though gRPC client logic change still needs to be done in the future. (See https://github.com/coreos/fleet/issues/1672)
I'll merge this PR tomorrow, as it's a critical fix.

Issue #1672, opened 2016-08-25
Change logic of grpc client without calling ClientConn.State()


dongsupark:
As noted in https://github.com/coreos/fleet/pull/1671, we cannot update `google.golang.org/grpc` to v1.0 or master, because `ClientConn.State()` is now gone. That's why we must stay with [178b68e2](https://github.com/grpc/grpc-go/commit/178b68e2819d3ed711040301c12f3f97b6c270ce) ("Merge pull request 556 from bradfitz/testcleanup", 2016-02-17) to avoid build errors.

So solution would be to change `registry/rpc/rpcregistry.go` in fleet to avoid using `ClientConn.State()`. Though AFAICT it's not trivial to change the logic.

See also https://github.com/coreos/etcd/pull/5583

/cc @hectorj2f

PR #1673, opened 2016-08-26
registry/rpc: use simpleBalancer instead of ClientConn.State()

dongsupark:
As `ClientConn.State()` of gRPC has disappeared, we need to avoid using `ClientConn.State()`. Instead we should make use of gRPC rebalancer mechanism, just like etcdv3 is doing. To do that, introduce `simpleBalancer`, as a minimum structure to be used for `grpc.Balancer`.

Note that this change could be possibly invasive to the current behavior of fleet's gRPC. A careful review is needed.

/cc @hectorj2f 
Fixes https://github.com/coreos/fleet/issues/1672


registry/rpc/rpcregistry.go
```diff
@@ -105,26 +79,16 @@ func (r *RPCRegistry) Close() {
...
-		log.Infof("Getting server status...")
-		status, err := r.Status()
-		if err != nil {
-			log.Errorf("unable to get the status of the registry service %v", err)
-			return false
-		}
-		log.Infof("Status of rpc service: %d, connection state: %s", status, connState)
-		return status == pb.HealthCheckResponse_SERVING && err == nil
-	}
+	// NOTE: checking for readyc doesn't work as expected.
```
hectorj2f: Is it WIP ?

vendor/github.com/coreos/go-systemd/activation/listeners.go
```diff
@@ -48,8 +48,6 @@ func TLSListeners(unsetEnv bool, tlsConfig *tls.Config) ([]net.Listener, error)
 	}
 
 	if tlsConfig != nil && err == nil {
-		tlsConfig.NextProtos = []string{"http/1.1"}
```
hectorj2f: Why do you remove this code ?

hectorj2f:
@dongsupark could add a bigger description about what the balancer does ? 
Also, this PR is a bit big. What do you think about a PR for the vendoring packages and another with the logic ? That would make the reviewing much easier and faster.

vendor/github.com/coreos/go-systemd/activation/listeners.go
dongsupark: What I did was just updating vendors with glide.

More precisely, this line in go-systemd was removed by https://github.com/coreos/go-systemd/commit/d736403b010a1169d7a34a9a638b26f07a0658b1.
The change was supposed to be included in https://github.com/coreos/fleet/pull/1671, but actually for some reason it was not included there. That's why you're seeing this in this PR. I think the change should stay there, because glide.lock is already pointing to version https://github.com/coreos/go-systemd/commit/422fafb691792ca9fdf7097a5ebd49f10839676c of `go-systemd` that contains the https://github.com/coreos/go-systemd/commit/d736403b010a1169d7a34a9a638b26f07a0658b1.

registry/rpc/rpcregistry.go
dongsupark: Yeah right. I need to have a deeper look into this one. Unfortunately checking the readyc doesn't work as expected. If you have a better idea, please let me know. TBH I'm just a beginner of the new API of gRPC v1.0 or etcdv3.

dongsupark:
@hectorj2f Sure, I will split it into multiple PRs, and add more descriptions.

vendor/github.com/coreos/go-systemd/activation/listeners.go
hectorj2f: All good then 👍

dongsupark:
Force-pushed it by changing the following:

* Fixed bugs in `IsRegistryReady()`. This function still needs to check `r.Status()` like before. It also needs to check if `r.registryConn` or `r.balancer` is nil.
* Fixed a bug in `rpcAcquireLeadership()` by inserting `reg.IsRegistryReady()` to the condition for checking existing lease. This bug resulted in another failure in `TestDynamicClusterNewMemberUnitMigration()`, which got me locked up in a debugging nightmare. (Now it's solved)
* Simplified balancer.go, removing unnecessary parts like `getHost()`.
* Added high-level descriptions on simpleBalancer.

And I created 2 PRs, https://github.com/coreos/fleet/pull/1676 (only vendor updates) and https://github.com/coreos/fleet/pull/1677 (relevant code), for easier reviews.

dongsupark:
FYI.
Running benchmarks with this PR included, I can observe performance regressions. Nomi benchmarks starting thousands of units over a multi-node cluster, unit starts take much longer time than the current master branch.
I'm not sure why. (gRPC balancer issue? Bug in my patches?)
Until we figured out its reason, let's not merge this PR.

dongsupark:
> Running benchmarks with this PR included, I can observe performance regressions. Nomi benchmarks starting thousands of units over a multi-node cluster, unit starts take much longer time than the current master branch.

I finally got some time to do nomi benchmarks again. This time, I cannot see the regression any more. 

When tested with etcd v3, results of master branch and this branch are nearly the same.
With etcd v2, this branch is even faster than master.
That's actually an expected result, as it's already known that performance got improved with gRPC v1.0, compared to the previous versions.

What I saw in September was probably just a testing mistake.
Tomorrow I'll run again benchmarks several times, and if it's still green, I'll merge this PR.

dongsupark:
I can confirm that its performance is all right.
With etcd2, unit start time was improved by 16%. With etcd3, it's nearly the same.
Benchmark raw data is available on https://github.com/dongsupark/fleet-benchmarks.
I'll merge this PR. Thanks!

PR #1674, opened 2016-08-29
functional: skip TestReplaceSerialization for now

dongsupark:
As on semaphoreci `TestReplaceSerialization()` started to fail much frequently than before, it's a huge pain to make it succeed every time. The failure brings a negative impact on productivity. So skip the entire test for now.


dongsupark:
Today I discussed briefly about this topic with @tixxdz , and he agreed with removing the test for now.

PR #1677, opened 2016-09-08
registry/rpc: use simpleBalancer instead of ClientConn.State()

dongsupark:
This PR contains only the relevant code of https://github.com/coreos/fleet/pull/1673, without vendor updates.
# 

As `ClientConn.State()` of gRPC has disappeared, we need to avoid using `ClientConn.State()`. Instead we should make use of gRPC rebalancer mechanism, just like etcdv3 is doing. To do that, introduce `simpleBalancer`, as a minimum structure to be used for grpc.Balancer.

/cc @hectorj2f

### DIFF
--- a/engine/rpcengine.go
+++ b/engine/rpcengine.go
@@ -110,7 +110,12 @@ func rpcAcquireLeadership(reg registry.Registry, lManager lease.Manager, machID 
 		return l
 	}
 
-	if existing != nil && existing.Version() >= ver {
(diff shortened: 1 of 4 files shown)


hectorj2f:
@dongsupark I'll have a look to the PRs and test them in my local cluster.

dongsupark:
As it's just a part of https://github.com/coreos/fleet/pull/1673, I'll close.

Issue #1675, opened 2016-09-08
Introduce cAPI.UnitState() for a single unit


dongsupark:
At the moment cAPI doesn't support any possibility for retrieving unit states of a single unit. Instead fleet needs to call `cAPI.UnitStates()` to fetch states for all units, which is of course not efficient.

To support `cAPI.UnitState()` for a single unit, we need to do the following:
- Add `UnitState()` to the API interface in `client/api.go` as well as to the HTTPClient in `client/http.go`.
- Add a struct `UnitStateGetCall` and its necessary methods like `Get()`, `Fields()`, `Do()` in schema/v1-gen.go. Of course schema/v1-json.go needs to be regenerated.
- Add `UnitState()` to the `UnitRegistry` interface in registry/interface.go
- Add `UnitState()` also to both `EtcdRegistry` and `RPCRegistry`.
- Add `GetUnitState()` to the Registry service in protobuf/fleet.proto. Of course protobuf/fleet.pb.go needs to be regenerated.

Not sure if I've missed anything.

See also https://github.com/coreos/fleet/pull/1667#discussion_r77678587

PR #1679, opened 2016-09-09
api: support cAPI.UnitState() for a single unit

dongsupark:
- api: support `GET` method for `stateResource`
- schema: introduce `Get` method for `UnitState`
- client: introduce `UnitState()` for API interface
- protobuf: define `GetUnitState()` for registryClient and registryServer
- registry: introduce `UnitState()` for `etcdRegistry` and `RPCRegistry`
- fleetctl: replace `cAPI.UnitStates()` with `.UnitState()`
- registry: add new unit tests for `GetUnitState()` and relevant ones
- fleetctl: periodically check for systemd states using `waitForState`
- fleetctl: make use of `waitForState` also in `assertUnitState()`

Fixes https://github.com/coreos/fleet/issues/1675

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -64,7 +64,7 @@ recommended to upgrade fleetctl to prevent incompatibility issues.
 	clientDriverEtcd = "etcd"
 
 	defaultEndpoint  = "unix:///var/run/fleet.sock"
-	defaultSleepTime = 500 * time.Millisecond
+	defaultSleepTime = 2000 * time.Millisecond
 )
 
 var (
@@ -1116,7 +1116,7 @@ func checkSystemdActiveState(name string, maxAttempts int, wg *sync.WaitGroup, e
 	}
 
 	for attempt := 0; attempt < maxAttempts; attempt++ {
-		if err := assertFetchSystemdActiveState(name); err == nil {
+		if err := assertSystemdActiveState(name); err == nil {
 			return
 		} else {
 			errchan <- err
@@ -1128,29 +1128,29 @@ func checkSystemdActiveState(name string, maxAttempts int, wg *sync.WaitGroup, e
 	}
 }
 
-func assertFetchSystemdActiveState(name string) error {
-	if err := assertSystemdActiveState(name); err == nil {
+// assertSystemdActiveState determines if a given systemd unit is actually
+// in the active state, making use of cAPI.
+// It repeatedly checks up to defaultSleepTimeout. If ActiveState of the given
+// unit is active and LoadState of the given unit is loaded.
+// If it cannot get the expected states within the period, return error.
+func assertSystemdActiveState(unitName string) error {
+	fetchSystemdActiveState := func() error {
+		us, err := cAPI.UnitState(unitName)
+		if err != nil {
+			return fmt.Errorf("Error getting unit state of %s: %v", unitName, err)
+		}
+
+		// Get systemd state and check the state is active & loaded.
+		if us.SystemdActiveState != "active" || us.SystemdLoadState != "loaded" {
+			return fmt.Errorf("Failed to find an active unit %s", unitName)
+		}
 		return nil
 	}
-	time.Sleep(defaultSleepTime)
-	if _, err := cAPI.UnitState(name); err != nil {
-		return err
-	}
-	return nil
-}
 
-// assertSystemdActiveState determines if a given systemd unit is actually
-// in the active state, making use of cAPI. It returns true if ActiveState
-// of the given unit is active and LoadState of the given unit is loaded.
-func assertSystemdActiveState(unitName string) error {
-	us, err := cAPI.UnitState(unitName)
+	timeout, err := waitForState(fetchSystemdActiveState)
 	if err != nil {
-		return fmt.Errorf("Error retrieving list of units: %v", err)
-	}
-
-	// Get systemd state and check the state is active & loaded.
-	if us.SystemdActiveState != "active" || us.SystemdLoadState != "loaded" {
-		return fmt.Errorf("Failed to find an active unit %s", unitName)
+		return fmt.Errorf("Failed to find an active unit %s within %v, err: %v",
+			unitName, timeout, err)
 	}
 
 	return nil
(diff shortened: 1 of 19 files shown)


dongsupark:
Fixed bugs.

* Made `checkSystemdActiveState()` retrieve unit state after sleeps, to avoid occasional errors from assertion of active states.
* Got rid of `getSingleUnitState()` to simplify the logic in `checkSystemdActiveState()`.
* Introduced methods `inmemoryRegistry.UnitState()` as well as `rpcserver.GetUnitState()`. Without these, grpc-enabled tests cannot work at all. It was my fault to have missed these methods. :-(
* Let this PR depend also on https://github.com/coreos/fleet/pull/1673. This part is indeed subtle, because in theory updating vendors gRPC & protobuf would have nothing to do with `cAPI.UnitState()`. Due to dependency through protobuf, however, I cannot manage to make this PR work on the older version of gRPC & protobuf.

My suggestion is to merge first https://github.com/coreos/fleet/pull/1673 and https://github.com/coreos/fleet/pull/1667. After that, we could take some time to review this PR for `cAPI.UnitState()`.

Anyway every functional test should now work as expected. :-)

dongsupark:
Added necesssary unit tests too.

jonboulle:
should I hold off until #1667 #1673 before taking a look?

dongsupark:
@jonboulle Yes.
I think we should first have a look into https://github.com/coreos/fleet/pull/1673 and https://github.com/coreos/fleet/pull/1667 to merge those PRs.

dongsupark:
I managed to get this PR decoupled from #1673. Now it's working even without updating gRPC alltogether.

At the beginning it didn't work with old gRPC, because for some reason the new method `GetUnitState()` was introduced only to `RegistryClient` interface, but not to `RegistryServer`. That's why gRPC connections failed with messages like `"unknown method GetUnitState"`. With `GetUnitState()` included in `RegistryServer` interface, it's working fine.

It's ready for review.

dongsupark:
Although it contains many lines of codes, I think its logic is relatively straightforward.
I'm going to merge it in the next week.

fleetctl/fleetctl.go
```diff
@@ -1140,64 +1128,34 @@ func checkSystemdActiveState(apiStates []*schema.UnitState, name string, maxAtte
...
-	// is not as badly inefficient as it looks, because it will be anyway
-	// rarely called only when the assertion failed. - dpark 20160907
-
 	time.Sleep(defaultSleepTime)
-
-	var errU error
-	apiStates, errU = cAPI.UnitStates()
-	if errU != nil {
-		return fmt.Errorf("Error retrieving list of units: %v", errU)
+	if _, err := cAPI.UnitState(name); err != nil {
```
jonboulle: I'm missing something, what is the point of the sleep + this call here?

fleetctl/fleetctl.go
```diff
@@ -1140,64 +1128,34 @@ func checkSystemdActiveState(apiStates []*schema.UnitState, name string, maxAtte
...
-// in the active state, making use of cAPI
-func assertSystemdActiveState(apiStates []*schema.UnitState, unitName string) error {
-	uState, err := getSingleUnitState(apiStates, unitName)
+// in the active state, making use of cAPI. It returns true if ActiveState
+// of the given unit is active and LoadState of the given unit is loaded.
+func assertSystemdActiveState(unitName string) error {
+	us, err := cAPI.UnitState(unitName)
 	if err != nil {
-		return err
+		return fmt.Errorf("Error retrieving list of units: %v", err)
```
jonboulle: this needs an update

jonboulle:
A couple of questions about the fleetctl code, but overall this looks really great, thanks!

fleetctl/fleetctl.go
dongsupark: @jonboulle Will do.

dongsupark:
Rebased, and pushed 2 more commits, ["fleetctl: periodically check for systemd states using waitForState"](https://github.com/coreos/fleet/pull/1679/commits/3f36e2e2aa59a2286e4dc5bb81d2392c30282b6e) and ["fleetctl: make use of waitForState also in assertUnitState()"](https://github.com/coreos/fleet/pull/1679/commits/08891bf103fd71cdec5bdf067e388c335f9c666e).
I came up with an idea of making use of waitForState, just like that in functional tests. Now the ugly sleeps can be removed.

fleetctl/fleetctl.go
```diff
@@ -1140,64 +1132,37 @@ func checkSystemdActiveState(apiStates []*schema.UnitState, name string, maxAtte
...
-
-	time.Sleep(defaultSleepTime)
+// assertSystemdActiveState determines if a given systemd unit is actually
+// in the active state, making use of cAPI.
+// It repeatedly checks up to defaultSleepTimeout. If ActiveState of the given
+// unit is active and LoadState of the given unit is loaded.
+// If it cannot get the expected states within the period, return error.
+func assertSystemdActiveState(unitName string) error {
+	var errbuf error
+	fetchSystemdActiveState := func() bool {
```
jonboulle: Why not just have this `func() err` and propagate the error instead of using this buffer variable?

dongsupark: @jonboulle Done.

fleetctl/fleetctl.go
```diff
@@ -1263,3 +1222,26 @@ func runWrapper(cf func(cCmd *cobra.Command, args []string) (exit int)) func(cCm
...
+	timeout := defaultSleepTime
+	alarm := time.After(timeout)
+	ticker := time.Tick(250 * time.Millisecond)
+
+	for {
+		select {
+		case <-alarm:
+			return timeout, fmt.Errorf("Failed to fetch systemd active states within %v", timeout)
+		case <-ticker:
+			isChecked, err := stateCheckFunc()
```
jonboulle: Sorry to keep nitting on this, but I might not have explained clearly: `err` will _always_ be `nil` when `isChecked` is `true`, right? which means that you can just remove the `bool` value and rely on the nil-ness of the error :-)

dongsupark:
I suppose everything was addressed. I'm merging it.
In case of any other issues, feel free to comment. Nitpicking is also welcome. :-) I would then create another PR to fix it.

Issue #1665, opened 2016-08-18
Official rkt images


monder:
Are there any official `aci` images for the release? Could we have something similar to https://github.com/coreos/etcd/releases please :)

jonboulle:
@dongsupark any chance you could take a look at this?

dongsupark:
@jonboulle Sure, I'll try to do it this week.

dongsupark:
@jonboulle Using the script written via https://github.com/coreos/fleet/pull/1682, I created 2 ACI images for [v0.12.0](https://github.com/coreos/fleet/releases/download/untagged-363e0bcd48d932709996/fleetd-v0.12.0-amd64.aci) and [v0.13.0](https://github.com/coreos/fleet/releases/download/untagged-17375173c52162328e64/fleetd-v0.13.0-amd64.aci).

Next step would be to release them on https://github.com/coreos/fleet/releases. Currently I cannot do that, because I don't have a gpg private key for "CoreOS ACI Builder". How should I proceed?

Another question: How can I allow the ACI image to be discovered by rkt run, like "rkt run ...  coreos.com/fleet:v0.13.0"?

dongsupark:
ACI images are already provided for v0.11.8, v0.12.0, and v0.13.0. I'm going to provided ACI also for the future release v1.0.
So I suppose this issue can be closed.

PR #1681, opened 2016-09-13
fleetd,main: make github.com/coreos/fleet go-getable

dongsupark:
So far it hasn't been possible to do `"go get github.com/coreos/fleet"` because there was no main.go in the top source directory. Create main.go, a simple wrapper for fleetd to make it _"go-getable"_.

This is a preparation for https://github.com/coreos/fleet/issues/1665, as `goaci` requires a go-getable package.

### DIFF
--- a/build
+++ b/build
@@ -20,7 +20,7 @@ fi
 
 if [ ${GOOS} = "linux" ]; then
   echo "Building fleetd..."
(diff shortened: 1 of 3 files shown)


jonboulle:
LGTM if everything else continues to work!

PR #1683, opened 2016-09-14
Documentation: fix unavailable rkt options in examples

dongsupark:
Apparently rkt doesn't support a single dash in its option any more. For example, `"rkt -volume"` doesn't work any more. Use double dash instead: `"rkt --volume"`

Also `--insecure-skip-verify` doesn't exist. Use `--insecure-options=image` instead.


jonboulle:
lgtm

PR #1685, opened 2016-09-19
registry,fleetctl: fix bugs regarding shadowed error variables

dongsupark:
* registry: make `CreateUnit()` explicitly return error

: When `marshal()` returns a non-nil error `'err'`, which shadows the pre-defined `'err'`, we should explicitly return the second error returned from `marshal()`. Otherwise it will always return `nil`. To fix other potential errors, let's just return `err` explicitly. This was discovered by running "go tool vet":

```
$ go tool vet --shadow ./registry
registry/job.go:316: declaration of "err" shadows declaration at registry/job.go:315
```

* fleetctl: fix wrong error value from `matchLocalFileAndUnit()`

: In `matchLocalFileAndUnit()`, if `os.Stat()` returns nil and `getUnitFromFile()` returns non-nil error, then in the end `matchLocalFileAndUnit()` will return `(false, nil)`. That's not the expected result. It should have returned the second error that came from `getUnitFromFile()`. To avoid further potential bugs, let's follow the idiomatic coding style: first check "err != nil", and fail-fast. This issue was discovered by running "go tool vet".

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -833,18 +833,19 @@ func lazyCreateUnits(cCmd *cobra.Command, args []string) error {
 // Returns true if the contents of the file matches the unit one, false
 // otherwise; and any error encountered.
 func matchLocalFileAndUnit(file string, su *schema.Unit) (bool, error) {
-	result := false
 	a := schema.MapSchemaUnitOptionsToUnitFile(su.Options)
 
 	_, err := os.Stat(file)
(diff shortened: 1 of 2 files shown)


tixxdz:
lgtm thanks @dongsupark , don't know why or if golang is supposed to warn about a local variable that shadows another local variable.

dongsupark:
@tixxdz Merged. Thanks for the review!

PR #1686, opened 2016-09-21
functional: introduce build tags all and dummytest

dongsupark:
For each source file for functional tests, define build tags, `"all"` and `"dummytest"`. Only one of the two tags must be enabled. Accordingly change functional/test script too.

Also define `TestDummy()` that only sets up test clusters, without running any tests. This test might be useful for manual debugging.

### DIFF
--- a/functional/client_test.go
+++ b/functional/client_test.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// +build all !dummytest
+
 package functional
 
 import (
--- a/functional/cluster_test.go
+++ b/functional/cluster_test.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// +build all !dummytest
+
 package functional
 
 import (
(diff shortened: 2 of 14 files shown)


jonboulle:
sorry I don't really get the point of this - can you explain further?

dongsupark:
@jonboulle A little background:
When I need to debug a functional test, one of my typical debugging approaches is to insert `time.Sleep(3600 * time.Second)` to a certain point in a particular test. Then I run `./run-in-qemu --run TestSomething`, which will be blocked for a long time (an hour) at that very point. This way I'm able to achieve those things: launch a test VM, set up a testing cluster with nspawn containers, run fleetd for each container, and do manual debugging.

However doing this every time looks like a dirty hack. So with this PR I wanted to create a dummy test to launch a dummy test cluster, in order to avoid changing functional tests every time.

Now I'm looking into this code again, and I have to admit, all the change indeed looks like overkill. Still I might need to change some parts in functional tests. So I'm fine with closing this PR.
Instead I should probably write a separate script to launch a test cluster, or a command-line tool that is based on go library exported by functional tests.

PR #1689, opened 2016-10-07
Documentation: fix a broken weblink to scripts/build-aci

dongsupark:
Fix weblink to build-aci to
https://github.com/coreos/fleet/blob/master/scripts/build-aci.
This was a side-effect of moving the script to the scripts directory.


jonboulle:
lgtm

PR #1692, opened 2016-10-20
*: introduce an option UseLeaseTTL for lease manager's TTL

dongsupark:
* Introduce an option UseLeaseTTL. Default is false.
* If UseLeaseTTL turned on, create lease manager with the TTL value.
* When gRPC turned on, remove previous workaround of setting leaseTTL
  to a huge value.
* Increase engine reconcile interval from 2 to 5 sec.

Originally written by @hectorj2f 
Taken from https://github.com/giantswarm/fleet/tree/patch_lease_ttl


dongsupark:
It turns out that this PR results in performance regressions. Nomi benchmark shows a large number of missing units, as well as a huge variation of unit start times.
Its reason is that `leaseTTL` is not set to a large value any more in `Engine.Run()`, in case of `enable_grpc==true`.
If the `leaseTTL` is set to a large value again, performance becomes normal again. Though in that case, there's no point in introducing the `UseLeaseTTL` option at all.

Until we could find out a solution, let's not merge.

PR #1693, opened 2016-10-21
registry/rpc: re-enable error handling after NewSystemdUnitManage

dongsupark:
Run `TestRegistryMuxUnitManagement()` only when user's euid is 0 and systemd is available at run-time. Such a distinction is necessary for avoiding test failures, especially when unit tests run on travis CI, where systemd is not available. With those checks, we can re-enable error handling after calling `systemd.NewSystemdUnitManager()`.

Besides fix also a hidden bug in `TestWrongHostKeyFile()` to skip the test on Linux if euid is 0. On Linux `syscall.Setuid()` is not supported in go. So running unit tests with root privileges fails when calling Setuid.

### DIFF
--- a/registry/rpc/registrymux_test.go
+++ b/registry/rpc/registrymux_test.go
@@ -17,6 +17,7 @@ package rpc
 import (
 	"io/ioutil"
 	"os"
+	"syscall"
 	"testing"
 
 	etcd "github.com/coreos/etcd/client"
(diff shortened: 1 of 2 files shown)


dongsupark:
Last week I changed this PR, to check euid and systemd not in a shell script, but in `registrymux_test.go`. That way we could get the checks applied to only a necessary test case, instead of an entire directory `registry/rpc`.

I'm going to merge this PR in this week, as it fixes real bugs, and it changes only unit tests.

Issue #1245, opened 2015-06-10
conflicts with mutilple units


duffqiu:
Hi,

I configure a unit with conflicts, but i do that to set muliple units in one line as below:

```
Conflicts=activemq-hub-slave1@*.service activemq-hub-master@%i.service activemq-hub-slave2@%i.service
```

but only the first one wokr, the other doesn't work.

I try to another way with mutiple Conflicts lines, one line with one unit,
it seems works
like below:

```
Conflicts=activemq-hub-master@*.service
Conflicts=activemq-hub-slave1@%i.service
Conflicts=activemq-hub-slave2@%i.service
```

what is the problem? it is a bug or not? or it is the systemd bug?

I check systemd spec, it mentions it can be configures with mutilple units in one line.

[systemd spec](http://www.freedesktop.org/software/systemd/man/systemd.unit.html)

the fleet version is 0.9.2 in coreos 647.2

duffqiu:
one more thing, if i start mutilple units at the same time, fleetctl can't find the proper way to assign the machine.

for example, 

one unit with blow conflicts:

```
Conflicts=activemq-hub-master@*.service
Conflicts=activemq-hub-slave1@%i.service
Conflicts=activemq-hub-slave2@%i.service
```

another unit with below  conflicts:

```
Conflicts=activemq-hub-slave1@*.service
Conflicts=activemq-hub-master@%i.service
Conflicts=activemq-hub-slave2@%i.service
```

after start activemq-hub-master@{1..3}.service,
i try to start activemq-hub-slave1@{1..3}.service

sometimes fleetctl will stuck because  (I only 3 machines)
activemq-hub-master@1 assigned to A
activemq-hub-master@2 assigned to B
activemq-hub-master@3 assigned to C

and 
activemq-hub-slave1@1 assigned to B
activemq-hub-slave1@2 assigned to A

then it is impossible to assign activemq-hub-slave1@3 because activemq-hub-master@3 assigned to C

Maybe we need to add some logic to handle the multiple assignment at the same time to aoivd the conflicts.

Dynom:
I'm not an expert in any way, although I've already spend several days struggling getting things to work (fleet, etcd, CoreOS, etc.). It's far from a smooth ride so far. Also more intelligent (re-)location of units would be a very desirable option.

[...]

PR #1695, opened 2016-10-26
agent: allow HasConflict() to handle multiple values defined in Conflicts

dongsupark:
So far when parsing values defined in the `Conflicts` option, it hasn't taken multiple values into account. So let's allow `j.Conflicts()` to parse multiple values, splitting with white-space delimiters into a slice of each string.

In `AgentState.HasConflict()`, split a common part for checking if a unit belongs to a Conflicts list, into a new helper `hasStringInSlice()` Also return a slice of conflicted strings, instead of a single string. Doing that, `HasConflict()` can be more readable.

Also extend unit tests and functional tests.

Fixes https://github.com/coreos/fleet/issues/1245

### DIFF
--- a/agent/reconcile_test.go
+++ b/agent/reconcile_test.go
@@ -387,6 +387,30 @@ func TestAbleToRun(t *testing.T) {
 			want: job.JobActionUnschedule,
 		},
 
+		// conflicts found
+		{
+			dState: &AgentState{
+				MState: &machine.MachineState{ID: "123"},
+				Units: map[string]*job.Unit{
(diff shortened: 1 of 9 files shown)


dongsupark:
Updated.

* Added more test cases for unit tests, `agent.TestHasConflicts()` and `job.TestJobConflicts()`.
* Renamed `isConflict` to `hasStringInSlice`, as it could be also used for other non-Conflicts cases.

dongsupark:
I just double-checked. It looks fine.
I'll merge it, as a fix for the old bug should be included in the release v1.0.

Issue #1694, opened 2016-10-26
systemd: turn off force flag when calling LinkUnitFiles()


dongsupark:
At the moment `systemdUnitManager.writeUnit()` [calls `systemd.LinkUnitFiles()` with the `force` flag turned on](https://github.com/coreos/fleet/blob/master/systemd/manager.go#L268). This brings negative impact on performance.

So `force` should be set to `false` by default. It might be also a good idea to introduce a new option in `fleetd.conf`, so that users are able to set the flag.

/cc @hectorj2f 
See also https://github.com/coreos/go-systemd/issues/189#issuecomment-244048755

PR #1696, opened 2016-10-26
fleetd: allow users to turn off force flag for systemd.LinkUnitFiles()

dongsupark:
So far `systemd.LinkUnitFiles()` has been called with the force option turned on. That was one of the reasons of performance issues.

Let's introduce an option `systemd_link_unit_force`, false by default. Users can set the option in `fleetd.conf`. That option will be passed directly to `systemd.LinkUnitFiles()`.

/cc @hectorj2f 
Fixes https://github.com/coreos/fleet/issues/1694

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -42,6 +42,7 @@ type Config struct {
 	UnitsDirectory          string
 	SystemdUser             bool
 	AuthorizedKeysFile      string
+	SystemdLinkUnitForce    bool
 }
 
 func (c *Config) Capabilities() machine.Capabilities {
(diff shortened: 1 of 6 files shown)


jonboulle:
What's the actual reason for the bad performance?

dongsupark:
@jonboulle According to a report(https://github.com/coreos/fleet/pull/1393#issuecomment-217883320), it's because systemctl link is being executed for each unit. I think the topic was also mentioned in the talk by @hectorj2f at the last CoreOS Fest.
BTW I suppose there could be other use cases where turning off the force option would be useful.

jonboulle:
@dongsupark ok, just seems like we could do better at communicating this to users? hinting at why they might want to use it for example

PR #1698, opened 2016-11-01
agent,engine: fix bugs in rescheduling for replaced units

dongsupark:
So far in engine reconciler, `decide()` returns a bool variable to determine whether it's possible to schedule or not. Reschedule was not handled in a proper way.

So let's create a new tri-state for `job.JobAction`, i.e. `JobActionSchedule`, `JobActionUnschedule`, and `JobActionReschedule`. And return the tri-state variable instead of bool. That way `AbleToRun()` can check return values from `HasConflict()` and `HasReplace()` in a correct manner.

With that, we can fix bugs in rescheduling for replaced units. While the `'Replaces'` option has been supported since https://github.com/coreos/fleet/pull/1572, the engine didn't actually unschedule units to be replaced. It was a bug.

So let's implement `GetReplacedUnit()` to expose the replaced unit from `AgentState` to the engine reconciler. And make the engine reconciler unschedule the replaced unit, and schedule the current unit. The engine scheduler's decision structure needs to also track if the unit needs to be rescheduled, to be used by the scheduling path.

### DIFF
--- a/agent/reconcile_test.go
+++ b/agent/reconcile_test.go
@@ -275,48 +275,48 @@ func TestAbleToRun(t *testing.T) {
 	tests := []struct {
 		dState *AgentState
 		job    *job.Job
-		want   bool
+		want   job.JobAction
 	}{
 		// nothing to worry about
 		{
 			dState: NewAgentState(&machine.MachineState{ID: "123"}),
 			job:    &job.Job{Name: "easy-street.service", Unit: unit.UnitFile{}},
-			want:   true,
+			want:   job.JobActionSchedule,
 		},
 
 		// match MachineID
 		{
 			dState: NewAgentState(&machine.MachineState{ID: "XYZ"}),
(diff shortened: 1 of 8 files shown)


dongsupark:
Updated.
Refined the scheduling logic, updated functional tests to do correct tests for unit replacement.
As it's about fixing bugs, I'm going to merge it soon, on Monday next week at latest.

dongsupark:
Updated.
Refined the scheduling logic, updated functional tests to do correct tests for unit replacement.
As it's about fixing bugs, I'm going to merge it soon, on Monday next week at latest.

dongsupark:
Updated.
Refined the scheduling logic, updated functional tests to do correct tests for unit replacement.
As it's about fixing bugs, I'm going to merge it soon, on Monday next week at latest.

dongsupark:
Oops, I didn't mean to close. It was just a bug in the github UI displaying nothing even after writing a comment.

Issue #1699, opened 2016-11-02
functional: TestNodeShutdown fails with CoreOS 1185


dongsupark:
Since CoreOS stable was updated to 1185.3.0 (with systemd 231), `TestNodeShutdown` always fails like that:

```
--- FAIL: TestNodeShutdown (fleet.conf=[enable_grpc=false]) (8.98s)
        node_test.go:83: Unit hello.service not reported as inactive:
```

This is rather an urgent issue. I'm working on it.

PR #1700, opened 2016-11-02
functional: fix error from message change from systemctl status

dongsupark:
Since systemd v231, `TestNodeShutdown` fails with the following message.

```
--- FAIL: TestNodeShutdown (fleet.conf=[enable_grpc=false]) (8.98s)
        node_test.go:83: Unit hello.service not reported as inactive:
```

It's because of behavioral change of `"systemctl status"`. Until systemd v230, `"systemctl status hello.service"` has printed

```
Loaded: not-found (Reason: No such file or directory)
Active: inactive (dead)
```

to stdout, when hello.service is not active. So the functional test `TestNodeShutdown` has expected a string `"Active: inactive"` in the message from stdout.

Since systemd v231, however, the output string has become

```
Unit hello.service could not be found
```
And it's printed to stderr, instead of stdout.

As `MemberCommand()` doesn't return stderr at all, we need to make it return stderr too. Changing prototype of `MemberCommand()` however requires a lot of changes all over functional tests. So let's add another helper `MemberCommandStderr()`, a distinct helper that returns both stdout and stderr. Then we can change only `TestNodeShutdown()`
without touching other tests.

In the long run, we need to change `MemberCommand()` directly, removing `MemberCommandStderr()`.

Fixes https://github.com/coreos/fleet/issues/1699

### DIFF
--- a/functional/node_test.go
+++ b/functional/node_test.go
@@ -77,9 +77,14 @@ func TestNodeShutdown(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// The member's unit should actually stop running, too
-	stdout, _ = cluster.MemberCommand(m0, "systemctl", "status", "hello.service")
-	if !strings.Contains(stdout, "Active: inactive") {
+	// The member's unit should actually stop running, too.
+	// NOTE: In case of no units, systemd v230 or older prints out
+	// "Active: inactive" to stdout, while systemd v231 or newer prints out
+	// "Unit NAME could not be found" to stderr. So we need to check for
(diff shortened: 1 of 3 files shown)


dongsupark:
Ok, functional tests work fine now.
Travis CI fails suddenly with an error "unable to access gnutls_handshake() failed", when cloning a git repo. Not sure why. Though that looks unrelated. 
I'm going to merge this PR as soon as possible. as it's rather an urgent fix.

dongsupark:
Yep, the gnutls handshake error was just intermittent. Now it's working fine.
I'll merge.

PR #1701, opened 2016-11-09
fleetctl: take experimentalAPI into account in getClient

dongsupark:
`getClient()` has not taken into account the case of `!experimentalAPI`, before calling registryClient. It has set endPoint to a specific value of URLs, but the new value would not be passed to `getRegistryClient()`, which simply fetches endPoint again. This was a regression since 848d3561 ("fleetctl: convert cli to cobra").

To fix that, introduce `getEndpoint()` that does GetString("endpoint") as well as the special handling for experimentalAPI. And make `getRegistryClient()` call `getEndpoint()`.

Fortunately, the experimentalAPI flag is set to true by default, so this special handling path is not actively used after all. Just for correctness.

### DIFF
--- a/fleetctl/fleetctl.go
+++ b/fleetctl/fleetctl.go
@@ -319,20 +319,7 @@ func getClientAPI(cCmd *cobra.Command) client.API {
 
 // getClient initializes a client of fleet based on CLI flags
 func getClient(cCmd *cobra.Command) (client.API, error) {
-	// The user explicitly set --experimental-api=false, so it trumps the
-	// --driver flag. This behavior exists for backwards-compatibilty.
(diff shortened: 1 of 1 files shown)


dongsupark:
I just manually tested it, and I confirm that it fixes the regression.

```
$ /home/core/fleet/bin/fleetctl --experimental-api=false --endpoint=http://172.18.1.1:54728 list-machines
MACHINE         IP              METADATA
e7da15e6...     172.18.1.1      hostname=smokee7da15e607aa4695be37708374d31ec6
```

It works fine now.
Creating another functional test for that option is probably overkill, as the option is obsolete after all.
I'll merge it.

jonboulle:
lgtm!

PR #1702, opened 2016-11-09
functional: allow MemberCommand to return stderr

dongsupark:
Remove `Cluster.MemberCommand()`, rename `MemberCommandStderr()` to `MemberCommand()`, and update every call site accordingly.
That way we can clean up leftovers from https://github.com/coreos/fleet/pull/1700.


dongsupark:
I just did double-checks. It looks fine.
As it's trivial clean-ups, I'm merging.

PR #1704, opened 2016-11-11
functional: reduce number of tests and total running time

dongsupark:
Functional test seems to have a limitation of 40 tests in total. When the number of tests becomes more than 40, no more cluster member can be created, so that test ends up with hanging mysteriously. To avoid that, let's reduce number of tests, mainly for dynamic metadata tests as well as unit action tests. It's just squashing similar tests.

As go test natively has an upper limit of 600sec in total for all tests, we should reduce total running time. So let's reduce number of units to be tested in the unit state tests, and reduce maximum block attempts for the metadata template test. This will fix the failure on semaphoreci, which has suddenly appeared since the dynamic metadata tests. This should make tests work for now. In the long run, it would be necessary to be able to tune the timeout.

### DIFF
--- a/functional/machines_test.go
+++ b/functional/machines_test.go
@@ -90,25 +90,11 @@ func TestMachinesList(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	testMachinesListBadNextPageToken(t, m)
 }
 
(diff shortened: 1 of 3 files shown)


jonboulle:
can this be reverted after #1706?

dongsupark:
@jonboulle Sure, I'll revert it.

Issue #1712, opened 2016-12-02
engine: TestScheduleMachineOf fails with gRPC turned on


dongsupark:
`TestScheduleMachineOf` fails when enable_grpc is set to true. That failure happens only occasionally. You can see the following logs:

```
fleetd[94]: transport: http2Client. notifyError got notified that the client transport was broken EOF.
fleetd[94]: DEBUG agent.go:87: HeartbeatJobs tick
fleetd[94]: DEBUG unit_state.go:222: Pushing UnitState(ping.0.service) to Registry: &unit.UnitState{LoadState:"loaded", ActiveState:"active", SubState:"running", MachineID:"10c94346e797450d94d3e9b0f657cae4", UnitHash:"b4f04719a20575e450899ac57097fc6d8ac12cc4", UnitName:"ping.0.service"}
fleetd[94]: DEBUG unit_state.go:222: Pushing UnitState(pong.0.service) to Registry: &unit.UnitState{LoadState:"loaded", ActiveState:"active", SubState:"running", MachineID:"10c94346e797450d94d3e9b0f657cae4", UnitHash:"a38421afebf0062e438f25c03c5f24e8202a3a70", UnitName:"pong.0.service"}
fleetd[94]: DEBUG unit_state.go:222: Pushing UnitState(ping.0.service) to Registry: &unit.UnitState{LoadState:"loaded", ActiveState:"active", SubState:"running", MachineID:"10c94346e797450d94d3e9b0f657cae4", UnitHash:"b4f04719a20575e450899ac57097fc6d8ac12cc4", UnitName:""}
fleetd[94]: DEBUG unit_state.go:222: Pushing UnitState(pong.0.service) to Registry: &unit.UnitState{LoadState:"loaded", ActiveState:"active", SubState:"running", MachineID:"10c94346e797450d94d3e9b0f657cae4", UnitHash:"a38421afebf0062e438f25c03c5f24e8202a3a70", UnitName:""}
fleetd[94]: ERROR registrymux.go:166: Retry to connect to new engine: dial tcp 172.18.1.1:50059: getsockopt: connection refused
fleetd[94]: DEBUG mux.go:56: HTTP GET /fleet/v1/machines?alt=json
fleetd[94]: ERROR registrymux.go:166: Retry to connect  to new engine: dial tcp 172.18.1.1:50059: getsockopt: connection refused
[...]

PR #1713, opened 2016-12-02
engine: fix a bug in engine being unreachable

dongsupark:
When gRPC is turned on, `TestScheduleMachineOf` fails sometimes, as the engine becomes unreachable with the following error messages:

```
transport: http2Client. notifyError got notified that the client transport was broken EOF.
ERROR registrymux.go:166: Retry to connect to new engine: dial tcp 172.18.1.1:50059: getsockopt: connection refused
ERROR registrymux.go:166: Retry to connect to new engine: dial tcp 172.18.1.1:50059: getsockopt: connection refused
ERROR registrymux.go:166: Retry to connect to new engine: dial tcp 172.18.1.1:50059: getsockopt: connection refused
```

This must have been a regression from commit ecb121a ("registry/rpc: use simpleBalancer instead of ClientConn.State()"). Remove the additional checking with IsRegistryReady, in order to avoid the occasional case of engine being unreachable.

Fixes https://github.com/coreos/fleet/issues/1712

### DIFF
--- a/engine/rpcengine.go
+++ b/engine/rpcengine.go
@@ -110,12 +110,7 @@ func rpcAcquireLeadership(reg registry.Registry, lManager lease.Manager, machID 
 		return l
 	}
 
-	// If reg is not ready, we have to give it an opportunity to steal lease
-	// below. Otherwise it could be blocked forever by the existing engine leader,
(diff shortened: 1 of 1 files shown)


dongsupark:
Yep, it seems to really help. I'll merge.

PR #1714, opened 2016-12-05
fleetctl: increase default sleep time from 2 to 4 seconds

dongsupark:
As fleetctl start sometimes needs to wait up to 4 seconds until systemd active states get confirmed, we need to increase the `defaultSleepTime` from 2 to 4 seconds. Otherwise, a simple `"fleetctl start`" could sometimes fail with timeout.

Also reduce total running time of all tests, to avoid crash of functional tests resulted by the global timeout 600 secs.


dongsupark:
Duh, this change is not necessary at all. "fleetctl start" already works just fine.
The issue is only a misleading error message. I'll create another PR to fix the message.

Issue #1716, opened 2016-12-08
fleet: server monitor fails to shutdown process


mwitkow:
After a short `etcd` blip, `fleet` has issues on its agent and engine, but the process remains up. This is affecting `v0.13.0`

The symptoms are as follows: the `Monitor` detects the server failed heartbeat, asks all `components` to shut down, but the shutdown of all components never completes. This means that most `components are dead, the server process is still up, but serves:
```{"error":{"code":503,"message":"fleet server unable to communicate with etcd"}}```

The full error log is here:
```
Dec 07 19:28:55 eu2-prod-core-hasu fleetd[3250]: ERROR engine.go:221: Engine leadership lost, renewal failed: client: etcd cluster is unavailable or misconfigured
Dec 07 19:28:56 eu2-prod-core-hasu fleetd[3250]: ERROR job.go:109: failed fetching all Units from etcd: client: etcd cluster is unavailable or misconfigured
Dec 07 19:28:56 eu2-prod-core-hasu fleetd[3250]: ERROR reconcile.go:120: Failed fetching Units from Registry: client: etcd cluster is unavailable or misconfigured
Dec 07 19:28:56 eu2-prod-core-hasu fleetd[3250]: ERROR reconcile.go:73: Unable to determine agent's desired state: client: etcd cluster is unavailable or misconfigured
Dec 07 19:28:59 eu2-prod-core-hasu fleetd[3250]: ERROR job.go:109: failed fetching all Units from etcd: client: etcd cluster is unavailable or misconfigured
Dec 07 19:28:59 eu2-prod-core-hasu fleetd[3250]: ERROR engine.go:236: Failed fetching Units from Registry: client: etcd cluster is unavailable or misconfigured
Dec 07 19:28:59 eu2-prod-core-hasu fleetd[3250]: ERROR reconciler.go:59: Failed getting current cluster state: client: etcd cluster is unavailable or misconfigured
Dec 07 19:28:59 eu2-prod-core-hasu fleetd[3250]: WARN engine.go:117: Engine completed reconciliation in 4.004575849s
Dec 07 19:29:01 eu2-prod-core-hasu fleetd[3250]: ERROR job.go:109: failed fetching all Units from etcd: client: etcd cluster is unavailable or misconfigured
Dec 07 19:29:01 eu2-prod-core-hasu fleetd[3250]: ERROR reconcile.go:120: Failed fetching Units from Registry: client: etcd cluster is unavailable or misconfigured
[...]

PR #1717, opened 2016-12-08
make Monitor shut down the process hard if it fails to shutdown grace…

mwitkow:
Should fix:
https://github.com/coreos/fleet/issues/1716

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -246,7 +246,7 @@ func (s *Server) Supervise() {
 	select {
 	case <-done:
 	case <-time.After(shutdownTimeout):
-		log.Errorf("Timed out waiting for server to shut down, shutting down server *hard*")
+		log.Errorf("Timed out waiting for server to shut down, shutting down server without cleanup.")
 		os.Exit(1)
 		sd = true
 	}
--- a/server/server.go
+++ b/server/server.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"net"
 	"net/http"
+	"os"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jonboulle:
Interesting catch. I wonder if we should just panic or something similar to dump a stack trace in this situation?

mwitkow:
Hmm, panic would work, but I'm wondering why the tests are failing:
```
--- FAIL: TestNodeShutdown (9.70s)
	node_test.go:83: Unit hello.service not reported as inactive:
```

dongsupark:
@mwitkow That test failure looks just intermittent. If you just trigger another round of functional tests, that failure could disappear. (Yes I know, it is annoying.)

mwitkow:
@dongsupark thanks, stupid question though: how do I trigger semaphore? Do i need to push another thing?

dongsupark:
@mwitkow Right. Change anything, commit it, and force-push it.
If you don't want to change anything, run "git commit --amend" to append a space to the commit message, to be able to force-push it again. That should do the trick.

dongsupark:
@mwitkow Apart from the test failure on semaphoreci, my local test works fine.
I also did run tests with the branch rebased on master, which works also fine for both gRPC and !gRPC.
I think, what needs to be done is only to call `panic()` instead of `os.Exit(1)`. Just let me know, once you change that. Then I could merge.

dongsupark:
@mwitkow gentle ping.
I'd like to get it merged in this week, before release v1.0. Should I just merge this one first, then create another PR to replace os.Exit(1) with panic()?

mwitkow:
@dongsupark thanks for the ping, I changed it to `panic` with a message.

dongsupark:
LGTM. Thanks for the bug report and its fix! :+1:

